PR #10, opened 2020-03-25
🎉 App NavBar

mamadOuologuem:
image

Mobile version

Screen Shot 2020-03-26 at 5 32 33 PM

Screen Shot 2020-03-26 at 5 32 41 PM


src/components/App/App.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { ThemeProvider } from '@material-ui/core/styles';
+import { Router, RouteComponentProps } from '@reach/router';
+
+import NavBar from '../NavBar/NavBar';
+import muiTheme from '../../muiTheme';
+
+// eslint-disable-next-line no-unused-vars
+const Hello = (props: RouteComponentProps) => <div>hello</div>;
```
ericboucher: is it unused?

src/components/NavBar/utils.tsx
```diff
@@ -0,0 +1,71 @@
+import { chain, map, startCase, camelCase, mapKeys } from 'lodash';
```
bencpeters: This isn't actually a `tsx` file is it? It's just normal typescript (no JSX)

src/components/NavBar/utils.tsx
```diff
@@ -0,0 +1,71 @@
...
+import climate from '../images/icon_climate.png';
+import impact from '../images/icon_impact.png';
+
+const icons: { [key: string]: string } = {
+  baseline,
+  climate,
+  impact,
+};
+
+export interface ILayer {
```
bencpeters: I suspect this is a type that is going to be used widely in the App (the data processing sections of the page will also need access to it). Let's not define it in a `NavBar` specific path.

Maybe `src/config/types.ts`?

src/components/NavBar/utils.tsx
```diff
@@ -0,0 +1,71 @@
+import { chain, map, startCase, camelCase, mapKeys } from 'lodash';
+
+import appJSON from '../../config/prism.json';
+import layersJSON from '../../config/layers.json';
+import baseline from '../images/icon_basemap.png';
+import climate from '../images/icon_climate.png';
+import impact from '../images/icon_impact.png';
+
+const icons: { [key: string]: string } = {
```
bencpeters: This is going to be config-specific (some countries might have different options). I think the `src/components/Navbar/**/*` should be all generic - anything that's deployment specific should go in `src/config`

src/components/NavBar/utils.tsx
```diff
@@ -0,0 +1,71 @@
...
+function formatLayersList(layersList: { [key: string]: String[] }): TLayers {
+  const formattedLayersList: any = map(
+    layersList,
+    (layersKey, layersListKey) => ({
+      title: startCase(layersListKey),
+      layers: chain(layersJSON)
+        .pick(layersKey as any)
+        .map((value, layerKey) => ({
+          id: layerKey,
+          ...mapKeys(value, (_value: string, key: string | undefined) =>
```
bencpeters: is the `key` ever actually `undefined`?

src/components/NavBar/utils.tsx
```diff
@@ -0,0 +1,71 @@
...
+        }))
+        .value(),
+    }),
+  );
+  return formattedLayersList as TLayers;
+}
+
+export const categories: TCategories = chain(appJSON)
+  .get('categories')
+  .map((layersList, categoryKey) => {
```
bencpeters: This could just be a direct object-value-return? `... => ({ ...`

src/config/prism.json
```diff
@@ -0,0 +1,19 @@
+{
+  "country": "Indonesia",
+  "map": {
+    "latitude": -3.88927,
+    "longitude": 118.621,
+    "zoom": 5
+  },
+  "categories": {
+    "baseline": {
+      "baseline": ["popupaltion_below_poverty", "population_density"]
```
bencpeters: spelling - `population_below_poverty`

src/index.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import './index.css';
 import { Provider } from 'react-redux';
-import App from './App';
+import App from './components/App/App';
```
bencpeters: This import path is a bit ugly - could be maybe move to `components/App/index.tsx` and import from `./components/App`? (I _think_ that works, check it though)

src/components/App/App.test.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import App from './App';
+
+jest.mock('../NavBar/NavBar', () => 'mock-NavBar');
```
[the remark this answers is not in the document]
mamadOuologuem: I had those warnings by doing that

image

src/components/NavBar/MenuItem/Category/__snapshots__/Category.test.tsx.snap
```diff
@@ -0,0 +1,54 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`renders as expected 1`] = `
+<div>
+  <div
+    class="Category-container-1"
+  >
+    <p
```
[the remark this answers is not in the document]
mamadOuologuem: Agreed on that 👍🏽

src/components/NavBar/utils.tsx
mamadOuologuem: @bencpeters 

`icons` represent all icons. So from the `src/config/prism.json` at `categories` we can determine which one to use in the navbar.

Another way?

src/index.tsx
mamadOuologuem: @bencpeters 

I am facing an error with that 😶

src/components/App/App.test.tsx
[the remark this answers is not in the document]
bencpeters: hrm, weird, I wonder if there's a change in the rendering library.

Could you do `React.createElement('NavBar')`? or does that give you an error too?

src/components/NavBar/utils.tsx
bencpeters: Ah, I see what you're saying. This is probably a fine way to do it then.

src/index.tsx
bencpeters: Is it similar to [this issue](https://github.com/Microsoft/vscode/issues/45289)?

It looks like the default export of the `index.tsx` file might need to match the name of the module (rather than being wrapped in a HOC which `withStyles` does.

If so, you could try something like:

```javascript
function AppComponent(...) {
  ...
}

const App = withStyles(AppComponent);
export default App;
```

Does that work?

bencpeters: This might depend on the babel config with TypeScript too

src/components/NavBar/NavBar.tsx
```diff
@@ -1,24 +1,59 @@
...
+  },
+  {
+    title: 'Github',
+    icon: faGithub,
+    link: '/',
+  },
+];
+
+function NavBar({ classes }: NavBarProps) {
+  const [openMobileMenu, setOpenMobileMenu] = useState(false);
```
bencpeters: Do you need this boolean state or could you just save off the anchor element?

mamadOuologuem: Oups the anchor wasn't needed so I get rid of it 👍🏽

src/components/App/App.test.tsx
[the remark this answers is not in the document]
mamadOuologuem: image

PR #17, opened 2020-03-31
🎉 Add date selector

mamadOuologuem:
### TODO
- Pass the unavailables dates to the `DateSelector` component
- Include the date in the wms query (cf branch `wms-get-date`)

As example, the nine first day are unactive in the screenshot
image


bencpeters:
I'll review the actual code in a minute, but I'm wondering if you've thought at all about how we can display the set of available dates? There will be some variation, but I think the most common data sets starting out are going to be:

- Monthly interval data (published on a given day on a monthly interval - "valid" for that whole month).

- 10 day interval data (same idea, but published every 10 days, and valid for a 10 day period)

- Discrete events - Data related to specific disasters/discrete events at unpredictable historic intervals (e.g. September 2009, January 2013, April 2015, ...)

I think the _most_ important thing that our date selection interface can do is to let us easily which dates in the past have data, and navigate to those dates. A "nice to have" would be an easy way to see the period/interval that the particular data is valid for (e.g. valid for 30 days, or 10 days), but that could easily be handled in a text description of the data, I think it's less important to have that visually represented than it is to have a really easy way of seeing what data is available. In the past versions, we've found that a month-by-month selector like this doesn't work very well for this purpose.

Do you have any thoughts/ideas for a better interface? Maybe work a quick mock up before coding?  Maybe some kind of variable period timeline that graphically shows dates? (e.g. it could be monthly for monthly interval or longer data, zoomed in for other?). Other ideas?

src/context/filters/filtersSlice.ts
```diff
@@ -23,20 +25,30 @@ export const slice = createSlice({
...
       { layers, ...state },
       { payload }: PayloadAction<LayerType>,
     ) => ({
       ...state,
       layers: layers.delete(payload),
     }),
+
+    updateDate: ({ layers, ...state }, { payload }: PayloadAction<Date>) => ({
+      ...state,
+      // 'layers' is getting converted into a object instead of Set<LayerType>
```
bencpeters, line 39: This seems unsustainable - we don't want to need to specifically call out parts of state in unrelated reducers all over (and maintain that as we add to state). Generally speaking, we're going to run into issues if we mix and match Immutable.js with regular JS like this.

[Here's some good reading on redux with TypeScript](https://redux.js.org/recipes/using-immutablejs-with-redux)

If we want to use Immutable.js for state, I think we need to use it completely (which means foregoing de-structuring), and having the whole of state be a `Map` rather than a plain JS object. Then a reducer would involve `state.set` or `state.withMutation` or something like that.

- or - 

We could just go back to regular JS objects for state.

I confess that I do generally like Immutable structures a lot, but if you don't want to deal with the complexity of that, that's fine. But I don't think we should be mixing & matching JS objects & Immutable ones like this.

src/context/filters/filtersSlice.ts
```diff
@@ -23,20 +25,30 @@ export const slice = createSlice({
...
+      layers: layers as Set<LayerType>,
+      currentDate: payload,
+    }),
   },
 });
 
 // Getters
-export const selectlayers = (state: RootState) => state.filters.layers;
+export const selectLayers = (state: RootState) => state.filters.layers;
+export const selectCurrentDate = (state: RootState) =>
```
bencpeters, line 48: If you keep immutable state you _probably_ can still use destructoring on the selectors if you want.

src/components/MapView/DateSelector/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+import 'react-datepicker/dist/react-datepicker.css';
+
+import {
+  selectCurrentDate,
+  updateDate,
+} from '../../../context/filters/filtersSlice';
+
+interface InputProps {
+  value?: string;
+  onClick?: () => void;
```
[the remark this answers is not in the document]
mamadOuologuem: To set a custom input with the library we can pass this prop `customInput={<Input />}`. Then `DatePicker` will be working with the custom input trough the `ref`. As `<Input />` is not specifying directly the props, that is why I set the props as optional

```
<DatePicker
          ...
          customInput={<Input />}
        />
```

src/context/mapStateSlice.ts
```diff
@@ -0,0 +1,39 @@
...
+    removeLayer: (state, { payload }: PayloadAction<LayerType>) =>
+      state.update('layers', layers => (layers as LayersSet).delete(payload)),
+
+    updateDate: (state, { payload }: PayloadAction<Date>) =>
+      state.set('currentDate', payload),
+  },
+});
+
+// Getters
+export const selectLayers = (state: RootState) =>
```
ericboucher, line 31: The name `selectLayers` sounds weird.
Do you mean something like `getAvailableLayers` or `selectedLayers`?

mamadOuologuem: @ericboucher 
I named it like that as it is used like that `const layers = useSelector(selectLayers);` I tried to follow the convention on [redux documentation](https://react-redux.js.org/api/hooks#useselector-examples)

Another option can be `layersSelector`?

ericboucher: Yeah I think I'd prefer that :)

src/components/MapView/DateSelector/index.tsx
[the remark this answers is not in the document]
bencpeters: Ah, that makes sense. They must use `React.cloneElement` or something similar internally.

src/components/MapView/DateSelector/index.tsx
```diff
@@ -31,41 +36,109 @@ const Input = forwardRef(
...
+      dispatch(updateDate(date));
+    }
+  }
+
+  function incrementYear() {
+    dispatch(updateDate(new Date(selectedYear + 1, selectedMonth, 1)));
+  }
+
+  function decrementYear() {
+    dispatch(updateDate(new Date(selectedYear - 1, selectedMonth, 1)));
```
bencpeters, line 73: So the way this works right here, when we incr/decr the year on the timeline, it changes both the timeline element AND the selected date for data display purposes, right? (and potentially sets the "active" date to a date with no data available, since it defaults to the first of the month in a different year). Am I understanding that correctly?

I think we probably want to separate that state out into 2 different pieces:

- 1 date that is the "active" date for purposes of displaying data on the map. This is the date that gets set when you click on a month (or on a day in the calendar picker).

- The current year that the timeline is showing. This lets you quickly scroll through the timeline by year to see when data is available for without the map constantly changing/updating (which might involve a lot of data downloads, or putting us into dates with no data available).

src/components/MapView/DateSelector/index.tsx
```diff
@@ -31,41 +36,109 @@ const Input = forwardRef(
...
+  }
+
   function updateStartDate(date: Date) {
-    if (!isUnavailableDate(date)) {
+    if (isAvailableDate(date)) {
```
bencpeters, line 77: Can we use the UI element itself for this check? e.g. this function _only_ is callable (via `disabled` or something) when the date actually is available, rendering this check redundant?

If that's hard, it's not a big deal.

src/context/mapStateSlice.ts
```diff
@@ -0,0 +1,39 @@
+import { Map, Set } from 'immutable';
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+import { RootState } from './store';
+import { LayerType } from '../config/types';
+
+interface LayersSet extends Set<LayerType> {}
+interface MapState extends Map<string, LayersSet | Date> {}
```
bencpeters, line 8: This *might* want to be `Map<string, Any>`? In general more specific types are better, but this state could end up growing into a pretty large union of types depending on what we add...

src/components/MapView/DateSelector/index.tsx
```diff
@@ -37,46 +40,34 @@ const Input = forwardRef(
...
-    if (date) {
-      dispatch(updateDate(date));
-    }
+    const { startDate, endDate } = getMonthStartAndEnd(month, selectedYear);
+    dispatch(updateDateRange({ startDate, endDate }));
   }
 
   function incrementYear() {
-    dispatch(updateDate(new Date(selectedYear + 1, selectedMonth, 1)));
+    setSelectedDate(selectedDate.year(selectedYear + 1).clone());
```
bencpeters, line 61: Is this the best way to do this? I can't remember the `moment` interface, but does `.year` mutate the original object? It probably doesn't matter much in this case, but would `selectedDate.clone().year(selectedYear + 1)` make more sense? 

I assume `.year` mutates the object rather than a copy?

src/components/MapView/DateSelector/utils.ts
```diff
@@ -15,6 +15,48 @@ export const months = [
...
+ * @param year
+ * @param availableDates
+ * @return
+ */
+export function isAvailableMonth(
+  month: number,
+  year: number,
+  availableDates: Date[],
+) {
+  return findAvailableDayInMonth(month, year, availableDates).length;
```
bencpeters, line 61: Let's convert this to a `boolean` - especially using an inferred return type, Typescript is going to "think" of this function as returning an integer, not a boolean

PR #18, opened 2020-04-07
🔧 Firebase config

mamadOuologuem:
Adding firebase configuration

**Prod:** https://prism-frontend.firebaseapp.com
**Staging:** https://staging-prism-frontend.web.app

The layers request are blocked by _Mixed content error_ (the server is on HTTP and the Firebase domain on HTTPS)
image

[Possible solutions](https://stackoverflow.com/questions/34523385/can-sites-on-firebase-hosting-include-non-https-resources)

Recommended solution: Use an HTTPS protocol on the server

### New 404 page
image


package.json
```diff
@@ -72,6 +72,8 @@
     "start": "EXTEND_ESLINT=true react-scripts start",
     "build": "EXTEND_ESLINT=true react-scripts build",
     "test": "react-scripts test",
+    "deploy:prod": "yarn build && firebase deploy --only hosting:prod-target",
```
bencpeters, line 75: `scripts` can include a `pre` task that runs before the actual deploy by default, but separates the actual deploy (uploading to Firebase) and the necessary precondition (building).

```
"predeploy:prod": "yarn build",
"predeploy:staging": "yarn build",
"deploy:prod", ...
```

(See the package.json on Ovio Explore)

Also, does the static build involve "baking" in a URL for assets? Do we want another parameter for `yarn build` to specify a staging build vs. a production build?

public/404.html
```diff
@@ -0,0 +1,88 @@
...
+          box-shadow: none;
+        }
+        body {
+          border-top: 16px solid #ffa100;
+        }
+      }
+    </style>
+  </head>
+  <body>
+    <div id="message">
```
bencpeters, line 79: What about rendering the 404 page within the usual template so that it's easy to navigate back? Or if that's hard, maybe have a link back to the index here?

package.json
mamadOuologuem: For now, there is no difference between a staging build and a production build ...

yarn.lock
```diff
@@ -11752,13 +11740,16 @@ walker@^1.0.7, walker@~1.0.5:
   dependencies:
     makeerror "1.0.x"
 
+<<<<<<< HEAD
```
bencpeters, line 11743: Uh, oh, this shouldn't be in here! Looks like you've got some artifacts from a merge conflict - probably just regenerate this file...

ericboucher:
@mamadOuologuem you need to mock the `new Date()` creation otherwise the snapshots will change every day ;)

PR #20, opened 2020-04-17
Load admin boundaries

ericboucher:
This implements the basic functionalities required to display the admin boundaries and get info on click.

Screenshot:

Screen Shot 2020-04-16 at 6 25 25 PM

### DIFF
--- a/src/components/MapView/Boundaries/index.test.tsx
+++ b/src/components/MapView/Boundaries/index.test.tsx
@@ -1,14 +1,27 @@
 import React from 'react';
+import ReactMapboxGl from 'react-mapbox-gl';
 import { render } from '@testing-library/react';
 
-import MapView from '..';
+import Boundaries from '.';
 
-// Boundaries need to live in a "Map", so we
-// mock sub-components except for Boundaries
-jest.mock('../Layers', () => 'mock-Layers');
-jest.mock('../DateSelector', () => 'mock-DateSelector');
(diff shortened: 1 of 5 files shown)


ericboucher:
I am getting an error in the tests:
`TypeError: Cannot read property 'addSource' of undefined`

I think it is because the style doesn't have time to load: https://stackoverflow.com/questions/25448553/mapbox-gl-source-addsource-bug, any thoughts?

src/components/MapView/Boundaries/index.test.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+// Boundaries need to live in a "Map", so we
+// mock sub-components except for Boundaries
+jest.mock('../Layers', () => 'mock-Layers');
+jest.mock('../DateSelector', () => 'mock-DateSelector');
+
+test('renders as expected', () => {
+  const { container } = render(<MapView />);
+  expect(container).toMatchSnapshot();
+});
```
mamadOuologuem, line 14: What about?

```
import React from 'react';
import { render } from '@testing-library/react';
import Boundaries from '.';

  const { container } = render(<Map><Boundaries /></Map>);
  expect(container).toMatchSnapshot();
});
```

ericboucher: might need to import Map though? ;)

PR #21, opened 2020-04-17
Add GitHub link and About

ericboucher:


ericboucher:
Note: in the future we will have a popup for "about"

PR #25, opened 2020-04-24
Use selected startDate in layer query

ericboucher:
Use startDate info in the layer query.

This surfaced a few issues/needs:
- [x] need for a better definition of layers. We could probably get rid of the path and use a config for servers, then a server_layer_id and server_style. (#29 )

- [x] For now, we should only be able to select one layer that has dates at a time. (#28 )

- [x] handle errors when fetching data. In particular, have the ability to parse time mismatch "Time dimension value '2020-02-10' not valid for this layer". (#27 )

- [x] We are currently getting the "Time dimension value '2020-02-10' not valid for this layer" because of a timezone difference which creates a one-day diff. 2020-02-11 is available. We should indicate the timezone in the server definition. (Fixed by adding a moment default timezone)


src/components/MapView/DateSelector/index.test.tsx
```diff
@@ -6,10 +6,16 @@ import DateSelector from '.';
 import { store } from '../../../context/store';
 
 test('renders as expected', () => {
+  const realDateNow = Date.now.bind(global.Date);
```
mamadOuologuem, line 9: I was thinking about mocking the whole store

This shouldn't be a problem anymore because the initial values of dates are number

src/components/MapView/DateSelector/index.tsx
```diff
@@ -27,6 +27,9 @@ import {
 import { months, getMonthStartAndEnd, isAvailableMonth } from './utils';
 import { AvailableDates } from '../../../config/types';
 
+// Set default timezone to UTC
```
mamadOuologuem, line 30: Do it make sense to set that in `src/components/App/index.tsx`?

src/config/types.ts
```diff
@@ -4,6 +4,7 @@ export interface LayerType {
   id: string;
   title: string;
   serverType: string;
+  serverLayer: string;
```
mamadOuologuem, line 7: Should it be optional?

src/components/MapView/index.tsx
```diff
@@ -17,10 +17,12 @@ const Map = ReactMapboxGl({
 function MapView({ classes }: MapViewProps) {
   const layers = useSelector(layersSelector);
   const serverAvailableDates = useSelector(availableDatesSelector);
-  const selectedLayersDates = serverAvailableDates.filter((dates, layerId) =>
-    layers.has(layerId),
+  const selectedLayersDates = serverAvailableDates.filter((_, layerId) =>
```
mamadOuologuem, line 20: For example in `src/config/layers.json`: 

```
"ndvi": {
    "title": "Modis NDVI",
    "server_type": "wms",
--> "server_layer": "ModisIndices",
    "server_uri": "http://185.30.8.6:5000/wms?styles=ndvi&layers=ModisIndices",
    "has_date": true,
    "date_interval": "month",
    "opacity": 0.3,
    "legend_text": "Modis NDVI"
},
```

The `layerId` is `ndvi` and `server_layer` is `ModisIndices`.

I will keep `layers.has(layerId)` 🤔?

src/components/MapView/DateSelector/index.test.tsx
ericboucher: This is a problem because the maxDate keeps changing

src/components/MapView/DateSelector/index.tsx
ericboucher: Let's have you play around with that when fixing #30

src/components/MapView/index.tsx
ericboucher: The names from the server and the layerId are not the same. We can discuss that if thtat's not clear but I will likely change a few things in #29

PR #32, opened 2020-04-28
🎉 Fetching server dates

mamadOuologuem:
### What is done
- Fetching `availableDates` from the specified servers in `src/config/prism.json`

### Notes
- [x] Access to fetch at 'https://mng-wfp.ovio.org:8443/geoserver/prism/wms?request=GetCapabilities' from origin 'http://localhost:3000' has been blocked by CORS policy
- [x] The layers returned by the server are squared. We may investigate if we can make then fit on the country boundaries.
- [ ] More test on dates (server available dates, layers' `time` argument, date selector)
- [x] Double-check the `serversUrls` in `src/config/prism.json`


ericboucher:
```
Access to fetch at 'mng-wfp.ovio.org:8443/geoserver/prism/wms?request=GetCapabilities' from origin 'http://localhost:3000' has been blocked by CORS policy
```

I think that's because the server is down atm... Will check on it

ericboucher:
```
The layers returned by the server are squared. We may investigate if we can make then fit on the country boundaries.
```

This is fine for now and probably part of the how tilling works in the data cube

src/config/prism.json
```diff
@@ -1,5 +1,9 @@
 {
   "country": "Mongolia",
+  "serversUrls": {
+    "wcs": ["http://185.30.8.6:5000"],
```
ericboucher, line 4: So technically in Mongolia we have 2 servers with base URLS:

geoserver: https://mng-wfp.ovio.org:8443/geoserver/prism/
openDataCube: http://185.30.8.6:5000/


They can **both** return WMS & WCS layers using /wms?... or /wcs?... . The difference is that they don't return GetCapabilities the same way for WMS, which is why the python function I shared with you was a bit convoluted.

ericboucher: We will be using `geoserver` only for WMS though and `openDataCube` for both.
I think we should restructure the layers in `layers.json` to take:

- `server_id` instead of the `server_uri`
- `server_type`
- `server_layer`
- `server_style` (optional)

So that we can build URLs like this one programmatically -> `http://185.30.8.6:5000/wms?styles=ndvi&layers=ModisIndices`

ericboucher: Thoughts @bencpeters ?

ericboucher:
@mamadOuologuem mng-wfp.ovio.org should be back up

ericboucher:
@mamadOuologuem the date selector is working for me locally but not when deployed on https://staging-prism-frontend.firebaseapp.com/

ericboucher:
Let's make sure the top links for About and GitHub look activated:

Screen Shot 2020-05-05 at 2 46 40 PM

src/context/mapStateSlice.ts
```diff
@@ -1,30 +1,33 @@
...
   startDate?: number;
   endDate?: number;
 }
 interface MapState extends Map<string, any> {}
 
 const initialState: MapState = Map({
-  layers: Set<LayerType>(),
-  dateRange: {},
+  layers: Map() as LayersMap,
+  dateRange: {} as DateRange,
```
bencpeters, line 15: Again, don't mix and match Immutable & regular JS objects. Since this is all Redux state with Immutable, we'd like `DateRange` (and `MapState` for that matter) to be strongly typed `Maps` instead of objects or weak (because of the `any`) types.

[This SO post is useful for that](https://stackoverflow.com/questions/46284131/typescript-and-immutable-js-interface) - read the first answer.

You can do something like:

```
// The default Map interface accepts <K,V>: Key, Value.
// Build an interface that also accepts 'T': the shape of your data.
// You might want to add other methods, e.g. `update`? Probably can cut and paste more or less from
// https://github.com/immutable-js/immutable-js/blob/7f4e61601d92fc874c99ccf7734d6f33239cec8c/type-definitions/Immutable.d.ts#L2449
export interface IImmutableMap<T, K, V> extends Map<K, V> {
  toJS(): T;
  get<I extends keyof T>(key: I & K): T[I] & V;
  set<S extends keyof T>(key: S & K, value: T[S] & V): Map<K, V>;
}


export interface IMapState extends Map<string, any> {
  layers: LayersMap,
  // Do a similar typed definition for `DateRange`
  dateRange: DateRange,
};

export type MapState = IImmutableMap<IMapState, string, any>;

// Update the type definition on initial state to your type.
const initialState: MapState = Map({
  // Not sure if you'll have to cast these or if TS will just pick them up as a compatible type
  layers: Map(),
  dateRange: Map(),
});
```

(haven't tested this, but something like it should work)

src/context/mapStateSlice.ts
```diff
@@ -1,30 +1,33 @@
...
 
 export const mapStateSlice = createSlice({
   name: 'mapState',
   initialState,
   reducers: {
     addLayer: (state, { payload }: PayloadAction<LayerType>) =>
-      state.update('layers', layers => (layers as LayersSet).add(payload)),
+      state.update('layers', layers =>
+        (layers as LayersMap).set(payload.id, payload),
+      ),
```
bencpeters, line 25: Could you just do `state.setIn(['layers', payload.id], payload)`?

src/context/mapStateSlice.ts
```diff
@@ -1,30 +1,33 @@
...
 
     removeLayer: (state, { payload }: PayloadAction<LayerType>) =>
-      state.update('layers', layers => (layers as LayersSet).delete(payload)),
+      state.update('layers', layers =>
```
bencpeters, line 28: `deleteIn`?

src/context/serverStateSlice.ts
```diff
@@ -0,0 +1,34 @@
...
+import { AvailableDates } from '../config/types';
+import { LAYERS_AVAILABLE_DATES } from '../constants';
+
+interface ServerState extends Map<string, any> {}
+
+const initialState: ServerState = Map({
+  availableDates: Map(
+    JSON.parse(localStorage.getItem(LAYERS_AVAILABLE_DATES) || '[]'),
+  ) as AvailableDates,
+});
```
bencpeters, line 14: Is this an expensive call? It seems like it just adds a bunch of unnecessary state & complexity to bother with local storage, why not just init this on load?

src/context/store.ts
```diff
@@ -1,15 +1,67 @@
...
+import { AvailableDates } from '../config/types';
 
-export const store = configureStore({
-  reducer: {
+const initializeStore = async (getCurrentState: any) => {
+  const isStorageSet = localStorage.getItem(LAYERS_AVAILABLE_DATES) !== null;
+
+  if (isStorageSet) {
+    // Aborting the initialization
+    return Promise.reject();
```
bencpeters, line 24: You're already in an `async` function, you shouldn't ever need to explicitly call `Promise`. You probably want to `throw` here?

src/context/store.ts
```diff
@@ -1,15 +1,67 @@
...
+  const currentState = getCurrentState();
+  const { serverState } = currentState;
+
+  localStorage.setItem(
+    LAYERS_AVAILABLE_DATES,
+    JSON.stringify(layersAvailableDates),
+  );
+
+  return new Promise(resolve => {
+    resolve({
```
bencpeters, line 37: This doesn't actually do anything - the return from an `async` function is already wrapped in a Promise, just return the object.

src/context/store.ts
```diff
@@ -1,15 +1,67 @@
...
+  }
+
+  const layersAvailableDates = await getLayersAvailableDates();
+  const currentState = getCurrentState();
+  const { serverState } = currentState;
+
+  localStorage.setItem(
+    LAYERS_AVAILABLE_DATES,
+    JSON.stringify(layersAvailableDates),
+  );
```
bencpeters, line 34: I would just skip this, unless it's a very expensive call?

src/context/store.ts
```diff
@@ -1,15 +1,67 @@
...
+import { getLayersAvailableDates } from '../utils/server-utils';
+import { LAYERS_AVAILABLE_DATES } from '../constants';
+import { AvailableDates } from '../config/types';
 
-export const store = configureStore({
-  reducer: {
+const initializeStore = async (getCurrentState: any) => {
+  const isStorageSet = localStorage.getItem(LAYERS_AVAILABLE_DATES) !== null;
+
+  if (isStorageSet) {
```
bencpeters, line 22: The dates change on the server as new data loads, I don't think this is a safe condition to have. We need to periodically update our view of available data.

src/context/store.ts
```diff
@@ -1,15 +1,67 @@
...
   // serialize the state
-  middleware: getDefaultMiddleware({
-    serializableCheck: false,
-  }),
+  middleware: [
+    middleware(initializeStore),
+    ...getDefaultMiddleware({
+      serializableCheck: false,
+    }),
+  ],
```
bencpeters, line 64: Is this just boilerplate from async redux? It seems complicated, but I haven't used that package before.

src/utils/server-utils.ts
```diff
@@ -0,0 +1,165 @@
...
+  unset(parsedUrl, 'search');
+
+  const selectedDateString = selectedDate
+    ? moment(selectedDate).format('YYYY-MM-DD')
+    : '';
+
+  return decodeURI(
+    format({
+      ...parsedUrl,
+      query: merge(query, queryProp, { time: selectedDateString }),
```
bencpeters, line 34: `{ time: selectedDateString }` should probably be conditionally included depending on whether we have a `selectedDate`?

Also, you can probably just use normal es6 object: `{...query, ...queryProp, ...selectedDate && { time: selectedDateString }}`

src/utils/server-utils.ts
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Format the raw data to { [layerId]: availableDates }
+ * @param rawLayers Layers data return by the server 'GetCapabilities' request
+ * @param layerIdPath path to layer's id
+ * @param datesPath path to layer's available dates
+ * @returns an object shape like { [layerId]: availableDates }
+ */
+function formatCapabilitiesInfo(
+  rawLayers: any,
```
bencpeters, line 47: You don't really need to explicitly define `any` types - that's the default. In general, you should _try_ to do stronger typing than `any`, but if it's not possible (or worth it), you don't need to actually type them as `any`

src/utils/server-utils.ts
```diff
@@ -0,0 +1,165 @@
...
+      return {
+        ...acc,
+        [layerId]: union(availableDates, oldLayerDates),
+      };
+    }
+
+    return {
+      ...acc,
+      [layerId]: availableDates,
+    };
```
bencpeters, line 74: Probably can just do:
```
const { [layerId]: oldLayerDates } = acc;
return { ...acc, [layerId]: union(availableDates, oldLayerDates) };
```

src/utils/server-utils.ts
```diff
@@ -0,0 +1,165 @@
...
+          flattenLayers,
+          'Name._text',
+          'Dimension._text',
+        );
+
+        resolve(Map(layers));
+      })
+      .catch(error => {
+        console.error(
+          `Server returns an error for request GET/${requestUri}, error: ${error}`,
```
bencpeters, line 106: `s/returns/returned`

src/utils/server-utils.ts
```diff
@@ -0,0 +1,165 @@
...
+          : get(rawLayers, 'Layer', []);
+
+        const layers = formatCapabilitiesInfo(
+          flattenLayers,
+          'Name._text',
+          'Dimension._text',
+        );
+
+        resolve(Map(layers));
+      })
```
bencpeters, line 103: We're _generally_ using `async-await`, any particular reason you're using a Promise chain here?

src/components/MapView/DateSelector/index.tsx
```diff
@@ -89,7 +92,7 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
             showYearDropdown
             dropdownMode="select"
             customInput={<Input />}
-            includeDates={availableDates}
+            includeDates={flatten(availableDates.valueSeq().toJS())}
```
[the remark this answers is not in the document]
mamadOuologuem: @bencpeters 

`availableDates.valueSeq().flatten()` return an iterable (`IndexedCollection`)
`availableDates.valueSeq().flatten().toJS()` return an array of array

eg. `{ layer1: [1, 2, 3], layer2: [2, 3, 4] }` with Immutable.flatten I have `[ [1, 2, 3, 2, 3, 4] ]` and I need to access `[0]` for the result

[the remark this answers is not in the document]
mamadOuologuem: My bad, I stopped mixing Immutable and JS objects. I am using `Immutable.flatten`

src/context/mapStateSlice.ts
mamadOuologuem: Cf Issue #41

src/context/store.ts
mamadOuologuem: Without the returned `Promise`, I am having an error.
Looks like it's needed (cf [docs](https://github.com/KELiON/redux-async-initial-state#partial-replace))

src/context/store.ts
mamadOuologuem: Yes, it was the easiest one I found ([link](https://github.com/KELiON/redux-async-initial-state))

src/utils/server-utils.ts
mamadOuologuem: Yeah, I think it's not worth giving a type to `rawLayers` because it comes from the server and it is formatted as soon as we received it.

Should I turn off the `strict mode` on `tsconfig.json`? 
The strict mode automatically turn on `noImplicitAny`
image

I kinda prefer explicit any.

src/utils/server-utils.ts
mamadOuologuem: Not a big reason ... I prefer the `then/catch` syntax than `try/catch` (I think it more readable and less `await`)

But less make that more conventional.

PR #45, opened 2020-05-16
Upload NSO data for baseline layers

dubwalla:
Adding JSON and a few CSV files to be included in the baseline layers and as inputs for impact calculations


src/config/layers.json
```diff
@@ -32,5 +32,14 @@
     "date_interval": "day",
     "opacity": 0.3,
     "legend_text": "ModisLST"
+  },
+  "DzudRiskMap": {
+    "title": "Dzud Risk Map",
+    "server_type": "wms",
+    "server_uri": "http://3.136.214.209:8080/geoserver/prism/wms?layers=prism%3ADZUDRM12311",
```
ericboucher, line 39: @dubwalla you should be using the https address here for the geoserver.

src/config/layers.json
```diff
@@ -32,5 +32,14 @@
...
+    "has_date": true,
```
ericboucher, line 40: Also, does it really have dates?

ericboucher: Dzud is actually already on master. Maybe you haven't refreshed to the latest version?

Issue #7, opened 2020-03-15
Error and alerting tools


ericboucher:
Tools for monitoring system performance and alerting critical errors are published

ericboucher:
@dubwalla let's get WFP's preference here. The question is: what error monitoring tool would they recommend for a simple static website.

Otherwise, we can simply implement the free version of Sentry - https://sentry.io/

dubwalla:
Our colleagues in Rome said:
We make a heavy use of [Nagios](https://www.nagios.org/) for the monitoring and alerting part and [Grafana](https://grafana.com/) for the data visualization, both open source products. But we don’t really have any recommendation on which tool to use.

ericboucher:
@dubwalla do they have an organization account for WFP for Nagios?

dubwalla:
Looking into it

dubwalla:
Nope, we do not have access to Nagios through an organization account

PR #46, opened 2020-05-16
Add Sentry

ericboucher:


ericboucher:
Fixes #7

src/components/App/index.tsx
```diff
@@ -7,6 +8,13 @@ import MapView from '../MapView';
 import NotFound from '../404Page';
 import muiTheme from '../../muiTheme';
 
+if (process.env.NODE_ENV && process.env.NODE_ENV !== 'development') {
+  Sentry.init({
+    dsn:
+      'https://a3cc5f470dac4b50b288851a0390f686@o213971.ingest.sentry.io/5242966',
```
bencpeters, line 14: This probably shouldn't be hardcoded. It's not a secret _strictly_ speaking (in the cybersecurity sense), since it's included in client side code in a "public" app, but it's probably better not to leak it, and given that each deployment is basically autonomous (and quite likely based on slightly different codebases), we might well want to set up a different Sentry account for each country deployment.

I'd pull this from an env variable (_maybe_ with a fall-back value to this key if not found just to make deployments easier, although I think it'd be fine to just throw an error in that case too)

PR #47, opened 2020-05-16
Add national statistics office data

ericboucher:
Adding a new type of layer to create vector layers out of baseline data and the admin boundaries geojson.

Screen Shot 2020-05-18 at 11 32 40 AM


ericboucher:
@bencpeters @mamadOuologuem what's the best way to handle the "Unallowed reassignment" here?

ericboucher:
Screen Shot 2020-05-16 at 5 04 05 PM

src/config/baselines.tsx
```diff
@@ -0,0 +1,55 @@
+import nsoDisabled from '../data/nso/NSO_Disabled_Admin1_Total.json';
```
bencpeters, line 1: Should we put this in a subdirectory of `config`? I think the original intention was for everything that might change country-to-country in `config`, so that you could be confident that that's the only place to make changes... but we could just add this `data` directory too.

src/config/baselines.tsx
```diff
@@ -0,0 +1,55 @@
...
+import nsoPop from '../data/nso/NSO_Population_Admin2_Total.json';
+
+export interface NsoData {
+  TBL_ID: string;
+  Period: string;
+  CODE?: string;
+  SCR_MN: string;
+  SCR_ENG: string;
+  CODE1?: string;
+  CODE2?: null;
```
bencpeters, line 17: Is this really `null` typed? This type definition literally says `CODE2` will be either `undefined` or `null`, was that the intention?

src/config/baselines.tsx
```diff
@@ -0,0 +1,55 @@
...
+  SCR_MN1: string;
+  SCR_ENG1: string;
+  SCR_MN2?: null;
```
bencpeters, line 20: Same with these, are these actually `null` typed?

src/config/baselines.tsx
```diff
@@ -0,0 +1,55 @@
...
+  nsoHerders,
+  nsoHerdsize,
+  nsoChild,
+  nsoLivestock,
+  nsoElderly,
+  nsoPoverty,
+  nsoPop,
+};
+
+type DatasetKeys =
```
bencpeters, line 40: `type DatasetKeys = keyof nsoDatasets` I think should work here and be more maintainable/less prone to typos.

src/config/baselines.tsx
```diff
@@ -0,0 +1,55 @@
...
+  | 'nsoDisabled'
+  | 'nsoChild'
+  | 'nsoHerders'
+  | 'nsoHerdsize'
+  | 'nsoLivestock'
+  | 'nsoElderly'
+  | 'nsoPoverty'
+  | 'nsoPop';
+
+export function getNSOData(dataset: string | undefined) {
```
bencpeters, line 50: This type signature is better written as `dataset?: string`.

More generally though, this function is throwing away the safety that TypeScript should be giving us. There's no reason to make this an optional parameter and then do all these checks that the compiler can make much easier than we can.

Writing the function this way:

```typescript
export function getNSOData (dataset: DatasetKeys = 'nsoPoverty') {
  return nsoDatasets[dataset];
}
```

Should give us the same result, but ensure that the compiler gives us a compile-time error if we try to look-up using an invalid key, which is really the whole point of using TypeScript.

src/components/MapView/NSOLayers/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    get(evt.features[0], 'properties.ADM2_PCODE'),
+    get(evt.features[0], 'properties.data'),
+  );
+}
+
+const linePaint: MapboxGL.LinePaint = {
+  'line-color': 'grey',
+  'line-width': 1,
+  'line-opacity': 0.3,
+};
```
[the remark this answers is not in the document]
ericboucher: Actually not needed here, good catch

src/config/baselines.tsx
ericboucher: string, typo

src/config/baselines.tsx
ericboucher: Hmm good point, not urgent but let's think about how we want to handle country specific data

src/config/baselines.tsx
ericboucher: Tried a few options but couldn't get it to work as I wanted. I'll add a Todo

src/config/baselines.tsx
ericboucher: Looks like it. I am getting an error if I say string.
Should I say `string | null`?

bencpeters: Ya, that's the appropriate type. Javascript handles `null` and `undefined` differently - the `?` only covers `undefined`.

It's actually somewhat intentional - `undefined` is intended to signify that nothing was, well, defined, whereas `null` is intended to an explicit signal that the value is null'ed out. In vanilla JS it gets confused because the type-casting with boolean-ish values is so imprecise that it's easy to get sloppy with the distinction.

src/config/baselines.tsx
bencpeters: Sorry, it should be `type DatasetKeys = keyof typeof nsoDatasets`

PR #50, opened 2020-05-18
Add groundstation points

ericboucher:
Screen Shot 2020-05-18 at 12 31 23 AM


src/components/MapView/GroundstationLayers/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,9 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`renders as expected 1`] = `
```
bencpeters, line 3: I made this comment in your other PR too - I don't think these types of tests are worth having. The test should either be re-written to actually render something that tests the layer code, or we should just get rid of it - it's just adding extra complexity without extra value as is here.

(Unless the specific goal is to test whether the layer handles an empty dataset properly)

src/components/MapView/GroundstationLayers/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+declare module 'geojson' {
+  export const version: string;
+  export const defaults: Object;
+  export function parse(
+    data: Object,
+    propreties: Object,
+    callback?: Function,
+  ): Object;
+}
```
bencpeters, line 18: This belongs in `custom.d.ts`. But probably better to use [@types/geojson](https://www.npmjs.com/package/@types/geojson) for more complete definitions (I think this was in your other PR?)

src/components/MapView/GroundstationLayers/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+        value: string;
+        color: string;
+      }[]
+    | undefined,
+) {
+  return (legend || []).map(({ value, color }: any) => [
+    parseFloat(value),
+    color,
+  ]);
+}
```
bencpeters, line 56: This should probably be a shared utility function with the one in PR #47 - same comments apply to this version of it if you keep them separate.

src/components/MapView/GroundstationLayers/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { LayersMap } from '../../../config/types';
+
+import groundstationDataJson from '../../../data/groundstations/longterm_data.json';
+
+declare module 'geojson' {
+  export const version: string;
+  export const defaults: Object;
+  export function parse(
+    data: Object,
+    propreties: Object,
```
bencpeters, line 15: spelling - should be `properties`

src/components/MapView/GroundstationLayers/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  console.log(
+    get(evt.features[0], 'properties.index'),
+    get(evt.features[0], 'properties.aimagname'),
+    get(evt.features[0], 'properties.sumname'),
+    get(evt.features[0], 'properties.rasterheight'),
+  );
+}
+
+const groundstationDataGeoJSON = GeoJSON.parse(
+  (groundstationDataJson as unknown) as GroundstationData,
```
bencpeters, line 38: What's the point of this cast? `parse` doesn't know about our `GroundstationData` type, and we're not actually getting any type-safety since we're just making a cast from `unknown` (TypeScript doesn't add anything in that case).

src/components/MapView/GroundstationLayers/index.tsx
ericboucher: It doesn't have the parse function sadly... the types are more for the GeoJSON format itself, not the lib, which is a bit confusing...

ericboucher: When i try to move it I am getting issues, not sure how I would go about extending the geojson type already in types?

src/utils/layer-utils.ts
```diff
@@ -0,0 +1,3 @@
+export function legendToStops(legend: { value: string; color: string }[] = []) {
```
bencpeters, line 1: With this project, I'd say it's still up for debate whether it needs a global `utils` folder - I think until there end up being a lot of cross-dependencies, my preference is to keep the code close to where its used in mostly self-contained modules. To that end, I feel like this like file maybe should live in `MapView`?

PR #52, opened 2020-05-19
MapTooltip popup for Mongolia map and data

### LINKED ISSUE #35: Add tooltip on click
Popup with data details
When a user chooses a layer and then clicks on a polygon (administrative area), they should see a popup with details for the layer and that area. There are two types of data and requirements for this interaction:
- Data already in the admin boundary GEOJSON, usually "baseline" data
- Raster data, such as climate data derived from the Mongolia Data Cube (MDC)

[
Screen Shot 2020-05-05 at 2 42 50 PM
](url)

We can take inspiration from https://github.com/bayesimpact/encompass/tree/master/frontend/src/components/MapTooltip

More info is available here:
https://docs.google.com/document/d/1uD3S33TkKVUPRqkx2hQP65Nl6dOHBYenP1rroZHHbC4/edit

cindyzhang12:
- created generic MapTooltip popup according to Google doc design 
- MapTooltip function gets coordinates and location from fillOnClick functions in Boundaries.js, and MapView passes that data as props into MapTooltip to display on popup

# Todo
- [ ] raster layers should not refresh on every click

Screen Shot 2020-05-15 at 10 34 27 PM

Fixes #35


src/components/MapView/index.tsx
```diff
@@ -41,6 +46,11 @@ function MapView({ classes }: MapViewProps) {
     map: { latitude, longitude, zoom },
   } = appConfig;
 
+  useEffect(() => {
+    setpopupData({});
+    setshowPopup(false);
+  }, [layers]);
+
```
bencpeters, line 53: All this state living here is code smell. This should probably all live in Redux, we don't generally want to combine a bunch of component state with redux state. If this lived in Redux, we wouldn't need an effect listener here - you could just hide the popup by hand when the layer change events get fired.

PR #60, opened 2020-05-26
Add pointer cursor over admin boundaries

### LINKED ISSUE #59: Use a pointer cursor over the country
When the cursor is over an admin boundary, we should change it to a pointer.

This example is a good starting point on how to implement this with our current library:

https://github.com/alex3165/react-mapbox-gl/blob/master/example/src/demos/londonCycle.tsx

ericboucher:
Fixes #59 and switch the cursor to a pointer when it is over any admin boundary.

(More useful after we merge the MapTooltips)

### DIFF
--- a/src/components/MapView/Boundaries/index.tsx
+++ b/src/components/MapView/Boundaries/index.tsx
@@ -26,18 +26,29 @@ function getAdminData(evt: any) {
   );
 }
 
+function onToggleHover(cursor: string, map: any) {
+  console.log(map);
+  // eslint-disable-next-line
+  map.getCanvas().style.cursor = cursor;
+}
+
 const linePaint: MapboxGL.LinePaint = {
   'line-color': 'grey',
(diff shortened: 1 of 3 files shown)


src/components/MapView/Boundaries/index.tsx
```diff
@@ -26,6 +26,11 @@ function getAdminData(evt: any) {
   );
 }
 
+function onToggleHover(cursor: string, targetMap: MapboxGL.Map) {
+  // eslint-disable-next-line
```
bencpeters, line 30: This is very rarely what you want to do (a blanket eslint disable. At most you generally want to disable a specific rule, if there's a good reason for it).

My guess is that in this case you're running into `no-param-reassign` or one of the `fp` rules? Which rule is this for?

ericboucher: both ;)

PR #65, opened 2020-05-26
Order layers below admin boundaries

ericboucher:
It seems to work locally but I am not sure how to best test this. Ideas @bencpeters ?


ericboucher:
Following the advice from https://github.com/alex3165/react-mapbox-gl/issues/35

PR #66, opened 2020-05-26
Only allow one layer of each type at a time

### LINKED ISSUE #28: Limit to one "has_date" layer at a time
As a user, I should only be able to select ONE layer with dates at a time.

ericboucher:
Fixes #28

### DIFF
--- a/src/context/mapStateSlice.ts
+++ b/src/context/mapStateSlice.ts
@@ -35,10 +35,7 @@ const initialState: MapState = {
 
 function keepLayer(layer: LayerType, payload: LayerType) {
   // Simple function to control which layers can overlap.
-  if (payload.type === layer.type) {
-    return false;
-  }
-  return layer.id !== payload.id;
+  return payload.type !== layer.type;
 }
 
 export const mapStateSlice = createSlice({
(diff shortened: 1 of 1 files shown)


ericboucher:
I could do

ericboucher:
We could do

```
return payload.type !== layer.type
```

But I wanted to make it clear that this function could evolve and become a bit more complex

We could also want to prevent baseline + impact (i.e. vector + vector) and any form of raster+raster

src/context/mapStateSlice.ts
```diff
@@ -33,13 +33,21 @@ const initialState: MapState = {
   layersData: [],
 };
 
+function keepLayer(layer: LayerType, payload: LayerType) {
+  // Simple function to control which layers can overlap.
+  if (payload.type === layer.type) {
+    return false;
+  }
+  return layer.id !== payload.id;
```
bencpeters, line 41: This is true by definition if the first condition is met (if we reach this line, it implies `payload.type !== layer.type`, which by definition means that `id` isn't equal, unless we have duplicate ids, which isn't allowed). I'd just simplify this logic to `return payload.type !== layer.type` (and probably put it inline for now, until we add more complicated logic.

PR #67, opened 2020-05-27
Add Tooltip with reducer

ericboucher:


src/context/tooltipStateSlice.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export interface MapTooltipState {
+  coordinates?: GeoJSON.Position;
+  locationName: string;
+  data: PopupData;
+  showing: boolean;
+}
+
+const initialState: MapTooltipState = {
+  coordinates: undefined,
```
bencpeters, line 17: Not a big deal, but if this is literally `undefined`, you don't have to include it. Typescript will yell at you if you forget to include a required initial value in here.

PR #73, opened 2020-05-30
Groundstation api

ericboucher:


src/config/layers.json
```diff
@@ -329,7 +329,8 @@
   "snow_groundstations": {
     "title": "Temperature (example data for 1-11 January 2020)",
     "type": "groundstation",
-    "data": "groundstationSnow",
+    "data": "http://119.40.103.16/temp/Temperature?beginDateTime=2020-03-01&endDateTime=2020-03-21",
```
mamadOuologuem, line 332: Should `data` contains `beginDateTime` and `endDateTime` params? Or should them be set is the app with data selector

src/components/DataDrawer/Chart/index.tsx
```diff
@@ -13,7 +13,7 @@ type ChartProps = {
 
 function colorShuffle(colors: string[]) {
   return colors.map((_, i) =>
-    (-1) ** i !== -1 ? colors[i] : colors[colors.length - i],
+    (-1) ** i !== -1 ? colors[i] : colors[colors.length - i - 1],
```
mamadOuologuem, line 16: is this `(-1) ** i !== -1` use to check the parity of `i`?

I commonly saw `i % 2` for that (return `0` for even and `1` for odd)

src/config/layers.json
ericboucher: We'll keep it like this for now, if they want to improve it later, they can

src/components/DataDrawer/Chart/index.tsx
ericboucher: yup that's better

PR #75, opened 2020-06-04
Recursively flatten WMS layers

ericboucher:
PRISM is capable of fetching WMS data from multiple sources, including Geoserver and Open Data Cube. Different ways to set up layers lead to `Layers` being stored at different depths.

To make PRISM more flexible, we recursively look for Layers until we find them.


src/utils/server-utils.ts
```diff
@@ -52,6 +52,28 @@ function formatCapabilitiesInfo(
...
+    [key: string]: any;
+  };
+  Dimension: {
+    _text: string;
+    [key: string]: any;
+  };
+  [key: string]: any;
+};
+
+function flattenLayers(rawLayers: any): flatLayer[] {
```
bencpeters, line 67: Would this work?

```typescript
type LayerContainer = { Layer: flatLayer[] | LayerContainer }
function flattenLayers(rawLayers: LayerContainer): flatLayer[] {
  if (Array.isArray(rawLayers.Layer)) {
    return rawLayers.Layers.map(({ Layer }) => Layer);
  } else {
    return flattenLayers(rawLayers.Layer);
  }
}
```

src/utils/server-utils.ts
```diff
@@ -52,6 +52,46 @@ function formatCapabilitiesInfo(
...
+  | FlatLayer[];
+
+const isArrayOfFlatLayerContainers = (
+  maybeArray: LayerContainer,
+): maybeArray is FlatLayerContainer[] => {
+  return (maybeArray as FlatLayerContainer[])[0].Layer !== undefined;
+};
+
+function flattenLayers(rawLayers: LayerContainer): FlatLayer[] {
+  if ('Layer' in rawLayers) {
```
bencpeters, line 80: Generally, we don't like using `in` to check for properties in Javascript, because it actually checks for properties that exist anywhere on this object's prototype chain, not just this object itself. It may well not matter in this particular case, but it's not considered good practice.

The "right" way to do this is to use `hasOwnProperty` (could either be: `rawLayers.hasOwnProperty('Layer')` or event "better": `Object.prototype.hasOwnProperty.call(rawLayers, 'Layer')` (the latter is preferred because if an object is created in a certain way, e.g. `const myObj = Object.create(null)`, `hasOwnProperty` might not be defined on it - again probably doesn't matter in this case, but it's the "right" way to do it).

You can also use `lodash#has` for this, which takes care of doing this check the "right" way in an easier-to-remember function call. JS is awesome ;-).

src/utils/server-utils.ts
```diff
@@ -52,6 +52,46 @@ function formatCapabilitiesInfo(
...
+  if (!Array.isArray(rawLayers) || rawLayers.length === 0) {
+    return [];
+  }
+  if (isArrayOfFlatLayerContainers(rawLayers)) {
+    return rawLayers.reduce(
+      (acc, { Layer }) => acc.concat(Layer),
+      [] as FlatLayer[],
+    );
+  }
+  return rawLayers as FlatLayer[];
```
bencpeters, line 92: I'm confused by this cast. You're basically telling Typescript to assume that `rawLayers` is a `FlatLayer[]`. But the point of `isArrayOfFlatLayerContainers` is to test whether `rawLayers` is `FlatLayer[]`. So how do you imagine this condition ever being reached? It seems like almost by definition, if this logic branch is reached, `rawLayers` is _not_ a `FlatLayer[]` (since it failed the `isArrayOfFlatLayerContainers` test?)

ericboucher: If we are ever reaching this point, then we're pretty safe to assume that it is not an array of FlatLayerContainer.

Basically the first `if` takes care of the depth of `Layers` and the other two take care of the difference between the way geoserver layers are stored and how its done in Open Data Cube

bencpeters: Ah yes. I think my confusion is that I was thinking `isArrayOfFlatLayerContainers` was checking for `FlatLayer[]` rather than `FlatLayerContainer[]`.

Just curious - does TS still require you to cast this last line if you wrap it in an `else` block from the `isArrayOfFlatLayerContainers`? It _seems_ like the logic of the union type shouldn't require the cast, but TS might not be smart enough to realize that...

ericboucher: It does not but I wanted to make it a bit clearer :)

PR #76, opened 2020-06-10
Link to the groundstation API for MNG

ericboucher:


src/context/layers/groundstation.ts
```diff
@@ -14,7 +14,6 @@ declare module 'geojson' {
 
 export type GroundstationLayerData = {
   rasterheight: number;
```
dubwalla: 'rasterheight' is not present in all ground station data

PR #79, opened 2020-06-12
Updated App layout to render Routes into a full screen component

### LINKED ISSUE #62: Remove vertical scroll from main map container
Right now there's a slight vertical scroll for the main map container. This is because the map view is sized using `100vh`, which doesn't account for the height of the navbar.

We should redo the structure of the page to make the the main wrapper a column-oriented flex container, with the navbar taking up its fixed height and the `main` container growing to fill the rest of the available space.

bencpeters:
# Updated App layout to render Routes into a full screen component

This works by making sure `html` & `body` & app parent are `height: 100%`, having `display: flex` on the app parent, then using flex grow to take up the available space on the main app display.

Fixes #62


SimplyBallistic:
Works great for me :+1:

SimplyBallistic:
It also looks like triggering a loading screen causes the bars to come back:
image

src/components/MapView/index.tsx
```diff
@@ -133,7 +132,7 @@ function MapView({ classes }: MapViewProps) {
 const styles = () =>
   createStyles({
     container: {
-      position: 'relative',
+      height: '100%',
```
SimplyBallistic, line 135:
```suggestion
      height: '100%',
      position: 'relative',
```
I believe both issues will be solved by re-introducing `position: 'relative'` so both the loading screen and legend are positioned from the non-static map container instead of body.

ericboucher:
> Works great for me 👍
> 
> There is one slight issue however:
> image
> The legend button seems out of place now. Possible to shift it down a little?

Same for me

PR #80, opened 2020-06-13
Add a Flask api to run zonal statistics

ericboucher:
Create a simple Flask API to run zonal statistics for PRISM.

The API has one endpoint for now `/stats` which takes as inputs through POST:
- geotiff_url, the link to a geotiff
- zones_url, the link to a geojson with admin boundaries
- ?group_by, a key to use to group zones in the geojson
- ?geojson_out, decide if the output should be a geojson or a list of data. Default is false -> List.

And a `/demo` endpoint which accepts a simple `GET` call and runs a sample raster / admin boundaries combination.


api-flask/app/caching.py
```diff
@@ -0,0 +1,49 @@
...
+@timed
+def cache_file(url, prefix):
+    """Locally cache files fetched from a url."""
+    cache_filepath = _get_cached_filepath(
+        prefix=prefix,
+        url=url,
+    )
+    # If the file exists, return path.
+    if os.path.isfile(cache_filepath):
+        logger.warning('Returning cached file for {}.'.format(url))
```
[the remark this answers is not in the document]
ericboucher: Yeah I was/am running into some issues with what loglevels I get to show/not show

ericboucher:
For SSL, this part of the doc could be handy:
https://github.com/tiangolo/uwsgi-nginx-flask-docker#overriding-nginx-configuration-completely

api-flask/app/nginx.conf
```diff
@@ -0,0 +1,59 @@
...
+uwsgi_read_timeout          2000;
+uwsgi_send_timeout          2000;
+
+# Avoid 413s for large POST bodies.
+client_max_body_size        30m;
+
+# Http server settings.
+server {
+
+    listen 80;
```
bencpeters, line 22: Probably want to [use nginx to redirect http -> https](https://serversforhackers.com/c/redirect-http-to-https-nginx) and listen for https connections.

api-flask/app/nginx.conf
```diff
@@ -0,0 +1,59 @@
...
+          #
+          add_header "Access-Control-Allow-Headers" "DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Content-Range,Range";
+          #
+          # Tell client that this pre-flight info is valid for 20 days
+          #
+          add_header "Access-Control-Max-Age" 1728000;
+          add_header "Content-Type" "text/plain; charset=utf-8";
+          add_header "Content-Length" 0;
+          return 204;
+       }
```
bencpeters, line 41: Isn't the point of adding a CORS configuration in the Flask app to let it handle the preflight requests? I can see why it might be nice to have the proxy do it without involving the app, but it seems like this route only applies to `/`, not `/stats`, which is where the actual requests are going to get sent. Do we need this definition here at all?

api-flask/docker-compose.override.yml
```diff
@@ -5,6 +5,7 @@ services:
       - ./app:/app
     ports:
       - "80:80"
+      - "443:80"
```
bencpeters, line 8: I don't think this is useful? You don't want to run HTTPS in development anyway, and the complicated part of HTTPS isn't the port forwarding, it's the need to respond with an SSL certificate and encrypt the traffic. This doesn't actually achieve anything.

api-flask/app/caching.py
```diff
@@ -21,7 +21,7 @@ def cache_file(url, prefix):
     )
     # If the file exists, return path.
     if os.path.isfile(cache_filepath):
-        logger.warning('Returning cached file for {}.'.format(url))
+        logger.info('Returning cached file for {}.'.format(url))
```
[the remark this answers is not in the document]
ericboucher: `rasterio` is being VERY VERY noisy in debug, so that's why I decided to up the level of this to make them useful. The better alternative would be to disable debug logging for rasterio but I tried a few things without luck...

api-flask/app/nginx.conf
ericboucher: Sorry the ngninx confi wasnt supposed to be part of the PR. I just wanted to show how you would go about adding one. In case we need to play with it for https

api-flask/app/main.py
```diff
@@ -0,0 +1,96 @@
...
+        geojson_out=geojson_out
+    )
+
+    # TODO - Properly encode before returning. Mongolian characters are returned as hex.
+    return jsonify(features)
+
+
+if __name__ == '__main__':
+    # Only for debugging while developing
+    app.run(host='0.0.0.0', debug=True, port=80)
```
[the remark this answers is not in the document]
ericboucher: Which part?

PR #83, opened 2020-06-18
Connect the frontend to the API

ericboucher:
Connecting the frontend to the new API endpoints.

In layers.json, impact layers now accept an optional `api` config:

```
"api": {
      "url": "https://prism-api.ovio.org/stats",
      "zones_url": "https://prism-admin-boundaries.s3.us-east-2.amazonaws.com/mng_admin_boundaries.json",
      "group_by": "ADM2_PCODE"
    },
```

If this `api` config is available and complete, zonal statistics will be calculated by the API. Otherwise, the frontend will default to running calculations on the client.


src/config/utils.ts
```diff
@@ -42,7 +43,10 @@ const getLayerByKey = (layerKey: layerKeys): LayerType => {
       return throwInvalidLayer();
     case 'impact':
       if (checkRequiredKeys(ImpactLayerProps, definition, true)) {
-        return definition;
+        const apiConfig = definition.api
+          ? (mapKeys(definition.api, (v, k) => camelCase(k)) as StatsApi)
```
SimplyBallistic, line 47: I'm still getting my ropes around this segment of the codebase, correct me if I'm wrong!

Since we generally want to convert every key in the config from `snake_case` to `camelCase`, would it be more worthwhile to make [getLayerByKey](https://github.com/oviohub/prism-frontend/blob/master/src/config/utils.ts#L20) convert keys at a deeper (recursive) level instead of doing it once here?

src/context/layers/impact.ts
```diff
@@ -28,6 +28,20 @@ export type ImpactLayerData = {
   impactFeatures: FeatureCollection;
 };
 
+const fetchApiData = async (url: string, apiData: any) =>
```
SimplyBallistic, line 31: Should we type `apiData`, here and on line 196?

src/context/layers/impact.ts
```diff
@@ -113,38 +127,39 @@ function getBaselineDataForFeature(
   );
 }
 
+const mergeFeaturesByProperty = (
+  baselineFeatures: {}[],
```
SimplyBallistic, line 131:
```suggestion
  baselineFeatures: Array<object>,
```
I think this works better for readability, for me at least. `{}[]` confused me at first haha, I normally use the `[]` after a explicitly defined type (e.g `LayerType[]`). If you agree, make sure to change the line right below this one as well.

of course, typing the object itself, even with something simple would help a ton here!

SimplyBallistic:
Reminder: After merging #81, we can't load adminBoundaries via direct import anymore. Before merging to master, make sure to merge master here.

src/context/layers/impact.ts
bencpeters: Yes ;-)

api-flask/app/main.py
```diff
@@ -44,15 +44,19 @@ def _calculate_stats(zones, geotiff, stats, prefix, group_by, geojson_out):
...
+    zones_url = data.get('zones_url', None)
     if not (geotiff_url and zones_url):
-        raise Exception('geotiff_url and zones_url are both required.')
-
-    geojson_out = request.form.get('geojson_out', 'False')
-    geojson_out = strtobool(geojson_out)
+        logger.error('Received {}'.format(data))
+        return Response(
+            response='501: geotiff_url and zones_url are both required.',
+            status=501
```
bencpeters, line 55: This isn't a 5xx response (server issue), it's a 4xx (client issue). I'd probably either use 400 (general malformed client request) or 422 (unprocessable entity)

src/config/utils.ts
bencpeters: We could definitely do that. Alternatively, we could add a helper function to both check the StatsApi object and camelcase the keys with a typeguard:

Something like this:

```typescript
function parseStatsApiConfig(maybeConfig: { [key: string]: any }): StatsApi {
  const config = mapKeys(maybeConfig, (v, k) => camelCase(k))
  // check all required keys here
  return config as StatsApi
} 
```

src/context/layers/impact.ts
bencpeters: Ya, I agree with both of these. I think in this case, `baselineFeatures` can _probably_ be typed as a `FeatureCollection`, although I haven't tried that.

(Might well work for `aggregateData` too?)

src/context/layers/impact.ts
```diff
@@ -113,38 +127,39 @@ function getBaselineDataForFeature(
   );
 }
 
+const mergeFeaturesByProperty = (
```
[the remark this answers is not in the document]
bencpeters: I actually kinda disagree (at least until the files get _really_ long, like > 700 LOC or so). I think of this file as an already-specific implementation (vs. the more general `layer-data.ts`), and I feel like requiring a yet a third file for getting context on what's happening here wouldn't be super clarifying. But as you say, totally personal preference!

src/context/layers/impact.ts
bencpeters: I think we should also type the response here - we know what the data from the API will look like.

src/context/layers/impact.ts
```diff
@@ -168,76 +183,134 @@ export async function fetchImpactLayerData(
...
-    [] as {
-      id: string;
-      baseline: BaselineRecord;
-      feature: GeoJsonBoundary;
-      bounds: ReturnType<typeof bbox>;
-    }[],
-  );
+  // eslint-disable-next-line fp/no-mutation
+  let activeFeatures: GeoJsonBoundary[];
+  if (layer.api) {
```
bencpeters, line 188: I feel like giant `if` statements like this generally make the code more difficult to follow/reason about (which is another reason that `fp/no-mutation` exists). I'd break this out into separate utility functions:

```typescript
const activeFeatures = layer.api ? loadFeaturesFromApi(layer, ...) : loadFeaturesFromClient(...)
```

src/context/layers/impact.ts
```diff
@@ -168,76 +183,134 @@ export async function fetchImpactLayerData(
...
-  const operation = layer.operation || 'median';
+    // eslint-disable-next-line fp/no-mutation
+    activeFeatures = mergedFeatures.filter(feature => {
+      const scaled = scaleValueIfDefined(
+        feature.properties.stats_median,
+        scale,
+        offset,
+      );
+      return thresholdOrNaN(scaled, layer.threshold);
+    }) as GeoJsonBoundary[];
```
bencpeters, line 220: You shouldn't need to cast this here - `mergeFeaturesByProperty` should have a typed return value.

src/context/layers/impact.ts
```diff
@@ -168,76 +183,134 @@ export async function fetchImpactLayerData(
...
+      await dispatch(
+        loadLayerData({
+          layer: hazardLayerDef,
+          extent,
+          date,
+        } as LayerDataParams<WMSLayerProps>),
+      );
+    }
 
+    // eslint-disable-next-line fp/no-mutation
```
bencpeters, line 237: Why is this `eslint-disable` necessary? It seems like it should be ok with the next line?

src/context/layers/impact.ts
```diff
@@ -113,38 +127,39 @@ function getBaselineDataForFeature(
...
 
-  if (!existingHazardLayer) {
-    const hazardLayerDef = LayerDefinitions[layer.hazardLayer];
-    await dispatch(
-      loadLayerData({ layer: hazardLayerDef, extent, date } as LayerDataParams<
-        WMSLayerProps
-      >),
-    );
-  }
+  const hazardLayerDef = LayerDefinitions[layer.hazardLayer] as any;
```
bencpeters, line 158: Why is this getting typed as `any`? We should be able to type it as a specific layer type...

src/context/layers/impact.ts
ericboucher: aggregate data is just the properties at this point, to keep the api payload smaller

src/context/layers/impact.ts
ericboucher: Still a WIP ;)

Basically, the way the layers are typed atm, hazardLayerDef does not know that it has a wcsConfig object and we don't really have the proper type readily available.

src/config/types.ts
```diff
@@ -186,12 +192,16 @@ export class ImpactLayerProps extends CommonLayerProps {
   legendText: string;
 
   hazardLayer: string;
+
```
ericboucher, line 195:
```suggestion
```

PR #87, opened 2020-06-22
Add Notification System for Errors

### LINKED ISSUE #27: Error handling for layers
Handle errors when fetching data.

In particular, have the ability to parse time mismatch "Time dimension value '2020-02-10' not valid for this layer".

And server errors (eg. server offline)

SimplyBallistic:
image

image

Closes #27 

While working on a system to notify the user when an error happens, I figured it would be better to just create an entirely new system to allow any component to create a notification to display to the user.

Of course, only errors are being displayed through a middleware, but in the future this system can easily be leveraged to display notifications of any type to the user.

A new `notificationStateSlice` was created to store these notifications and the reducer `addNotification` helps to create notifications that display immediately on dispatch.

To test:

Use Redux DevTools and dispatch the following action:
```js
{
type: 'mapState/loadLayerData/rejected', error: 'Failed!'
}
```
Or, trigger a fail by corrupting a layer in `layers.json` and loading the layer.

You can also dispatch a generic notification:
```js
{
type: 'notificationState/addNotification',payload:{message:'Hello World!',type:'info'}
}
```
**QOL Changes**
- Removed dependency `@material-ui/icons` as its not used anywhere. Added `@material-ui/lab` for Alert support - not included in core.


SimplyBallistic:
My bad, I extended this branch from the already merged #81 branch therefore the commit history is a little outdated. New commits start at https://github.com/oviohub/prism-frontend/pull/87/commits/c805536fb203b7f309eb1e510586586ffb7934e1

src/components/MapView/DateSelector/index.test.tsx
```diff
@@ -5,6 +5,8 @@ import DateSelector from '.';
 
 import { store } from '../../../context/store';
 
+jest.mock('../../Notifier', () => 'mock-Notifier');
```
laveesh, line 8: Shouldn't we write this test in `index.test.tsx` in `Notifer` folder instead of here in the `DateSelctor` tests

ericboucher: This is actually mocking the Notifier to NOT test it in here.

ericboucher:
@SimplyBallistic can you add a screenshot of what the implemented errors look like at the moment?

SimplyBallistic:
> 
> 
> @SimplyBallistic can you add a screenshot of what the implemented errors look like at the moment?

https://drive.google.com/file/d/1I46dTYpGeWItzOB2AGy1h_tuA7Kx_JIQ/view
This is a quick example of how the error popup might work. Errors have been moved to the console since I realized they take quite a bit of room on the page. Let me know what you think of this approach. 

Also, I realised if a 404 error occurs with `fetch`, it doesn't reject the promise but rather completes it with status 200. This makes it hard to catch such errors and are instead represented by a JSON parse error. Should we leave it as it is or find a way to improve the error reporting?

ericboucher:
Thanks for the video, looks great!

If you think the 404 issue is an easy fix, go for it, otherwise I'd recommend opening an issue and coming back to it later

SimplyBallistic:
> 
> 
> Thanks for the video, looks great!
> 
> If you think the 404 issue is an easy fix, go for it, otherwise I'd recommend opening an issue and coming back to it later

One way to solve it is to replace every usage of `fetch` with a bloated alternative that throws an error manually when it detects something inline with a 404 (fetch doesn't even tell us its a 404 so we need to look at other things like the body content to see if its empty)

Or, to make things cleaner we could use regex or something similar on the error string to see if the error is similar to what JSON would normally throw when its asked to parse an empty string. This can be directly put in the middleware.

Both methods aren't that great at solving the issue as they are essentially just parsing strings which makes them susceptible to small changes in the underlying browser APIs. Another possible alternative is to introduce `axios`, which properly handles these cases, but it does mean we are adding another package. 

Should I try implement one of these methods in this PR? Or make a new issue to address this? Not really a big issue as a 404 won't happen too often and depending on the person reading the error, they *most* likely will work out whats happening.

ericboucher:
Let's open a new issues for this.

`fetch` also has a **.catch** that we could use to kickoff the error bars, eg. in `checkValidServiceWorker`

bencpeters:
> One way to solve it is to replace every usage of `fetch` with a bloated alternative that throws an error manually when it detects something inline with a 404 (fetch doesn't even tell us its a 404 so we need to look at other things like the body content to see if its empty)

@SimplyBallistic I think you can use [`Response.ok`](https://developer.mozilla.org/en-US/docs/Web/API/Response) to check if the request is a 2xx code or something else (even though the promise won't be rejected - but you can than throw an error after the check). I would think `Response.status` would also be 404, but maybe that's not true?

src/context/notificationStateSlice.ts
```diff
@@ -0,0 +1,107 @@
...
+  removeNotification,
+  setAsDisplayed,
+} = notificationStateSlice.actions;
+
+// middleware to add error as notifications to this slice
+// I don't know how to type this properly
+export const errorToNotificationMiddleware: Middleware<{}, RootState> = () => (
+  dispatch: AppDispatch,
+) => (action: AnyAction) => {
+  const dispatchResult = dispatch(action);
```
bencpeters, line 87: I don't think this applies currently, but should we wrap this in a `try...catch` to allow us to handle synchronous reducer errors too?

package.json
```diff
@@ -80,6 +80,7 @@
     "lodash": "^4.17.15",
     "mapbox-gl": "^1.8.1",
     "moment": "^2.24.0",
+    "notistack": "^0.9.17",
```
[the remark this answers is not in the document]
SimplyBallistic: > Is there a reason you strongly prefer this library over [Material-UI Snackbars](https://material-ui.com/components/snackbars/)?
> 
> I'm just asking, because in general I'd like to minimize the number of external dependencies. But if there's a strong reason you prefer this library, it's not a problem!


I added the library since at the bottom of the Material UI Snack bar docs, they recommended using notistack for advanced usecases (stacking notifications), which is really the only big selling point of notistack. If having just one at a time is fine, we could definitely move to just using Material UI. Do you think stacking is worth?

SimplyBallistic:
> This looks like a good start, but I'm not sold on using this notifications package. It seems like we basically have duplicate state with it, since we have to enqueue/dismiss within the package's own internal state as well as keep our Redux version of state. I wonder if we could eliminate a dependency as well as simplify the logic by just using Material UI Snackbars rendered by `Notifier` and only using the redux state as our source of truth?

Yes! I was thinking the same too when I was making it but since they recommended this implementation in their docs (https://iamhosseindhv.com/notistack/demos#redux-/-mobx-example) I just went with it. It comes down to if we need stacking. Let me know with a final consensus and I'll switch over to Material UI snack bars 👍

SimplyBallistic:
> > One way to solve it is to replace every usage of `fetch` with a bloated alternative that throws an error manually when it detects something inline with a 404 (fetch doesn't even tell us its a 404 so we need to look at other things like the body content to see if its empty)
> 
> @SimplyBallistic I think you can use [`Response.ok`](https://developer.mozilla.org/en-US/docs/Web/API/Response) to check if the request is a 2xx code or something else (even though the promise won't be rejected - but you can than throw an error after the check). I would think `Response.status` would also be 404, but maybe that's not true?

Using Firefox, I came to this conclusion because console logging the fetch object and looking at its status in the Network tab always displayed the status as 200, even when I messed with the URL in `layers.json` to point to a nonexistent location. I'll try testing with chrome - it might be a bug in my browser?

src/context/notificationStateSlice.ts
SimplyBallistic: Yup! That's a good idea. I'll do that next commit :+1:

src/context/notificationStateSlice.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  // eslint-disable-next-line default-case
+  switch (action.type) {
```
SimplyBallistic, line 90: It's a little unusual that I decided to use a switch statement here. I'm going to 'switch' it to:
```js
if(['x','y'].includes(...))
```

SimplyBallistic:
> Using Firefox, I came to this conclusion because console logging the fetch object and looking at its status in the Network tab always displayed the status as 200, even when I messed with the URL in `layers.json` to point to a nonexistent location. I'll try testing with chrome - it might be a bug in my browser?

image
Same with Edge too, I'll make an issue for it like Eric stated - not high priority for this PR.





> Let's open a new issues for this.
> 
> `fetch` also has a **.catch** that we could use to kickoff the error bars, eg. in `checkValidServiceWorker`

How would I dispatch a Redux action from `checkValidServiceWorker`?

bencpeters:
> Using Firefox, I came to this conclusion because console logging the fetch object and looking at its status in the Network tab always displayed the status as 200, even when I messed with the URL in `layers.json` to point to a nonexistent location. I'll try testing with chrome - it might be a bug in my browser?

Hrm, that's odd. Possible it's a bug, although I feel like Firefox is generally the most standards-compliant of all the browsers.

At any rate, we definitely want to support Firefox, so if it doesn't work there, then we shouldn't rely on it.

src/components/Notifier/index.tsx
```diff
@@ -1,63 +1,65 @@
...
 
-// built with the help of https://codesandbox.io/s/github/iamhosseindhv/notistack/tree/master/examples/redux-example?file=/Notifier.js
 function Notifier({ classes }: NotifierProps) {
   const dispatch = useDispatch();
-  const { enqueueSnackbar, closeSnackbar } = useSnackbar();
   const notifications = useSelector(notificationsSelector);
   const autoHideDuration = 50000;
+  const [openNotification, setOpenNotification] = useState<Notification | null>(
+    null,
+  );
```
bencpeters, line 23: Why do we need to keep track of `openNotification` here in local state? Why can't we just have this component pull the array of notifications from Redux and render them (or render `null` if the array is empty). Then we can still get stacking easily, and just dispatch any dismiss actions.

src/context/notificationStateSlice.ts
```diff
@@ -84,23 +84,39 @@ export const {
...
+    dispatchResult = dispatch(action);
+  } catch (err) {
+    console.error(err);
+    dispatch(
+      addNotification({
+        type: 'error',
+        message: err.message || err,
+      }),
+    );
+    return dispatchResult;
```
bencpeters, line 100: This isn't going to be defined if `dispatch` threw an error. We probably want to just re-throw the error here, or possibly return `undefined` explicitly - I find it a bit misleading to return `dispatchResult` though.

src/context/notificationStateSlice.ts
```diff
@@ -84,23 +84,39 @@ export const {
...
   }
+  const errorActions = [
+    // async thunks that have errors included with them
+    'tableState/loadTable/rejected',
+    'serverState/loadAvailableDates/rejected',
+    'mapState/loadLayerData/rejected',
+  ];
+  if (errorActions.includes(action.type)) {
```
bencpeters, line 108: Maybe just do a regex on this so we don't have to remember to enumerate these everytime we add an async thunk?

SimplyBallistic:
Notes:
- We added a Notification system to build on, but there may be places in code that aren't built to report their errors effectively. E.g `checkValidServiceWorker` which may need discussion on implementation due to its complexity
- Error messages for 404 errors via fetch aren't easy to implement through my own research

We can move these to an issue and try solve these in the future, or possibly solve them now. Let me know your thoughts :+1:

Stacking isn't easy without notistack:
image
as MaterialUI uses `position: fixed` and notifications can be varying heights.

I think the best way to solve this is to disable stacking and only show the latest notification, with dismissing used to cycle to the next most recent notification. Let me know your thoughts or if you have a solution :+1:

cc @bencpeters

bencpeters:
> as MaterialUI uses `position: fixed` and notifications can be varying heights.
> 
> I think the best way to solve this is to disable stacking and only show the latest notification, with dismissing used to cycle to the next most recent notification. Let me know your thoughts or if you have a solution 👍

Couldn't you just override these styles? Use the base snackbar component, but then apply a few styles on top of it that override the base material UI ones?

bencpeters:
>     * Error messages for 404 errors via fetch aren't easy to implement through my own research, The only solution I see is to parse the result for every `fetch` we use to see if its empty before parsing to JSON, which doesn't seem worth for bettering an error message?

So you're saying that this is what you've found?

```javascript
const res = await fetch("/404path")

// Will be true!?
res.ok
```

Because the [docs](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) pretty clearly say that `res.ok` should be `false` for 404 or 500s.

SimplyBallistic:
> 
> So you're saying that this is what you've found?
> 
> ```js
> const res = await fetch("/404path")
> 
> // Will be true!?
> res.ok
> ```
> 
> Because the [docs](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) pretty clearly say that `res.ok` should be `false` for 404 or 500s.
image

It seems so :confused:
Any idea?
I just directly console logged the `await fetch(path)` from [boundary.ts](https://github.com/oviohub/prism-frontend/blob/a4a605fdc771684ff9929f659efaa4f815894fce/src/context/layers/boundary.ts#L13)

and changed the path of `admin_boundaries` in `layers.json` to `non-existent-file` as shown.

bencpeters:
> It seems so 😕
> Any idea?
> I just directly console logged the `await fetch(path)` from [boundary.ts](https://github.com/oviohub/prism-frontend/blob/a4a605fdc771684ff9929f659efaa4f815894fce/src/context/layers/boundary.ts#L13)

Ah, ok, I see whats happening here. When you're fetching from _the dev server_, it hits a server that has been configured to let React handle all the routing - e.g. by definition it will just respond to everything with a 200 request, the react app files, and let react do it's own thing. So I wouldn't expect you to see a 404 when fetching `http://localhost:3000/data/non-existent-file`, but I _would_ expect to see it when talking to a geoserver and fetching something that doesn't exist.

The other question is what happens in production - I think ideally if we're going to be serving data files from the same server as the static site, we would set up `/data/*` routes to _not_ direct to the React-based 404 page (which actually gives a 200 status code for the request, since it's only in the front end). [Here are the firebase hosting config docs](https://firebase.google.com/docs/hosting/full-config). I think we'd just need to modify our `firebase.json` with a different `rewrite` for `/data/**` paths (right now *all* paths get redirected to `index.html`, we'd want `data/` paths to not have that happen so that they get a 404 status properly).

If you feel like taking a stab at that, you can test it, otherwise let's just add the 404 handling that "makes sense" (e.g. checking `res.status` and/or `res.ok`) and create an issue to keep track of the fact that that error handling won't work for `/data` hosted files.

SimplyBallistic:
Ahh that makes more sense, nice catch! 

I'll find a way to test a 404 response and the error message that follows to make sure notifications work as intended. I might try to deploy to firebase for full prod. testing if I can work out my way around it - never used it before, but @laveesh could take a shot. Otherwise, I'll just set up a url that always returns a 404 for testing purposes.

If everything works, I'll approve this myself, otherwise I'll commit anything else to make sure 404 fetches are handled correctly 👍

SimplyBallistic:
Can't approve own PR, but everything is good on my end. 404 and most status works without too much extra work - `fetch` automatically logs statuses that aren't 2xx

PR #90, opened 2020-06-25
Fix import cycles

ericboucher:
Avoid 

```
\src\components\DataDrawer\Chart\index.tsx
  1:1  error  Resolve error: unable to load resolver "import/parsers"  import/extensions

✖ 1 problem (1 error, 0 warnings)
```


.eslintrc
```diff
@@ -50,29 +70,51 @@
...
     "spaced-comment": "warn"
   },
   "settings": {
     "import/resolver": {
       "node": {
-        "moduleDirectory": ["node_modules", "./src"],
-        "extensions": [".js", ".jsx", ".ts", ".tsx"]
-      },
-      "import/parsers": {
-        "@typescript-eslint/parser": [".ts", ".tsx"]
```
ericboucher, line 75: Removing this extra setting that is breaking the linter on Windows

ericboucher:
An issue in the `.eslintrc` prevented us from running the import/parser properly.
Fixing this bug unearthed a few circular import issues to investigate.

bencpeters:
> @bencpeters thoughts on how to approach solving those?

Generally to solve circular imports you either refactor your logic so that one file depends only on the other instead of having mutual dependencies, or if that's difficult, you extract the mutual dependencies into a new shared dependency that both can import. Hard to say more without specifics though.

.eslintrc
ericboucher: Actually, the indentation in the file was wrong and these had to be under settings instead of being under "import/resolver"

.eslintrc
```diff
@@ -7,12 +7,25 @@
     "prettier/react"
   ],
   "parser": "@typescript-eslint/parser",
-  "plugins": ["jsx-a11y", "fp", "@typescript-eslint", "prettier", "import"],
+  "plugins": [
```
bencpeters, line 10: Can we revert the formatting changes in this file so that we only see the things that actually changed?

src/context/layers/boundary.ts
```diff
@@ -1,4 +1,5 @@
 import { FeatureCollection } from 'geojson';
+// eslint-disable-next-line import/no-cycle
```
bencpeters, line 2: As of Typescript 3.8, TS supports `import type { LayerDataParams } from './layer-data'`. Let's change type-only imports to that syntax instead of disabling eslint and see if eslint still complains in that case? 

[Reference](https://github.com/benmosher/eslint-plugin-import/issues/1453)

src/context/layers/groundstation.ts
```diff
@@ -1,4 +1,5 @@
 import GeoJSON from 'geojson';
+// eslint-disable-next-line import/no-cycle
 import { LayerDataParams } from './layer-data';
```
bencpeters, line 3: Same - try `import type` here.

SimplyBallistic:
Note:
One worthwhile import cycle to fix is with `loadLayerData` and `mapStateSlice`. 

`loadLayerData` depends on `nso` which depends on `mapStateSlice` which depends on `loadLayerData`, meaning in some cases, `mapStateSlice` always gets an undefined version of `loadLayerData` resulting in a failiure while constructing the map slice.

This makes testing difficult and leads to undesirable mocks. Ran into this issue while writing tests for #92

bencpeters:
> This makes testing difficult and leads to undesirable mocks. Ran into this issue while writing tests for #92

Good catch! That's why I don't just want to suppress these eslint warnings - I'd like to use `import type` to figure out what are "just" type dependencies, and try to fix any remaining cyclic dependencies after that...

ericboucher:
@SimplyBallistic would you be able to take over this PR?

SimplyBallistic:
> @SimplyBallistic would you be able to take over this PR?

sure! I'll update here if we face any blockers

Issue #29, opened 2020-04-26
Improve config definition for servers and layers


ericboucher:
We need to centralize information on servers and have a "parameter" approach for layers / types.

SimplyBallistic:
What progress was made in #25? Was it solved?

If not, is it possible to elaborate on what needs to be achieved? Specifically on 

>  centralize information on servers

and 

> "parameter" approach for layers / types

ericboucher:
We are currently using a `base_url` in layers.json for raster servers, combines with a wcs_config if one is available.

Eg:  "base_url": "https://mongolia.sibelius-datacube.org:5000",

Instead, we should allow the creation of "severs" in the prism.json file, and use the server_id in the layers.json instead. This would make it easier to change what server we are pointing too.

The server config for rasters could also have a `wms_path` and an optional `wcs_path`, indicating that WCS capabilities are available.


I think we have started the "parameter" approach, although a lot of the names and functionalities are quite specific for Mongolia.

I was envisioning that we would have broader types to cover more "standard" options.

Eg:
- points_geojson
- vector_geojson
- raster
- impact

dubwalla:
Yes, this is a bit of a priority as the naming convention at present is overly specific to Mongolia. I had on my to do list to create an issue on this specifically.

The examples above (points_geojson, vector_geojson, and raster) are good. Impact is based on a current use case, but in practice the output of an analysis could be 'impact', 'spatial stats', and more. Future analysis outputs could include points and rasters too. Do we need to create separate types for each?

dubwalla:
I also wanted to note that WMS layers are not only rasters. A WMS could also return point or vector data

ericboucher:
For now the consensus is to rename layers to the following:

NSO -> admin_level_data
Groundstation -> point_data

wadhwamatic:
Addressed most of what was required here. New issues will be created to address gaps based on more current structure

PR #94, opened 2020-06-26
Flexible use of admin and data fields for admin level data

### LINKED ISSUE #93: Allow the data field in an NSO data type to be configurable
Currently the NSO data type expects the data attribute to be mapped for an admin area to be called 'DTVAL_CO'. This was specific to Mongolia. To make deployments more efficient, we should allow this to be configurable for each layer, the same way that the admin code is configurable.

ericboucher:
Fixes #93 
Addresses #29 

In particular:

- One data file with multiple data columns will only need to be imported once
- Names of the admin boundaries used in the tooltips can be configured easily in the admin_level_names of the admin boundary layer
- The name of the column with the admin code can be set in the config for both admin boundaries and data layers


dubwalla:
Looks good. One question though. In /BoundaryLayer/index.tsx, the current structure requires further editing to allow display of admin 3 level data. Can we make this more dynamic to access whichever level is specified for the active layer?

PR #102, opened 2020-07-31
Add Date support to Groundstation Layers

SimplyBallistic:
**TODO**
- [ ] Write this description


src/config/layers.json
```diff
@@ -697,7 +697,9 @@
   "avg_temperature_groundstations": {
     "title": "10-day average temperature",
     "type": "groundstation",
-    "data": "https://mng.prism.services/temp/Temperature?beginDateTime=2020-05-01&endDateTime=2020-05-01",
+    "data": "https://mng.prism.services/temp/Temperature",
```
ericboucher, line 700: Let's try to avoid being too specific for this particular API. Can we just use `https://mng.prism.services/temp/Temperature?beginDateTime=2000-01-01&endDateTime=2023-12-21` for now?

src/config/types.ts
```diff
@@ -228,6 +228,11 @@ export class GroundstationLayerProps extends CommonLayerProps {
   measure: string;
   @optional
   fallbackData?: string;
+
+  // the date range available to select from the date picker TODO api should just have a way to supply all possible dates
+  // in format YYYY-MM-DD where DD must be 01, 11 or 21. MM for endDate should be 12 to get max amount of dates
+  beginDate: string;
```
ericboucher, line 234: I don't think this should be part of the data type

src/context/layers/groundstation.ts
```diff
@@ -15,25 +16,37 @@ declare module 'geojson' {
...
+    data = (await (
+      await fetch(layer.data + dateQuery, { mode: 'cors' })
+    ).json()) as GroundstationLayerData;
   } catch (ignored) {
     // eslint-disable-next-line fp/no-mutation
-    data = await (await fetch(layer.fallbackData || '')).json();
+    data = ((await (
+      await fetch(layer.fallbackData || '')
+    ).json()) as GroundstationLayerData).filter(obj => {
+      // console.log(obj.date, formattedDate, moment(obj.date).valueOf(), date);
```
ericboucher, line 47: 👣

src/context/layers/groundstation.ts
```diff
@@ -15,25 +16,37 @@ declare module 'geojson' {
...
+      return moment(obj.date).valueOf() === moment(formattedDate).valueOf();
```
ericboucher, line 48: wouldn't it be easier to check for `string` equality here?

src/utils/server-utils.ts
```diff
@@ -1,9 +1,14 @@
...
-import { merge, get, isString, union, isEmpty } from 'lodash';
-
+import { get, isEmpty, isString, merge, union } from 'lodash';
 import config from '../config/prism.json';
-import { AvailableDates } from '../config/types';
+import { AvailableDates, GroundstationLayerProps } from '../config/types';
+import { LayerDefinitions } from '../config/utils';
+import { GroundstationLayerData } from '../context/layers/groundstation';
+
+// Note: PRISM's date picker only works with dates where their time is midnight in the UTC timezone (possible dates passed in should be 2020-xx-xx 00:00 Midnight in UTC    -    2020-xx-xx 10:00 in Australia East Coast)
```
ericboucher, line 9: The `midnight` here shouldn't matter. But the UTC should. The short version should be: Dates should be processed in UTC as soon as possible.

src/utils/server-utils.ts
```diff
@@ -111,9 +116,8 @@ async function getWMSCapabilities(serverUri: string) {
 
     return formatCapabilitiesInfo(flatLayers, 'Name._text', 'Dimension._text');
   } catch (error) {
-    throw new Error(
-      `Failed loading dates with request GET: ${requestUri} for WMS Capabilities, ${error}`,
-    );
+    console.error(error);
```
ericboucher, line 119: why this change?

src/utils/server-utils.ts
```diff
@@ -141,24 +145,66 @@ async function getWCSCoverage(serverUri: string) {
...
+  }
+  return data
+    .map(item => moment.utc(item.date).valueOf())
+    .filter((date, index, arr) => {
+      return arr.indexOf(date) === index;
+    }); // filter() here removes duplicate dates because indexOf will always return the first occurrence of an item
+}
+
+/**
+ * Given a WMS or WCS serverUri, return a Map of available dates.
```
ericboucher, line 187: Let rephrase this to:

Load available dates for WMS and WCS using serverUri and for GeoJSONs using their API endpoint.

ericboucher:
Finally, until we can handle multiple layers with dates at the same time, having groundstation aka point_data should become incompatible with other layers that have dates. Otherwise you risk having two layers (WMS and greoundstations) that have different set of available dates

src/context/layers/groundstation.ts
SimplyBallistic: sometimes the api returns `2020-4-01` instead of `2020-04-01`. This helps make sure that despite the inconsistencies in formatting, the date is always recognized correctly. I'll add a comment here to clarify this :+1:

SimplyBallistic:
> Finally, until we can handle multiple layers with dates at the same time, having groundstation aka point_data should become incompatible with other layers that have dates. Otherwise you risk having two layers (WMS and greoundstations) that have different set of available dates

I was thinking of reworking the `availableDates` array so it only stores dates that are duplicated throughout all the active layers. 
Example if WMS supports 2020-04-01 and 2020-05-01 but groundstation supports only 2020-04-01,
adding them to an array then filtering out dates that aren't duplicated would give dates compatible for both layers. 

Think the above is worthwhile to implement here?

src/utils/server-utils.ts
SimplyBallistic: #87 introduced this error in order to show a notification if the request to the WMS or WCS server failed. However, if we throw an error if one fails even if the others succeed, then dates aren't loaded for anything else such as groundstation layers. Currently the WCS API url is failing: https://github.com/WFP-VAM/prism-frontend/blob/50752fc36395c2b60a51139e3ef04306200aa7c5/src/config/prism.json#L4, meaning nothing else is allowed to load. Removing the error and just printing bypasses this.

This is moreso a temporary solution, what should we do?

ericboucher: That's fine for now then but maybe add a todo?

ericboucher:
> I was thinking of reworking the `availableDates` array so it only stores dates that are duplicated throughout all the active layers.
> Example if WMS supports 2020-04-01 and 2020-05-01 but groundstation supports only 2020-04-01,
> adding them to an array then filtering out dates that aren't duplicated would give dates compatible for both layers.
> 
> Think the above is worthwhile to implement here?

Only if it is super simple

SimplyBallistic:
> > I was thinking of reworking the `availableDates` array so it only stores dates that are duplicated throughout all the active layers.
> > Example if WMS supports 2020-04-01 and 2020-05-01 but groundstation supports only 2020-04-01,
> > adding them to an array then filtering out dates that aren't duplicated would give dates compatible for both layers.
> > Think the above is worthwhile to implement here?
> 
> Only if it is super simple. And we would need a warning snack bar if there are no overlapping dates at all

got it, I'll include it as a quick fix in this PR :+1:

src/components/MapView/index.tsx
```diff
@@ -80,33 +82,75 @@ function MapView({ classes }: MapViewProps) {
...
+     */
+    const ret = Object.keys(
+      pickBy(
+        selectedLayerDatesDupCount,
+        dupTimes => dupTimes >= layersWithDateSupport.length,
+      ),
+    ).map(Number); // convert back to number array - countBy converted all the numbers to strings
+    if (ret.length === 0) {
+      dispatch(
+        addNotification({
```
SimplyBallistic, line 145: Couldn't find a way to test this as all existing date layers have overlap (wms, impact & point_data)

src/context/layers/layer-data.ts
```diff
@@ -18,7 +15,8 @@ type LayerSpecificDataTypes = {
   wms: WMSLayerData;
   impact: ImpactLayerData;
   nso: NSOLayerData;
-  groundstation: GroundstationLayerData;
+  // eslint-disable-next-line camelcase
```
SimplyBallistic, line 18: This is ok?

src/components/MapView/DateSelector/index.tsx
```diff
@@ -177,7 +178,7 @@ const styles = (theme: Theme) =>
   });
 
 export interface DateSelectorProps extends WithStyles<typeof styles> {
-  availableDates: number[];
+  availableDates?: number[];
```
ericboucher, line 181: What happens to `includeDates={availableDates.map(d => new Date(d + userOffset))}` if availableDates is undefined?

ericboucher:
When loading groundstation data for the first time, we should probably update the datePicker "selected" date to match what we are showing. I.e. the latest data?

src/components/MapView/DateSelector/index.tsx
SimplyBallistic: the value is default set to an empty array here:
https://github.com/WFP-VAM/prism-frontend/blob/cad0637bd405c3d4d166d0ca7bc098efcb9711f1/src/components/MapView/DateSelector/index.tsx#L39
The reason I changed the type is because the default array setting was already there, so I updated the type definition to reflect this. Mostly to prevent an error here: 
https://github.com/WFP-VAM/prism-frontend/blame/ca955c5cc5d3088d591b0d05719be0f604116231/src/components/MapView/DateSelector/index.test.tsx#L18
but the Analysis PR #92 seems to have fixed it by just giving an empty array instead.

Which do you think is the best approach?
1. Revert this change and use what was done in #92 (passing empty array in test)
2. Use what was done here and rely on default if availableDates isn't given, and type the props accordingly

SimplyBallistic:
> When loading groundstation data for the first time, we should probably update the datePicker "selected" date to match what we are showing. I.e. the latest data?

Makes sense. This could be extended to all the other date layers such as WMS and Impact - currently loading a WMS layer doesn't hint at what date it is based off & Impact refuses to load until a date is chosen. Think this is a good idea?

src/components/MapView/DateSelector/index.tsx
ericboucher: Ok I think that's propably like that then

ericboucher:
> > When loading groundstation data for the first time, we should probably update the datePicker "selected" date to match what we are showing. I.e. the latest data?
> 
> Makes sense. This could be extended to all the other date layers such as WMS and Impact - currently loading a WMS layer doesn't hint at what date it is based off & Impact refuses to load until a date is chosen. Think this is a good idea?

For "impact" this is on purpose because we didn't want to trigger a long calculation for no reason. i.e. with dates the user is not interested in.

For WMS and Groundstation it might make sense to simply fetch the latest available dates and update the DateSelector for clarity.

src/components/MapView/DateSelector/index.tsx
SimplyBallistic: I'll leave it as it is if that's what you mean 🙂

SimplyBallistic:
> For "impact" this is on purpose because we didn't want to trigger a long calculation for no reason. i.e. with dates the user is not interested in.
> 
> For WMS and Groundstation it might make sense to simply fetch the latest available dates and update the DateSelector for clarity.

Implemented here: https://github.com/WFP-VAM/prism-frontend/pull/102/commits/39510bf937b28ea8c92bed42b24e6c92622134ae

SimplyBallistic:
New route to make this pr obsolete: https://mng.prism.services/temp/getDates

PR #106, opened 2020-08-24
Rename groundstation->point_data

SimplyBallistic:
Works towards solving WFP-VAM#29

This and #103 are the last PRs done within the Fellowship 🎉

### DIFF
--- a/src/context/layers/layer-data.ts
+++ b/src/context/layers/layer-data.ts
@@ -1,15 +1,13 @@
 import { createAsyncThunk, AsyncThunk } from '@reduxjs/toolkit';
 import { DiscriminateUnion, LayerType } from '../../config/types';
 import { Extent } from '../../components/MapView/Layers/raster-utils';
-import { fetchNSOLayerData, NSOLayerData } from './nso';
 
+import { fetchNSOLayerData, NSOLayerData } from './nso';
 import { fetchWCSLayerData, WMSLayerData } from './wms';
-import {
-  fetchGroundstationData,
-  GroundstationLayerData,
-} from './groundstation';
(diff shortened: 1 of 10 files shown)


src/context/layers/layer-data.ts
```diff
@@ -1,23 +1,22 @@
...
+
 import type { CreateAsyncThunkTypes, ThunkApi } from '../store';
 
 type LayerSpecificDataTypes = {
   boundary: BoundaryLayerData;
   wms: WMSLayerData;
   impact: ImpactLayerData;
   nso: NSOLayerData;
-  groundstation: GroundstationLayerData;
+  // eslint-disable-next-line camelcase
```
SimplyBallistic, line 18: this is ok?

src/context/layers/point_data.ts
```diff
@@ -10,21 +10,21 @@ declare module 'geojson' {
...
   [key: string]: any;
 }[];
 
-export const fetchGroundstationData: LazyLoader<GroundstationLayerProps> = () => async ({
+export const fetchPointLayerData: LazyLoader<PointDataLayerProps> = () => async ({
   date,
   layer,
 }) => {
-  // This function fetches groundstation data from the API.
+  // This function fetches point data data from the API.
```
ericboucher, line 27: no need to double the "data" here

src/context/layers/layer-data.ts
ericboucher: I think it's fine for now. The other option would be to parse the definition and update it to transform all keys to camelCase as we are doing in the layer utils.

Ultimately we should probably do a deep transformation of all the keys in the json files right on import.

ericboucher (approved):

PR #107, opened 2020-08-29
Fix useDefaultDate

### LINKED ISSUE #108: Fix issues with useDefaultDate
The new useDefaultDate hook introduced in #102 was not made to be timezone compatible, making it break in GMT timezones with a negative offset. 

Also, the dispatch done within the hook sets off "Cannot update a component from inside the function body of a different component" and could be solved using the following as a guide:
https://stackoverflow.com/questions/60526786/react-warning-cannot-update-a-component-from-inside-the-function-body-of-a-diff

SimplyBallistic:
Closes #108

### DIFF
--- a/src/components/MapView/DateSelector/index.tsx
+++ b/src/components/MapView/DateSelector/index.tsx
@@ -26,6 +26,10 @@ interface InputProps {
   onClick?: () => void;
 }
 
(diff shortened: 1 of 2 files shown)


ericboucher:
Integrated in #109

PR #113, opened 2020-10-02
Replace slow Date fetching with new Groundstation date API

SimplyBallistic:
Replaces #102

Test URL: http://prism-date-groundstation-test.surge.sh/

Possible Areas of Improvement:
- I made each layer have its own date fetching URL just in case dates start differing in the future.  ~This creates a small performance hit because we call the url 4 times instead of just once, but for now, doesn't create a huge performance hit. Make sure to confirm this!~ Fixed with cache

### DIFF
--- a/src/config/types.ts
+++ b/src/config/types.ts
@@ -139,7 +139,6 @@ export class CommonLayerProps {
   type: string;
   opacity: number;
 
-  // TODO not used
   @optional
   dateInterval?: string;
 
@@ -208,7 +207,6 @@ export class StatsApi {
   groupBy: string;
 }
 
-// first is display name, second is name we store in computers
 export enum AggregationOperations {
   Mean = 'mean',
   Median = 'median',
--- a/src/config/types.ts
+++ b/src/config/types.ts
@@ -139,6 +139,7 @@ export class CommonLayerProps {
   type: string;
   opacity: number;
 
+  // TODO not used
   @optional
   dateInterval?: string;
 
@@ -206,6 +207,7 @@ export class StatsApi {
   zonesUrl: string;
   groupBy: string;
 }
+
 // first is display name, second is name we store in computers
 export enum AggregationOperations {
   Mean = 'mean',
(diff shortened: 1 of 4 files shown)


ericboucher:
> I made each layer have its own date fetching URL just in case dates start differing in the future. This creates a small performance hit because we call the url 4 times instead of just once, but for now, doesn't create a huge performance hit. Make sure to confirm this!

I think it's fine that they have independent URLs, but we could have a smart way to collapse them afterwards and only fetch each URL once. Maybe add this as a todo or open an issue?

src/config/types.ts
```diff
@@ -139,6 +139,7 @@ export class CommonLayerProps {
   type: string;
   opacity: number;
 
+  // TODO not used
```
ericboucher, line 142: remove this comment

src/config/types.ts
```diff
@@ -206,6 +207,7 @@ export class StatsApi {
   zonesUrl: string;
   groupBy: string;
 }
+
 // first is display name, second is name we store in computers
```
ericboucher, line 211: let's remove this comment I don't think it's necessary

src/utils/server-utils.ts
```diff
@@ -154,32 +153,23 @@ async function getWCSCoverage(serverUri: string) {
       'domainSet.temporalDomain.gml:timePosition',
     );
   } catch (error) {
-    // TODO we used to throw the error here so a notification appears. Removed because a failure of one shouldn't prevent the successful requests from saving.
+    // TODO we used to throw the error here so a notification appears via middleware. Removed because a failure of one shouldn't prevent the successful requests from saving.
```
ericboucher, line 156: I'd rather open a small `enhancement` issue and reference it in here to add more context rather than detail everything.

TODO - send a notification on error. See issue #XXX for more information.

ericboucher (approved):

src/utils/server-utils.ts
SimplyBallistic: Done: #114

SimplyBallistic:
Reverted caching to #115

PR #115, opened 2020-10-04
Groundstation cache promises

SimplyBallistic:
Extension to #113 
Adds caching to getDates url so it's not called multiple times.
Not a great implementation, any better ideas?

### DIFF
--- a/src/components/MapView/index.tsx
+++ b/src/components/MapView/index.tsx
@@ -16,32 +16,23 @@ import Legends from './Legends';
 // layers
 import {
   BoundaryLayer,
-  PointDataLayer,
   ImpactLayer,
   NSOLayer,
+  PointDataLayer,
   WMSLayer,
 } from './Layers';
 
-import {
-  DiscriminateUnion,
-  PointDataLayerProps,
-  ImpactLayerProps,
-  LayerType,
-  WMSLayerProps,
-} from '../../config/types';
+import { DiscriminateUnion, LayerType } from '../../config/types';
 
-import {
-  getBoundaryLayerSingleton,
-  LayerDefinitions,
-} from '../../config/utils';
+import { getBoundaryLayerSingleton } from '../../config/utils';
 
 import DateSelector from './DateSelector';
 import {
+  dateRangeSelector,
   isLoading,
   layersSelector,
 } from '../../context/mapStateSlice/selectors';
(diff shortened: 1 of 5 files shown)


SimplyBallistic:
@bencpeters is this an ok implementation? Not sure if keeping a dictionary like this is a good practice.

bencpeters (approved):
I think the general approach (using a key-value object store as a cache and only making a new request if the key isn't found) is fine; however I think I'd store the actual resolved values in the cache rather than the request promises. Your check would then become something like `const data = pointDataCache[url] || (await loadPointLayerDataFromUrl)`. There's a _slight_ performance advantage to this - you don't do any extra async actions when loading cached data - but it's probably negligible, I think the bigger thing is that it's easier to reason about and there's no benefit that I see to storing the promises themselves.

Another possible approach would be to move the cache from a file/module scoped variable to one living in redux state. This would come with a few advantages - easier debugging using redux history & state debugging tools, immutability checks (from our middleware), the ability to seamless serialize the cache to local/session storage automatically using middleware if we ever decided to go that route. It might add slightly more complication, and I think the approach you laid out in this PR will work fine, so I'm on the fence whether it's worth changing, but it might ultimately be a cleaner solution...

SimplyBallistic:
Thanks for the detailed response!
> however I think I'd store the actual resolved values in the cache rather than the request promises. 

I actually tried this initially but couldn't get it to work. I believe its because the date promises are executed concurrently via `Promise.all()` rather than sequentially. I think doing it like you mentioned would mean we cannot use `Promise.all` anymore and instead would need to call the promises sequentially - correct me if I'm wrong. In plain English: "Before I try do this task, see if its being done by someone else. If so, copy their result once they're done instead of starting a new one".

> Another possible approach would be to move the cache from a file/module scoped variable to one living in redux state. This would come with a few advantages - easier debugging using redux history & state debugging tools, immutability checks (from our middleware), the ability to seamless serialize the cache to local/session storage automatically using middleware if we ever decided to go that route. It might add slightly more complication, and I think the approach you laid out in this PR will work fine, so I'm on the fence whether it's worth changing, but it might ultimately be a cleaner solution...

That would be pretty neat in the future. I think its out of the scope of this PR if what I mentioned above is correct, but maybe in the future if more than just this result is cached, we could invest in a redux slice for this :thinking:

Issue #89, opened 2020-06-24
Link to an analysis or a view


ericboucher:
## Idea

Allow users to share a link to a specific analysis / selection / view

dubwalla:
Yes, definitely need this in our roadmap. We eventually need to support an alert system which generates an email / message to a user that draws them into PRISM with a specific view

SimplyBallistic:
Since redux state is serialized, you might be able to export a snapshot of the state as base64 and pass it around as a link?

ericboucher:
I was thinking something `simpler`. Rather than transferring the state, we should be able to link to an analysis by its parameters in the URL like this:

`prim_url.org/analysis?hazard_layer=pasture_anomaly&statistic=mean&basline_layer=children&date=2020-05-21`

dubwalla:
While we are doing this, can we make this apply to views more broadly, and not just an analysis? I.e. also load a specified view based on URL parameters: prim_url.org/view?hazard_layer=pasture_anomaly&date=2020-05-21

Ry-DS:
## Update 2021
Several PRs have attempted to solve this issue, but didn't follow the kind of pattern we were hoping for. You can view them here:
- #103
- #120 

We hope to build an approach which is both scalable (as new fields get added, little to no extra work is needed to update this code) and also simple/modular (we could eventually extend this code to support new types of layers as they come).

### Requirements:
- The URL query is created dynamically - we shouldn't be hard coding the query like what is done [here.](https://github.com/WFP-VAM/prism-frontend/pull/120/files#diff-314a2832937cda41c6f51176b41683977e412b5dc6b10d8ffff85ffa2f293bd1R197)
    - This can be done by storing the fields inside an object instead of in separate state fields. A single `useState` or `useReducer` - complex but might be worthwhile if you can think of a lot of actions to update from url, edit fields etc - should be used to store the whole field state.
- The analysis dropdown opens if there is a valid query, with all fields filled in.
- There is an error checker. Nothing advanced, but if there's something obviously wrong with the url we can create a notification with the notification state slice.

### Tips:
- I recommend `useReducer` since I envision many actions we can easily encapsulate such as `urlToState`, `stateToUrl`, `updateField` etc. However, until we put into practise we won't know if its really any better than a simple `useState`. It'll be up to you to experiment!
[...]

PR #120, opened 2020-12-25
Link analysis

khattakdev:
- [x] Works for Run Analysis button 
run Analysis

#89

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,4 +32,5 @@ yarn-error.log*
 .vscode/
 .idea/
 
+# VSCode extensions
 TODO
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,5 @@ yarn-error.log*
 # Editors
 .vscode/
 .idea/
+
+TODO
--- a/src/components/MapView/Analyser/index.tsx
+++ b/src/components/MapView/Analyser/index.tsx
@@ -34,6 +34,7 @@ import bbox from '@turf/bbox';
 
 import DatePicker from 'react-datepicker';
 
+import { extractPropsFromURL } from './util';
 import {
   getBoundaryLayerSingleton,
   LayerDefinitions,
@@ -44,6 +45,7 @@ import {
   NSOLayerProps,
   WMSLayerProps,
   isLayerKey,
+  isStatistic,
   LayerKey,
 } from '../../../config/types';
 import { LayerData } from '../../../context/layers/layer-data';
@@ -200,21 +202,17 @@ function Analyser({ classes }: AnalyserProps) {
   };
 
   useEffect(() => {
-    const queryString = history.location.search;
-    const params = new URLSearchParams(queryString);
-    const hazardLayerParamId = params.get('hazardLayerId');
-    const baselineLayerParamId = params.get('baselineLayerId');
-    const selectedParamDate: number = Number(params.get('date'));
-    const statisticParam = params.get('statistic');
-    const aboveThresholdParam: string = params.get('aboveThreshold') || '';
-    const belowThresholdParam: string = params.get('belowThreshold') || '';
+    const {
+      hazardLayerParamId,
+      baselineLayerParamId,
+      selectedParamDate,
+      statisticParam,
+      aboveThresholdParam,
+      belowThresholdParam,
+    } = extractPropsFromURL(history.location.search);
 
     if (!hazardLayerParamId || !baselineLayerParamId || !statisticParam) {
       return;
-    }
-
-    function isStatistic(layer: string): layer is AggregationOperations {
-      return layer in AggregationOperations;
     }
 
     if (isLayerKey(hazardLayerParamId)) {
--- a/src/components/MapView/Analyser/index.tsx
+++ b/src/components/MapView/Analyser/index.tsx
@@ -199,6 +199,8 @@ function Analyser({ classes }: AnalyserProps) {
   };
 
   useEffect(() => {
+    // eslint-disable-next-line no-debugger
+    debugger;
     const queryString = history.location.search;
     const params = new URLSearchParams(queryString);
     const hazardLayerParamId = params.get('hazardLayerId');
@@ -214,14 +216,14 @@ function Analyser({ classes }: AnalyserProps) {
     function isHazardLayerId(layer: string): layer is LayerKey {
       return typeof layer === 'string';
     }
-    function isBaselineLAyerId(layer: string): layer is LayerKey {
+    function isBaselineLayerId(layer: string): layer is LayerKey {
       return typeof layer === 'string';
     }
 
     if (isHazardLayerId(hazardLayerParamId)) {
       setHazardLayerId(hazardLayerParamId);
     }
-    if (isBaselineLAyerId(baselineLayerParamId)) {
+    if (isBaselineLayerId(baselineLayerParamId)) {
       setBaselineLayerId(baselineLayerParamId);
     }
 
@@ -400,7 +402,7 @@ function Analyser({ classes }: AnalyserProps) {
             {!analysisResult && (
               <Button
                 className={classes.innerAnalysisButton}
-                onClick={() => runAnalyser()}
+                onClick={runAnalyser}
                 disabled={
                   !!thresholdError || // if there is a threshold error
                   !selectedDate || // or date hasn't been selected
--- a/src/components/MapView/Analyser/index.tsx
+++ b/src/components/MapView/Analyser/index.tsx
@@ -95,17 +95,22 @@ function Analyser({ classes }: AnalyserProps) {
 
   // set default date after dates finish loading and when hazard layer changes
   useEffect(() => {
+    // eslint-disable-next-line no-debugger
+    debugger;
     const dates = hazardLayerId
       ? availableDates[
           (LayerDefinitions[hazardLayerId] as WMSLayerProps)?.serverLayerName
         ]
       : null;
+    if (Number(new URLSearchParams(history.location.search).get('date'))) {
+      return;
+    }
     if (!dates || dates.length === 0) {
       setSelectedDate(null);
     } else {
       setSelectedDate(dates[dates.length - 1]);
     }
-  }, [availableDates, hazardLayerId]);
+  }, [availableDates, hazardLayerId, history.location.search]);
 
   const onOptionChange = <T extends string>(
     setterFunc: Dispatch<SetStateAction<T>>,
@@ -163,10 +168,9 @@ function Analyser({ classes }: AnalyserProps) {
       return;
     } // hasn't been calculated yet
 
-    // Temporary Commented
-    // if (!selectedDate) {
-    //   throw new Error('Date must be given to run analysis');
-    // }
+    if (!selectedDate) {
+      throw new Error('Date must be given to run analysis');
+    }
 
     if (!hazardLayerId || !baselineLayerId) {
       throw new Error('Layer should be selected to run analysis');
@@ -182,7 +186,7 @@ function Analyser({ classes }: AnalyserProps) {
     const params: AnalysisDispatchParams = {
       hazardLayer: selectedHazardLayer,
       baselineLayer: selectedBaselineLayer,
-      date: selectedDate || 1606824000000,
+      date: selectedDate,
       statistic,
       extent: adminBoundariesExtent,
       threshold: {
@@ -204,6 +208,7 @@ function Analyser({ classes }: AnalyserProps) {
     const hazardLayerParamId: string = params.get('hazardLayerId') || '';
     const baselineLayerParamId: string = params.get('baselineLayerId') || '';
     const selectedParamDate: number = Number(params.get('date'));
+    const statisticParam: string = params.get('statistic') || '';
 
     function isHazardLayerId(layer: string): layer is LayerKey {
       return typeof layer === 'string';
@@ -219,10 +224,20 @@ function Analyser({ classes }: AnalyserProps) {
       setBaselineLayerId(baselineLayerParamId);
     }
 
+    // Mean is selected by default
+    if (statisticParam === 'median') {
+      setStatistic(AggregationOperations.Median);
+    }
+
     setSelectedDate(selectedParamDate);
 
     // Avoid Running Analyser if boundary Layer data is Null
-    if (boundaryLayerData && hazardLayerId && baselineLayerId) {
+    if (
+      boundaryLayerData &&
+      hazardLayerId &&
+      baselineLayerId &&
+      statisticParam
+    ) {
       runAnalyser();
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
--- a/src/components/MapView/Analyser/index.tsx
+++ b/src/components/MapView/Analyser/index.tsx
@@ -235,11 +235,11 @@ function Analyser({ classes }: AnalyserProps) {
     setBelowThreshold(belowThresholdParam);
 
     // Avoid Running Analyser if boundary Layer data is Null
-    if (boundaryLayerData && hazardLayerId && baselineLayerId && selectedDate) {
+    if (hazardLayerId && baselineLayerId && selectedDate) {
       runAnalyser();
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [boundaryLayerData]);
+  }, [availableDates, boundaryLayerData, history.location.search]);
 
   return (
     <div className={classes.analyser}>
--- a/src/components/MapView/Analyser/index.tsx
+++ b/src/components/MapView/Analyser/index.tsx
@@ -24,6 +24,7 @@ import {
   WithStyles,
 } from '@material-ui/core';
 import { grey } from '@material-ui/core/colors';
+import { useHistory } from 'react-router-dom';
 
 import { faCaretDown, faChartBar } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
@@ -68,6 +69,7 @@ const boundaryLayer = getBoundaryLayerSingleton();
 
 function Analyser({ classes }: AnalyserProps) {
   const dispatch = useDispatch();
+  const history = useHistory();
   const boundaryLayerData = useSelector(layerDataSelector(boundaryLayer.id)) as
     | LayerData<BoundaryLayerProps>
     | undefined;
@@ -134,6 +136,8 @@ function Analyser({ classes }: AnalyserProps) {
   };
 
   const adminBoundariesExtent = useMemo(() => {
+    // eslint-disable-next-line no-debugger
+    debugger;
     if (!boundaryLayerData) {
       // not loaded yet. Should be loaded in MapView
       return null;
@@ -155,13 +159,15 @@ function Analyser({ classes }: AnalyserProps) {
   const clearAnalysis = () => dispatch(clearAnalysisResult());
 
   const runAnalyser = async () => {
+    // eslint-disable-next-line no-debugger
+    debugger;
     if (!adminBoundariesExtent) {
       return;
     } // hasn't been calculated yet
 
-    if (!selectedDate) {
-      throw new Error('Date must be given to run analysis');
-    }
+    // if (!selectedDate) {
+    //   throw new Error('Date must be given to run analysis');
+    // }
 
     if (!hazardLayerId || !baselineLayerId) {
       throw new Error('Layer should be selected to run analysis');
@@ -177,7 +183,7 @@ function Analyser({ classes }: AnalyserProps) {
     const params: AnalysisDispatchParams = {
       hazardLayer: selectedHazardLayer,
       baselineLayer: selectedBaselineLayer,
-      date: selectedDate,
+      date: selectedDate || 1606824000000,
       statistic,
       extent: adminBoundariesExtent,
       threshold: {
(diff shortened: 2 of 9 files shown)


src/components/MapView/Analyser/index.tsx
```diff
@@ -348,7 +400,7 @@ function Analyser({ classes }: AnalyserProps) {
             {!analysisResult && (
               <Button
                 className={classes.innerAnalysisButton}
-                onClick={runAnalyser}
+                onClick={() => runAnalyser()}
```
ericboucher, line 403: is that change necessary?

.gitignore
```diff
@@ -31,3 +31,5 @@ yarn-error.log*
 # Editors
 .vscode/
 .idea/
+
+TODO
```
ericboucher, line 35: is this needed?

ericboucher (commented):
This is a great start. I would recommend moving the URL handling into a separate file to keep things clean and think about how you can DRY your code. a bit more.

src/components/MapView/Analyser/index.tsx
```diff
@@ -187,8 +192,55 @@ function Analyser({ classes }: AnalyserProps) {
     };
 
     await dispatch(requestAndStoreAnalysis(params));
+    // eslint-disable-next-line fp/no-mutating-methods
+    history.push(
```
ericboucher, line 196: Let's create a smarter, DRYer util function for that and make sure we handle edge cases. What happens when some values are undefined? eg. for thresholds?

ericboucher:
Some tests are failing, I think you just need to mock a few more functions around `history`

src/components/MapView/Analyser/index.tsx
```diff
@@ -187,8 +192,55 @@ function Analyser({ classes }: AnalyserProps) {
...
+    const aboveThresholdParam: string = params.get('aboveThreshold') || '';
+    const belowThresholdParam: string = params.get('belowThreshold') || '';
+
+    if (!hazardLayerParamId || !baselineLayerParamId || !statisticParam) {
+      return;
+    }
+    function isHazardLayerId(layer: string): layer is LayerKey {
+      return typeof layer === 'string';
+    }
+    function isBaselineLAyerId(layer: string): layer is LayerKey {
```
Ry-DS, line 217:
```suggestion
    function isBaselineLayerId(layer: string): layer is LayerKey {
```

Ry-DS: If you are stuck on how to complete these functions, I recommend checking the key exists here: https://github.com/WFP-VAM/prism-frontend/blob/master/src/config/utils.ts#L116, then checking the `layer.type` to verify its the correct type (Hazard - WMS or Baseline - NSO)

src/components/MapView/Analyser/index.tsx
```diff
@@ -98,12 +100,15 @@ function Analyser({ classes }: AnalyserProps) {
           (LayerDefinitions[hazardLayerId] as WMSLayerProps)?.serverLayerName
         ]
       : null;
+    if (Number(new URLSearchParams(history.location.search).get('date'))) {
```
Ry-DS, line 103: I don't think this is a good idea. The goal of this useEffect is to load default dates when a new hazard layer is selected. We will now disable this feature if there's something in the url.

src/components/MapView/Analyser/index.tsx
```diff
@@ -187,8 +192,55 @@ function Analyser({ classes }: AnalyserProps) {
...
+
+    if (isHazardLayerId(hazardLayerParamId)) {
+      setHazardLayerId(hazardLayerParamId);
+    }
+    if (isBaselineLAyerId(baselineLayerParamId)) {
+      setBaselineLayerId(baselineLayerParamId);
+    }
+
+    // Mean is selected by default
+    if (statisticParam === 'median') {
```
Ry-DS, line 229: Only mean is supported in the url?

src/components/MapView/Analyser/index.tsx
```diff
@@ -187,8 +192,55 @@ function Analyser({ classes }: AnalyserProps) {
...
+    setSelectedDate(selectedParamDate);
+    setAboveThreshold(aboveThresholdParam);
+    setBelowThreshold(belowThresholdParam);
+
+    // Avoid Running Analyser if boundary Layer data is Null
+    if (boundaryLayerData && hazardLayerId && baselineLayerId && selectedDate) {
+      runAnalyser();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [boundaryLayerData]);
```
Ry-DS, line 242: We shouldn't aim for this code to just be run on startup. If the url changes because it is manually edited, or the user goes back a page, the page should change accordingly.

Ry-DS (changes requested):

Ry-DS:
Good start! However, I think we might have to change our approach on how we load and update the url.

The goal of this PR is to help users easily share *completed* analysis results (all fields are filled in and run analysis was dispatched). There is a lot of extra work we are doing here to try handle all these smaller edge cases where certain fields are not filled in, and the code is also not very scalable if we decide to add new fields in the future.

I think its safe if we just ignore filling in the UI's fields, and just dispatch an analysis if all criteria are met. If a dispatch isn't possible, we could simply notify the user that their url is malformed. A url would only ever be malformed if the user edited it, or didn't copy it properly. 

The above approach would definitely simplify things, and you won't need to touch some of the more delicate pieces like https://github.com/WFP-VAM/prism-frontend/pull/120#discussion_r551077956. What do you think of this approach @ericboucher? This is the current approach #103 was taking.

Also, I don't think we should be doing any verification when we load the url like what we are doing now, but instead rely on the dispatch function to verify for us. Then, in the future we could use this function in other places and use the same verification code. This was the approach Laveesh and I were taking [here.](https://github.com/WFP-VAM/prism-frontend/pull/103/files#diff-edfb4dd8e4a2f9b60a39a96700305dadc49e358206c8f5b88f82343e67529c73R45)

Try give the above a shot, and possibly use #103 as a jumping point. Let me know if you have any questions :+1:

src/components/MapView/Analyser/index.tsx
khattakdev: It wasn't working initially, so I had to use an anonymous function which fixed the issue but reverted it back just now and it worked fine.

.gitignore
khattakdev: There is a VSCODE extension that you can use for adding **TODOs** so this was added intentionally to ignore `Todo`

src/components/MapView/Analyser/index.tsx
khattakdev: > If you are stuck on how to complete these functions, I recommend checking the key exists here: https://github.com/WFP-VAM/prism-frontend/blob/master/src/config/utils.ts#L116, then checking the `layer.type` to verify its the correct type (Hazard - WMS or Baseline - NSO)

The function is already complete, It's doing what I wanted that is to check whether the passed argument is a correct Layer ( `"ndvi" | "ModisLST" | "ModisVHI" ... `)

src/components/MapView/Analyser/index.tsx
khattakdev: The goal of this if statement is to check if the `URL`  has a date in it, ignore the code within the `useEffect` which actually mess up with the URL dates and you don't get the expected result.

src/components/MapView/Analyser/index.tsx
khattakdev:
```
const [statistic, setStatistic] = useState(AggregationOperations.Mean);
```
This means `Mean` will always be selected by default, so we don't need to worry about it.
Statistics can have only two values either `Mean`  or `Median`, one is selected by default when declaring the value, we only need to care about the other one.

src/components/MapView/Analyser/index.tsx
khattakdev: This takes care of all the scenarios you mentioned.
If the user manually edited the URL, the page is going to rerender everything and that will execute `runAnalyser`. The same thing happens when users navigate backs.

khattakdev:
> Good start! However, I think we might have to change our approach on how we load and update the url.
> 
> The goal of this PR is to help users easily share _completed_ analysis results (all fields are filled in and run analysis was dispatched). There is a lot of extra work we are doing here to try handle all these smaller edge cases where certain fields are not filled in, and the code is also not very scalable if we decide to add new fields in the future.
> 
> I think its safe if we just ignore filling in the UI's fields, and just dispatch an analysis if all criteria are met. If a dispatch isn't possible, we could simply notify the user that their url is malformed. A url would only ever be malformed if the user edited it, or didn't copy it properly - two cases we don't need to cover as developers.
> 
> The above approach would definitely simplify things, and you won't need to touch some of the more delicate pieces like [#120 (comment)](https://github.com/WFP-VAM/prism-frontend/pull/120#discussion_r551077956). What do you think of this approach @ericboucher? This is the current approach #103 was taking.
> 
> Also, I don't think we should be doing any verification when we load the url like what we are doing now, but instead rely on the dispatch function to verify for us. Then, in the future we could use this function in other places and use the same verification code. This was the approach Laveesh and I were taking [here.](https://github.com/WFP-VAM/prism-frontend/pull/103/files#diff-edfb4dd8e4a2f9b60a39a96700305dadc49e358206c8f5b88f82343e67529c73R45)
> 
> Try give the above a shot, and let me know if you have any questions 👍

Two reasons why I chose this approach.
* If a user opens such a link, he will have all the UI fields filled in, which can improve UX, Image if he wants to make a small change like changing the `Statistics` only, He will need to fill all the fields manually just for one change.

* Doing so, I don't need to make any changes in `runAnalyser` because it takes the values from input fields, all I need to do is to just update those fields and execute `runAnalyser`. 

```
 const params: AnalysisDispatchParams = {
      hazardLayer: selectedHazardLayer,
      baselineLayer: selectedBaselineLayer,
      date: selectedDate,
      statistic,
      extent: adminBoundariesExtent,
      threshold: {
        above: parseFloat(aboveThreshold) || undefined,
        below: parseFloat(belowThreshold) || undefined,
      },
    };
    
    await dispatch(requestAndStoreAnalysis(params));
```

This function's ( `requestAndStoreAnalysis` ) params can take specific values only, e.g. `hazardLayer` will only take valid layer names else TypeScript will return an error, so I'll still need to verify the user inputs to make sure it's not an unknown value.

src/components/MapView/Analyser/index.tsx
khattakdev: > Let's create a smarter, DRYer util function for that and make sure we handle edge cases. What happens when some values are undefined? eg. for thresholds?

I'll refactor the code once I get approval of my changes. 
Taking about the edge cases, I'm taking care of them already

```
    if (isHazardLayerId(hazardLayerParamId)) {
      setHazardLayerId(hazardLayerParamId);
    }
    if (isBaselineLayerId(baselineLayerParamId)) {
      setBaselineLayerId(baselineLayerParamId);
    }

    // Mean is selected by default
    if (statisticParam === 'median') {
      setStatistic(AggregationOperations.Median);
    }
```

Thresholds are optional, so I didn't add any checks for them. Also, they become undefined if they have an invalid number 
```
      threshold: {
        above: parseFloat(aboveThreshold) || undefined,
        below: parseFloat(belowThreshold) || undefined,
      },
```

.gitignore
Ry-DS: A comment then like  `# VSCode extensions` might be nice

src/components/MapView/Analyser/index.tsx
Ry-DS: We still need to find a better way to do this. 

If I run an analysis, and copy paste the url to a new page, then try editing the hazard layer field, the dates won't update like they're supposed to. The approach I mentioned below (https://github.com/WFP-VAM/prism-frontend/pull/120#issuecomment-753710241) could work here. Otherwise, if there's a solution you can think of which still keeps this feature working, feel free to include it :+1:

src/components/MapView/Analyser/index.tsx
Ry-DS: But if we rely on this default, things can easily break down the line. What if:

- We change our default?
- We add more options like `Mode`?
- The user triggers a re-run of this function while `Median` is selected but the url states `Mean`?

These cases will easily break this logic, therefore it would be great to not rely on defaults, especially in a useEffect.
If we also want to add Back support, where the user can navigate backwards, point 3 will come into play here.

src/components/MapView/Analyser/index.tsx
Ry-DS: I tried navigating back just now, but nothing happened. Which browser did you test on? (I used Edge)

We shouldn't ignore this eslint rule anyway, as its a paradigm of `useEffect` to exhaust all dependencies in use. Adding all dependencies will also add back support for all browsers.

src/components/MapView/Analyser/index.tsx
Ry-DS: Are you sure checking for a string type is good enough? We could pass absolutely anything like `not_a_layer` and cause a nasty error somewhere down the line. I tried just now, and it crashed the application without a graceful error/warning.

I recommend checking if the string is actually valid using the link I sent so we can verify the layer actually exists, and follow with a non-action or a graceful error.

src/components/MapView/Analyser/index.tsx
Ry-DS: One I can think of off the top of my head is using a new boolean state to dictate if the state was set by the user or by the computer. If it was set by the computer, we can ignore this useEffect, otherwise we should still run this code. Hopefully this is helpful enough as a jumping point

Ry-DS:
> Two reasons why I chose this approach.
> 
> * If a user opens such a link, he will have all the UI fields filled in, which can improve UX, Image if he wants to make a small change like changing the `Statistics` only, He will need to fill all the fields manually just for one change.
> * Doing so, I don't need to make any changes in `runAnalyser` because it takes the values from input fields, all I need to do is to just update those fields and execute `runAnalyser`.

This sounds fair, and filling all UI fields would be nice. Then, we should focus on the review comments I made, and making this more stable. 

For example, this might be a good place to use `useReducer`? Then we have the object keys available to us in order to construct a url, and if a new field is added it will be very easy to include it into the url without making a new condition for it (it would almost be automatic!) We could start this by removing all the `useState` uses related to the AnalysisUI with a reducer function, and `runAnalysis` can become an action. Loading a url could also be an action.

khattakdev:
I've fixed pretty much all the issues, let me know if I missed any.

src/components/MapView/Analyser/index.tsx
```diff
@@ -98,12 +103,15 @@ function Analyser({ classes }: AnalyserProps) {
           (LayerDefinitions[hazardLayerId] as WMSLayerProps)?.serverLayerName
         ]
       : null;
+    // if (Number(new URLSearchParams(history.location.search).get('date'))) {
```
Ry-DS, line 106: Remove this comment? Also, did you test if the `date` url query works after removing this? We'll need a solution that both allows this `useEffect` to work alongside the url query code. If you need help coming up with a way to do this let me know :+1:

src/components/MapView/Analyser/index.tsx
```diff
@@ -98,12 +103,15 @@ function Analyser({ classes }: AnalyserProps) {
...
+    //   return;
+    // }
     if (!dates || dates.length === 0) {
       setSelectedDate(null);
     } else {
       setSelectedDate(dates[dates.length - 1]);
     }
-  }, [availableDates, hazardLayerId]);
+  }, [availableDates, hazardLayerId, history.location.search]);
```
Ry-DS, line 114: Where did this extra dependency come from?

src/components/MapView/Analyser/index.tsx
```diff
@@ -187,8 +195,47 @@ function Analyser({ classes }: AnalyserProps) {
     };
 
     await dispatch(requestAndStoreAnalysis(params));
+    // eslint-disable-next-line fp/no-mutating-methods
+    history.replace(
+      `?hazardLayerId=${hazardLayerId}&baselineLayerId=${baselineLayerId}&date=${selectedDate}&statistic=${statistic}&aboveThreshold=${aboveThreshold}&belowThreshold=${belowThreshold}`,
```
Ry-DS, line 200: We need to find a better way than hard coding all the fields into the url. After talking with Eric, `useReducer` seems like a perfect fit here, as it will give you access to the object key names, so the url will automatically add new fields as they are introduced without any extra work. Let me know if you need help implementing this.

src/components/MapView/Analyser/index.tsx
```diff
@@ -187,8 +195,47 @@ function Analyser({ classes }: AnalyserProps) {
...
+    if (isStatistic(statisticParam)) {
+      setStatistic(statisticParam);
+    }
+
+    setSelectedDate(selectedParamDate);
+    setAboveThreshold(aboveThresholdParam);
+    setBelowThreshold(belowThresholdParam);
+
+    // Avoid Running Analyser if boundary Layer data is Null
+    if (hazardLayerId && baselineLayerId && selectedDate) {
```
Ry-DS, line 233: I'd prefer if we'd use a try/catch so we can reuse the existing verification inside `runAnalyser()`

src/components/MapView/Analyser/index.tsx
```diff
@@ -187,8 +195,47 @@ function Analyser({ classes }: AnalyserProps) {
...
+      runAnalyser();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Ry-DS, line 236: You should be able to remove this, and use all dependencies ESLint asks you to include. Let me know if you face issues with this :+1: (we want this code to be written in a way that's independent of the current state of the application, so it remains stable. This eslint rule will help ensure this.)

src/components/MapView/Analyser/util.ts
```diff
@@ -0,0 +1,18 @@
+export const extractPropsFromURL = (path: string) => {
```
Ry-DS, line 1: Move this into an action along with most the code in your `useEffect` once you start using `useReducer`

src/components/NavBar/index.tsx
```diff
@@ -16,6 +16,10 @@ import {
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faInfoCircle, faBars } from '@fortawesome/free-solid-svg-icons';
 import { faGithub } from '@fortawesome/free-brands-svg-icons';
+import { useDispatch } from 'react-redux';
+// Basic CSS Layout for the whole page
```
Ry-DS, line 20: What is this comment referring to?

src/components/NavBar/index.tsx
```diff
@@ -64,6 +70,7 @@ function NavBar({ classes }: NavBarProps) {
               className={classes.logo}
               component={Link}
               to="/"
+              onClick={clearAnalysis}
```
Ry-DS, line 73: We really shouldn't need to do this. What if we navigate to `/` somewhere else in the codebase? We would need to include this function everywhere. 

A better way to do this would be to rely on the `useEffect` to clear the pending analysis if it detects the url has changed to `/`. This should happen automatically [here](https://github.com/WFP-VAM/prism-frontend/pull/120/files#diff-314a2832937cda41c6f51176b41683977e412b5dc6b10d8ffff85ffa2f293bd1R204)

src/config/types.ts
```diff
@@ -212,6 +212,9 @@ export enum AggregationOperations {
   Median = 'median',
 }
 
+export const isStatistic = (stat: string): stat is AggregationOperations =>
```
Ry-DS, line 215: We could have a function for this, but I personally feel `stat in AggregationOptions` is readable enough, and I'm pretty sure TS is smart enough to apply its own type predicate if it sees this. This is up to discussion however, and you are free to keep this if you think its better :+1:

src/components/NavBar/index.test.tsx
```diff
@@ -1,10 +1,16 @@
...
+import { Provider } from 'react-redux';
+import { store } from '../../context/store';
 import NavBar from '.';
 
 jest.mock('./MenuItem', () => 'mock-MenuItem');
 
 test('renders as expected', () => {
-  const { container } = render(<NavBar />);
+  const { container } = render(
+    <Provider store={store}>
```
Ry-DS, line 11: This should be undone as per the comment I made on NavBar/index.tsx

Ry-DS (changes requested):
I recommended some changes to help make this code more scalable, as right now it is quite restricted to our current use case.

src/components/MapView/Analyser/index.tsx
khattakdev: That will always return throw an error. The point of `if` statement is to avoid executing function at all. Before the values are updated, it tries to execute `runAnalyser` (and the layers are not defined yet).

src/components/MapView/Analyser/index.tsx
khattakdev: An extra dependency I added for dates but forgot to remove it. Nvm.

src/components/MapView/Analyser/index.tsx
khattakdev: Removed!

src/components/MapView/Analyser/index.tsx
khattakdev: In that case, I'll have to add `hazardLayerId` `baselineLayerId` `selectedDate` etc as well which will cause some issues. (If you do that, you won't be able to change the Analysis from UI, it will go into some infinity loop)

src/components/MapView/Analyser/index.tsx
khattakdev: @Ry-DS Do you mean to have a reducer just for the URL which updates on every input field update?

PR #129, opened 2021-01-29
Use accordion menu on mobile

hav142:
Change menu for mobile screen using accordion

before:
image

after:
image


dubwalla:
This functionality looks good. I've asked @ericboucher to have a look at how this was implemented.

ericboucher:
@hav142 could you please add screenshots before/after to assess the visual changes? :)

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+        margin: 'auto',
+        marginLeft: 0,
+        width: '100%',
+      },
+      boxShadow: 'none',
+    },
+    expanded: {},
+  })(MuiAccordion);
+
+  const AccordionSummary = withStyles(theme => ({
```
ericboucher, line 66: I think you could create classes to pass to the different elements instead and generate styles only once. For example, look at how it is implemented in https://github.com/WFP-VAM/prism-frontend/blob/master/src/components/MapView/Legends/index.tsx

src/components/NavBar/index.tsx
```diff
@@ -144,9 +147,10 @@ const styles = (theme: Theme) =>
 
     drawerContent: {
       backgroundColor: theme.palette.primary.main,
-      padding: 16,
+      paddingTop: 16,
```
ericboucher, line 150: Are you sure that this is not affecting the UI on desktop?

src/components/NavBar/MenuItemMobile/index.tsx
hav142: @ericboucher thanks for the suggestion. I have updated the codes in the new commit

src/components/NavBar/index.tsx
hav142: Should be safe, because the class is only used in mobile screen (drawer that appears when users click the button to show the menu) and it's hidden in the desktop

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+        content: '""',
+        display: 'inline-block',
+        width: 0,
+        height: 0,
+        marginLeft: 3.5,
+        verticalAlign: 3.5,
+        borderTop: '3.5px solid',
+        borderRight: '3.5px solid transparent',
+        borderBottom: 0,
+        borderLeft: '3.5px solid transparent',
```
ericboucher, line 155: I am not sure this is the right approach. @Ry-DS do you have any recs?

hav142: @ericboucher Sorry I forgot to remove that class, it was previously used by a button in the MenuItem before I replaced it using accordion. I updated it in the new commit 🙏

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+  Switch,
+  Theme,
+  Typography,
+  withStyles,
+  WithStyles,
+} from '@material-ui/core';
+import Accordion from '@material-ui/core/Accordion';
+import AccordionSummary from '@material-ui/core/AccordionSummary';
+import AccordionDetails from '@material-ui/core/AccordionDetails';
+
```
Ry-DS, line 16: Remove newlines for your import statements :+1:

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faCaretDown } from '@fortawesome/free-solid-svg-icons';
+
```
Ry-DS, line 19:
```suggestion
```

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+import { LayerType, TableType } from '../../../config/types';
+import { addLayer, removeLayer } from '../../../context/mapStateSlice';
+import { loadTable } from '../../../context/tableStateSlice';
+import { layersSelector } from '../../../context/mapStateSlice/selectors';
+import { menuList } from '../utils';
+
+function MenuItemMobile({ classes }: MenuItemMobileProps) {
+  const [expanded, setExpanded] = useState<string>('');
```
Ry-DS, line 27: the `<string>` type assertion isn't necessary, since it can be inferred with `''`

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+  const selectedLayers = useSelector(layersSelector);
+
+  const dispatch = useDispatch();
+
+  const handleChange = (panel: string) => (
+    event: object,
```
Ry-DS, line 34: Try avoiding types such as `object` or `any` since they aren't very strict. Here, we can type this as `React.ChangeEvent<{}>` or the whole constant as `AccordianProps['onChange']` (I prefer the latter)

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+    newExpanded: boolean,
+  ) => {
+    setExpanded(newExpanded ? panel : '');
```
Ry-DS, line 37: why is expanded a string? Makes sense its a boolean...

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+                    <Typography
+                      variant="body2"
+                      className={classes.categoryTitle}
+                    >
+                      {categoryTitle}
+                    </Typography>
+                    <hr />
+
+                    {layers.map(layer => {
+                      const { id: layerId, title: layerTitle } = layer;
```
Ry-DS, line 90: can we move this to the function args? If there's a specific reason its a new line, all good :+1:

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+                      const selected = Boolean(
```
Ry-DS, line 91: You should use array.some() here, then there's no need to cast to boolean

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+                        selectedLayers.find(
+                          ({ id: testId }) => testId === layerId,
+                        ),
+                      );
+                      return (
+                        <div key={layerId} className={classes.layersContainer}>
```
Ry-DS, line 97: We should replace this with a `Box` from MUI, since the class isn't too complex.

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+                          <Switch
+                            size="small"
+                            color="default"
+                            checked={selected}
+                            onChange={() => toggleLayerValue(selected, layer)}
```
Ry-DS, line 102: It probably isn't too important, but it would be better to fetch the new switch state from the event object from `onChange` (by convention), kind of like how we do so for text fields. This isn't that big of a deal however.

This could be done by making `toggleLayerValue` return a function:
```js
const toggleLayerValue=(selected, layer)=>(event)=>{}
```
where the checked state is stored in: `event.target.checked`

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+    },
+
+    layersContainer: {
+      display: 'flex',
+      marginBottom: 8,
+    },
+
+    accordion: {
+      boxShadow: 'none',
+      '&.Mui-expanded': {
```
Ry-DS, line 156: Could we use this to the global MUI theme? If we aren't using Accordion anywhere else, this might be cleaner/less prone to breaking.

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+        margin: 0,
+        '&:before': {
```
Ry-DS, line 158: Not sure what this is selecting exactly. A comment to explain why this is here?

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+        minHeight: 56,
+      },
+      '&:focus': {
+        background: theme.palette.primary.main,
+      },
+    },
+
+    accordionDetail: {
+      backgroundColor: theme.palette.primary.dark,
+      opacity: '0.9',
```
Ry-DS, line 177: this could be a number

src/components/NavBar/MenuItemMobile/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+        '&:before': {
+          opacity: 1,
+        },
+      },
+    },
+
+    accordionSummary: {
+      backgroundColor: theme.palette.primary.main,
+      minHeight: 56,
+      '&.Mui-expanded': {
```
Ry-DS, line 167: Moving this to the MUI theme would be nice too :+1: let me know if it isn't feasible.

src/components/NavBar/index.tsx
```diff
@@ -107,7 +108,9 @@ function NavBar({ classes }: NavBarProps) {
                     <Grid container justify="space-around" item>
                       {buttons}
                     </Grid>
-                    <Grid item>{menu}</Grid>
+                    <Grid container direction="column" item>
+                      <MenuItemMobile />
```
Ry-DS, line 112: `<MenuItem/>` took props for the menu items, but `<MenuItemMobile/>` doesn't? For consistency's sake we should follow a similar approach to `<MenuItem/>` and define a `menu` constant for mobile. Or, we can rename both `MenuItem` and `MenuItemMobile` to `MenuItem*s*...` so its more clear they are multiple items. :wink:

Another approach: We could use `useBreakpoint` and based off whether we are on mobile or not, we can replace the use of `MenuItem` with `MenuItemMobile` on line 40. This means we don't have to touch this part of the file at all.

Ry-DS: Let me know if you have any questions :+1:

src/components/NavBar/index.tsx
Ry-DS: Maybe we could prepend the classname with `drawer` -> `mobileDrawerContent` so its more clear its mobile? Just a suggestion, since `drawer` is normally associated with mobile.

Ry-DS (approved):
Looking good! Here are a few suggestions to help improve code quality.

src/components/NavBar/MenuItemMobile/index.tsx
Ry-DS: On second look, this makes sense. For the comment I mentioned [here](https://github.com/WFP-VAM/prism-frontend/pull/129/files#r593741103), it might be best to just name this file `MobileMenuItems` plural. What do you think?

Ry-DS (changes requested):
Accidently made the above review an approve haha. If we can fix the important ones, we can pass over the nit-picky ones.

dubwalla:
not sure what the status is of this one. @hav142 - have ryan's inputs been addressed?

hav142:
> 
> 
> not sure what the status is of this one. @hav142 - have ryan's inputs been addressed?

@dubwalla I am still working on this. will update when each of inputs have been resolved 🙏

src/components/NavBar/MenuItemMobile/index.tsx
hav142: @Ry-DS I don't really get the `AccordianProps['onChange']` part, do you have an example that I can follow? for now I change it to `React.ChangeEvent<{}>`

src/components/NavBar/MenuItemMobile/index.tsx
hav142: I tried to refactor the codes so that the approach is similar to `<MenuItem/>` component. I think the plural one should be the `const menu` since it consist multiple item. The `<MenuItem/>` or `<MenuItemMobile/>` itself is still singular. What do you think about that?

src/components/NavBar/MenuItemMobile/index.tsx
hav142: This one actually I followed the codes from `<MenuItem/>`, but if I see again the `layer` argument is used in the bottom. How about we remove one line `const { id: layerId, title: layerTitle } = layer;` and use `layer.id` and `layer.title` directly instead?

src/components/NavBar/MenuItemMobile/index.tsx
hav142: just learn that I can use `elevation={0}` so I removed the `boxShadow: none` for this one

src/components/NavBar/MenuItemMobile/index.tsx
hav142: So I want to custom the accordion a little bit because I don't really like how the default element added padding in the top and bottom when the accordion is expanded. the border also disappear when it expanded so I added 2 custom styles here:
`margin: 0` to remove the padding, and `opacity: 1` to keep the border

image

It just my preference but it also fine to remove the custom if we want to use the default one. What do you think?

src/components/NavBar/MenuItemMobile/index.tsx
hav142: I moved all the class related to accordion to the MUI theme 🙏

src/components/NavBar/index.tsx
hav142: I have updated the codes like I mentioned in [here](https://github.com/WFP-VAM/prism-frontend/pull/129#discussion_r595683092). Will need your help to check it again after I push the commit 🙏

hav142:
@Ry-DS Thank you for the inputs, very helpful as a guideline for improving the codes. I think i am done with the updates in my local machine now. I will push the commit after some of the questions that I replied above are clear 🙏

src/components/NavBar/MenuItemMobile/index.tsx
Ry-DS: It would look something like:
```ts
const handleChange: (panel: string)=>AccordianProps['onChange'] = (panel)=>(event)=>{
// etc 
}
```

src/components/NavBar/MenuItemMobile/index.tsx
Ry-DS: It makes sense for `<MenuItem/>` to be singular since it's rendered multiple times with this array: 
https://github.com/WFP-VAM/prism-frontend/blob/22cc5e452ddf855965e800578c7dd0f55b36bff8/src/components/NavBar/index.tsx#L40

The new mobile component you made shouldn't be singular because you use the `menuList` inside of it, meaning we only render one `<MenuItemMobile/>`
https://github.com/WFP-VAM/prism-frontend/blob/011d9f95c397feca04992d4f984610ca6099ff6b/src/components/NavBar/MenuItemMobile/index.tsx#L54

Overall, the `MenuItemMobile` component should be refactored to be similar to `MenuItem`, or it should be renamed to something like `MenuItemsMobile` so it's name makes more sense :+1:

src/components/NavBar/MenuItemMobile/index.tsx
Ry-DS: For the comment I mentioned, we can leave as it is :+1:

On another note, I didn't realise how much of the `MenuItem` was duplicated in `MenuItemMobile`. We should really cut this down with helper components or something else so the code is more maintainable. For example, the switch and table button code are almost identical.

src/components/NavBar/MenuItemMobile/index.tsx
Ry-DS: we should always prioritize using MUI props over classes. If `elevation` works fine definitely use that over a CSS style.

Ry-DS: Reminder: we shouldn't have to do `&.Mui-expanded`. Try doing it in the MUI global theme if you really need this style.
To do this, go here:
https://github.com/WFP-VAM/prism-frontend/blob/ca955c5cc5d3088d591b0d05719be0f604116231/src/muiTheme.tsx#L59
and add this: 
image

src/components/NavBar/MenuItemMobile/index.tsx
Ry-DS: All good, we just need a comment explaining this in code. Also, you should move this to the global theme as indicated here:
https://github.com/WFP-VAM/prism-frontend/pull/129#discussion_r596799486

src/components/NavBar/MenuItemMobile/index.tsx
hav142: I separated the switch/table so that it can be reused by `MenuItem` and `MenuItemMobile`. kindly check in my new commit 🙏

src/components/NavBar/MenuSwitch/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+} from '../../../config/types';
+import { addLayer, removeLayer } from '../../../context/mapStateSlice';
+import { loadTable } from '../../../context/tableStateSlice';
+import { layersSelector } from '../../../context/mapStateSlice/selectors';
+
+function MenuSwitch({ classes, title, layers, tables }: MenuSwitchProps) {
+  const selectedLayers = useSelector(layersSelector);
+  const dispatch = useDispatch();
+
+  const toggleLayerValue = (layer: LayerType) => (event: any) => {
```
Ry-DS, line 26: Possible to type this event so it isn't any?

src/muiTheme.tsx
```diff
@@ -56,6 +56,37 @@ const theme: any = createMuiTheme({
         color: 'black',
       },
     },
+    MuiAccordion: {
+      root: {
+        '&$expanded': {
```
Ry-DS, line 61: Why are we selecting `expanded` like this? You should be able to style `expanded` from outside `root` in `MuiAccordion`.
like:
```ts
MuiAccordion: {
    root: {},
    expanded: {}

}
```

Ry-DS (approved):
Looks good! Just 2 small things

src/muiTheme.tsx
hav142: I tried and it didn't work as expected. I'm not sure, maybe because the generated expanded class is `Mui-expanded`, not `MuiAccordion-expanded`. i got the solution [here](https://github.com/mui-org/material-ui/issues/13576)

Ry-DS: @hav142 all good. In the future if something hacky is used from somewhere else, make sure to link it in a comment!

Issue #123, opened 2021-01-13
Export map as a graphic


dubwalla:
Allow a user to export the map which is currently in view in a downloadable format (JPEG, PNG, PDF)

Issue #100, opened 2020-07-28
Make PRISM mobile friendly


ericboucher:
Make the app more mobile-friendly or add modal notifying users that the app is not optimized for mobile yet.

hav142:
Hi @ericboucher ,

I did some improvements in [idn_dev](https://github.com/WFP-VAM/prism-frontend/tree/idn_dev/) branch based on these 3 issues:

### 1. Menu on mobile screen
![image](https://user-images.githubusercontent.com/25983260/102595944-065c5880-414b-11eb-9654-b827e18b33f7.png)
#### Problems:
- The spacing for each menu is too small. Some users might have accidentally clicked on the wrong menu while accessing it from a mobile device.
- There are menus that is not aligned with the others due to different lengths
- When a user wants to move from one menu to another, they need to take several steps: click one of the menu to open the drop-down, then click anywhere to close the previous drop-down. After that, click the menu that they want to access to open the new drop-down

#### Proposed solution:
Simplify the process using accordion. Each accordion element has the same width and spacing so they are aligned with each other. Using the accordion only requires one click to access each menu or move from one menu to another. In the user experience, even a single click can make a difference

![image](https://user-images.githubusercontent.com/25983260/102596580-eda07280-414b-11eb-8c7f-3719dd503b47.png)

### 2. Button style and position
![image](https://user-images.githubusercontent.com/25983260/102596879-656e9d00-414c-11eb-9972-db58390dc0cc.png)
#### Problems:
- The buttons (analysis and legend) have different style and not aligned each other
- Each button has an absolute position with adjustment on the top and left/right. It can become a problem when we need to add more buttons and customize it for the mobile screen, for example if we want to hide the button label and show only the icon in the mobile screen to save some spaces, the button’s position cannot be aligned automatically.
#### Proposed solution:
Change the legend button style according to the existing standard and make a grid container to place the buttons in it. The absolute position is moved from the buttons to the container so that they will be aligned automatically when the screen size is changed.

![image](https://user-images.githubusercontent.com/25983260/102597107-bbdbdb80-414c-11eb-9165-3a9d5eba6a50.png)

When we need to add more buttons, we can do some customization. For example by making it simpler by hiding the label so that the button only shows its icon on the mobile screen. We can also make the button align vertically as needed.

![image](https://user-images.githubusercontent.com/25983260/102597552-68b65880-414d-11eb-8c58-c491eccc4413.png)

### 3. Date-picker component
[...]

PR #131, opened 2021-02-02
Export map as image 

hav142:
These update is based on point number 2 on issue #100 
- Add container to place all buttons in one grid
- Update legend button style

before:
image

after:
image

And this one based on issue #123
- Add button to export map as graphic (JPEG, PNG, PDF)

The download button is accessible on the top right, next to the legend button. It will show a map preview before users can download different types of images
image


dubwalla:
Let's sync up on some UX thoughts on how this is implemented. I think we should consider 1) how individual layers can be downloaded as images, 2) how to best address additional features that relate to layer interaction (i.e., how can we create a UX which allows users to download / export / arrange layers. I'll add some additional thoughts here and likely in a new issue this week.

ericboucher:
@hav142 let's make sure you update your tests when making changes so that the CI passes. You can simply run `yarn tests -u`

dubwalla:
@ericboucher @Ry-DS - I am not sure if the initial dropdown for downloading is implemented in a consistent UI style. Can you review how this was implemented?

src/components/MapView/Download/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+  ListItemIcon,
+  ListItemText,
+  MenuItem,
+  Theme,
+  Typography,
+  WithStyles,
+  withStyles,
+} from '@material-ui/core';
+import Menu, { MenuProps } from '@material-ui/core/Menu';
+
```
Ry-DS, line 20: Make sure to clean up newlines when you import, unless its for organization.

src/components/MapView/Download/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+      const context = canvas!.getContext('2d');
+      context!.drawImage(activeLayers, 0, 0);
+    }
+  });
+
+  const modalClose = () => {
+    setOpen(false);
+  };
+
+  const download = (format: String) => {
```
Ry-DS, line 89: why is `String` capital?

src/components/MapView/Download/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+      const activeLayers = selectedMap.getCanvas();
+      const canvas = previewRef.current;
+      canvas!.setAttribute('width', activeLayers.width.toString());
+      canvas!.setAttribute('height', activeLayers.height.toString());
+      const context = canvas!.getContext('2d');
+      context!.drawImage(activeLayers, 0, 0);
+    }
+  });
+
+  const modalClose = () => {
```
Ry-DS, line 85: for functions this simple, I think writing them inline is fine. But up to personal preference

src/components/MapView/Download/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
+
+  useEffect(() => {
+    if (open && selectedMap) {
+      const activeLayers = selectedMap.getCanvas();
+      const canvas = previewRef.current;
+      canvas!.setAttribute('width', activeLayers.width.toString());
```
Ry-DS, line 78: you should check for canvas being null instead of using `canvas!` since TS has a right to complain here.

src/components/MapView/Download/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+    }
+  });
+
+  const modalClose = () => {
+    setOpen(false);
+  };
+
+  const download = (format: String) => {
+    const ext = format === 'pdf' ? 'png' : format;
+    const canvas = previewRef!.current;
```
Ry-DS, line 91: Try avoiding the use of `canvas! ` and `previewRef!.` because we cannot guarantee these are not null, and its better to be safe with optional chaining `previewRef?.current`

src/components/MapView/Download/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+      link.setAttribute('download', `map.${ext}`);
+      link.click();
+    }
+    setOpen(false);
+  };
+
+  return (
+    <Grid item>
+      <Button variant="contained" color="primary" onClick={handleClick}>
+        <FontAwesomeIcon style={{ fontSize: '1.2em' }} icon={faFileExport} />
```
Ry-DS, line 115: Maybe we can use MUI icons for better theme consistency? We can use FontAwesome only if the icon doesn't exist in MUI's library.

Ry-DS (commented):
Looks good overall!  A few small fixes

src/components/MapView/Download/index.tsx
hav142: oops 😅. i updated it

src/components/MapView/Download/index.tsx
hav142: I tried to update using MUI icons for the top buttons: 
image

but got an error when i run yarn test
image

do i need to update setupTests or something?

Ry-DS: hm, maybe you can mock the MUI icon with a simple string like `jest.mock('icon path', ()=>'<ChevronArrow/>')` 
If that fails, then we can just use the FontAwesome icon (they are both similar anyway)

hav142: i've tried using like `jest.mock('@material-ui/icons/Visibility', () => 'mock-visibility');`, all tests are passed after i updated the snapshot, but the error message above still remaining.
so, use FontAwesome instead?

ericboucher: Did you trying a similar implementation as https://github.com/WFP-VAM/prism-frontend/blob/master/src/setupTests.tsx ?

hav142: i installed `@material-ui/icons` and added `jest.mock` for specific test file where i used the icon. i think it similar with the implementation in `setupTest`.

i just remembered, do we need to include the `Accordion` component to the list in `setupTest`?

ericboucher: You could spot which lib components take multiple lines and mock them in `setuptTest`

Another thing you can do is break down your component into a few different ones, and test them separately. So that you only test the component within the `.map` once. @Ry-DS any other ideas?

Ry-DS: You don't need to mock Accordion if it work fine without the mock imo (we want to capture the children and mocking prevents that). But if you gave errors, or the snapshot is immensely large because of Accordion, we can mock it :)

ericboucher (approved):

Issue #128, opened 2021-01-18
Add support for full timestamp on WMS layers


dubwalla:
PRISM will soon integrate data on tropical storms and earthquakes. Both of these hazards require full timestamp support. Currently, PRISM only supports YYYY-MM-DD. 

As an example, a layer from WFP's ADAM (accessed through GeoNode WMS - https://geonode.wfp.org/geoserver/prism/wms?request=GetCapabilities) includes the following time format: 2021-01-09T00:00:00.000Z. Future outputs would include the fully formed hour, minute, seconds.

While addressing this issue, we should also look into potentially replacing Moment.js as noted here: https://momentjs.com/docs/

mwarizmiwafiq:
Hi Amit. Yes while we're on it, I think we also need to agree on the standard of handling datetime "available data" in frontend. One of suggestion from @reed1 is to use EPOCH data format as example in this request to get available date for impact layer in PRISM https://vampire.idn.wfp.org/home/get_date/25?start=%22&end=2020-01-01.

You see, that only return set of time array in epoch format. We also need to agree on how we standardize the JSON format as this also will effect on how we handle the timeseries socio-economic data. Please advise @dubwalla. Thank you.

dubwalla:
@ericboucher - thoughts on this issue and input from wafiq?

dubwalla:
@JorgeMartinezG - I think this can be closed via https://github.com/WFP-VAM/prism-frontend/pull/132. Is that correct?

dubwalla:
Addressed via https://github.com/WFP-VAM/prism-frontend/pull/132

Issue #127, opened 2021-01-13
Allow overlay of layers of the same type including new layer visualization elements


dubwalla:
In some cases, multiple data layers of the same type should be overlayed as they collectively facilitate an analysis. Both in the configuration and in the UI, PRISM needs a mechanism to allow more than one layer of the same type to be toggled on. On the configuration side, this could be done with an 'allow overlay' attribute. 

In the UI, the layer should also include the ability to change opacity and order. Opacity should be controlled via a slider. A good UI solution is needed for layer orders, potentially including the ability to drag a layer up/down. A new overall UI for layer selection is needed. Please add suggestions to this issue.

dubwalla:
@hav142 - thanks for looking into this. let me know if unclear.

dubwalla:
Addressed the current need through https://github.com/WFP-VAM/prism-frontend/pull/132

PR #132, opened 2021-02-02
Adam/geonode integration with PRISM

JorgeMartinezG:
This PR references issues #127 and #128 with the goal of including Geonode/Adam layers within PRISM.

**Changes:**
- Enable the project to work with engines higher than 12.x
- Modify getcapabilities response by removing time values within TIME dimension.
- Include adam layers within layers.json and prism.json files.
- Include serverurl for getcapabilities request.
- Group layers to be queried at the same time. The group is set within the prism.json file, _group_layers_ field.
- Modify reducer with the goal of adding and removing multiple layers using a single toggle button.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,6 +120,6 @@
     "eslint-import-resolver-typescript": "^2.0.0"
   },
   "engines": {
-    "node": ">=12.x"
+    "node": "12.x"
   }
 }
--- a/package.json
+++ b/package.json
@@ -120,6 +120,6 @@
     "eslint-import-resolver-typescript": "^2.0.0"
   },
   "engines": {
-    "node": "12.x"
+    "node": ">=12.x"
   }
 }
--- a/src/components/NavBar/MenuItem/index.tsx
+++ b/src/components/NavBar/MenuItem/index.tsx
@@ -12,7 +12,12 @@ import {
 } from '@material-ui/core';
 
 import { MenuItemType, LayerType, TableType } from '../../../config/types';
-import { addLayer, removeLayer } from '../../../context/mapStateSlice';
+import {
+  removeLayers,
+  addLayer,
+  removeLayer,
+  addLayers,
+} from '../../../context/mapStateSlice';
 import { loadTable } from '../../../context/tableStateSlice';
 import { layersSelector } from '../../../context/mapStateSlice/selectors';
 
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -120,6 +120,6 @@
     "eslint-import-resolver-typescript": "^2.0.0"
   },
   "engines": {
-    "node": "12.x"
+    "node": ">=12.x"
```
ericboucher, line 123: Is this change necessary or is it to help for local development?
I am generally against having a "broad" version of node in here. If we think we are ready to update to node 14, then we should aim to set the engine to `14.x` instead and fully test end-to-end that this is not breaking anything (build, deploy, ...)

src/components/NavBar/MenuItem/index.tsx
```diff
@@ -12,7 +12,12 @@ import {
 } from '@material-ui/core';
 
 import { MenuItemType, LayerType, TableType } from '../../../config/types';
-import { addLayer, removeLayer } from '../../../context/mapStateSlice';
+import {
+  removeLayers,
```
ericboucher, line 16: Let's try to keep an alphabetical order when possible.

For simplicity, maybe we can just keep the `multiple` version and use `removeLayers([layerId])` if we need to apply it to a single entity?

src/utils/server-utils.ts
```diff
@@ -68,9 +68,10 @@ function formatCapabilitiesInfo(
 
     const availableDates = dates
       .filter(date => !isEmpty(date))
+      .map(date => get(date, '_text', date).split('T')[0])
       .map(date =>
         // adding 12 hours to avoid  errors due to daylight saving
-        moment.utc(get(date, '_text', date)).set({ hour: 12 }).valueOf(),
+        moment.utc(date).set({ hour: 12 }).valueOf(),
```
ericboucher, line 74: Not a big deal but you could just replace date with `get(date, '_text', date).split('T')[0])` in this line

src/components/NavBar/MenuItem/index.tsx
```diff
@@ -42,6 +47,20 @@ function MenuItem({ classes, title, icon, layersCategories }: MenuItemProps) {
     dispatch(loadTable(table.id));
   };
 
+  const toggleLayers = (layers: LayerType[]) => {
```
ericboucher, line 50: Can you explain a bit more what's happening here? Is this function specific to groupedLayers?

ericboucher (commented):
Nice work overall! Maybe make it clear that this PR is specific to Audtralia? (eg. zoom level change + new endpoints)

package.json
JorgeMartinezG: I had issues setting up locally. No problem removing this constraint

src/utils/server-utils.ts
JorgeMartinezG: Done

src/config/layers.json
```diff
@@ -1065,5 +1065,41 @@
         "color": "#0868ac"
       }
     ]
+  },
```
ericboucher, line 1068: If this is specific to Australia, let's make sure we ave a new config for Australia

Issue #82, opened 2020-06-17
New UI/Functionality for Legend Dropdown


Ry-DS:
The current Legend button:
![image](https://user-images.githubusercontent.com/25571542/84863247-01452080-b0b8-11ea-90c4-e1e2bb38f2aa.png)
Could use some improvement, notably in its style and functionality. 

Ideas: 
1. The Hide/Show button could be styled to fit more into a Material UI theme, making it complement the switches.

2. It could also be converted to a drop-down style, replacing the 'eye' with a chevron arrow.

3. We could add crosses on every legend section so a user can quickly remove a layer by clicking the cross on the layer's legend.

ericboucher:
Good ideas!

I like 1 and 3 a lot, but I think the `eye` works pretty well though and it's easy to guess what it's doing.

@dubwalla do you have any additional feedback?

dubwalla:
Yes, agreed that 1 and 3 are great suggestions. I was thinking that instead of 'removing' a layer, a toggle would be good so that a user can activate or deactivate a layer without having to expand the top nav again. 

It's also worth thinking about having the ability to hide / collapse a legend entry like suggested in 2, but at the individual layer level. An opacity slider at the layer level is also handy to have.

Not to complicate this issue, but also worth mentioning that we also need a UI solution for showing a table view of a layer and to download the layer directly. I was thinking this could be within what is currently the legend area. We can make this a separate issue though

PR #138, opened 2021-02-08
Add layer opacity setting in legend

hav142:
Layer opacity adjusment, currently works for WMS type layer

This feature can be accessed in the legend. The opacity of color indicators will also be adjusted together with the map layer.

image
image

Previous version didn't have slider for opacity yet, and the value is fixed from the layer config
image


ericboucher:
@hav142 could you please add screenshots before/after to assess the visual changes? :)

hav142:
Dear @ericboucher , I have updated the description. My bad, tagging a front-end label but didn't provide any visual 😅

dubwalla:
Hi @hav142. I was wondering why this was only applied to WMS layers. Can we make this feature uniformly available to all layers?

hav142:
> 
> 
> Hi @hav142. I was wondering why this was only applied to WMS layers. Can we make this feature uniformly available to all layers?

@dubwalla sorry for the delayed response. I modified a bit for the layer other than WMS so that the ID can be accessed and set its opacity. Now this should work for all type of layers (raster and polygon. Do we have any other type?). I will push new commit once I tested it on the main branch

I also added a switch to minimize the active legend and hide the map layer temporarily (regarding to issue #82) 

image

ericboucher:
Deployed on staging for testing - https://staging-prism-frontend.web.app/

src/components/MapView/Legends/ColorIndicator/index.tsx
```diff
@@ -11,13 +12,13 @@ const useStyles = makeStyles(() =>
       height: 10,
       width: 10,
       marginRight: 4,
-      backgroundColor: ({ color }: any) => color,
+      backgroundColor: ({ color, opacity }: any) => fade(color, opacity),
```
Ry-DS, line 15: While this was `any`before, we can definitely do better and give this a type :+1:

src/components/MapView/Legends/index.tsx
```diff
@@ -1,11 +1,15 @@
-import React, { PropsWithChildren, useState } from 'react';
+/* eslint-disable fp/no-mutation */
```
Ry-DS, line 1: Is it possible to move this down to the actual uses so it's easier to track?

src/components/MapView/Legends/index.tsx
```diff
@@ -138,8 +232,11 @@ export interface LegendsProps extends WithStyles<typeof styles> {
 interface LegendItemProps
   extends WithStyles<typeof styles>,
     PropsWithChildren<{}> {
```
Ry-DS, line 234: I would personally refactor this to `extends Pick<LayerType, 'id' | 'title' ... > `. This won't work too well however if the optionals are different between the layer type and props type. What do you think?

src/components/MapView/Legends/index.tsx
```diff
@@ -76,35 +89,113 @@ function Legends({ classes, layers }: LegendsProps) {
...
+                      max={1}
+                      aria-labelledby="opacity-slider"
+                      onChange={handleChangeOpacity}
+                    />
+                  </Box>
+                </Grid>
+              )}
+              {legend && (
+                <Grid item>
+                  {legend.map(({ value, color }: any) => (
```
Ry-DS, line 180: Possible to remove this any?

src/components/MapView/Legends/index.tsx
```diff
@@ -76,35 +89,113 @@ function Legends({ classes, layers }: LegendsProps) {
...
+                    <ColorIndicator
+                      key={value}
+                      value={value as string}
+                      color={color as string}
```
Ry-DS, line 184: If the any is removed we might not need these casts

src/components/MapView/Legends/index.tsx
```diff
@@ -76,35 +89,113 @@ function Legends({ classes, layers }: LegendsProps) {
...
+        return [`layer-${id}-circle`, 'circle-opacity'];
+      default:
+        return ['', ''];
+    }
+  })(type);
+
+  const allLayers = map!.getStyle().layers;
+  const isAvailable = allLayers!.find(layer => layer.id === layerId);
+  useEffect(() => {
+    if (!!isAvailable && layerTypes.includes(type!)) {
```
Ry-DS, line 133: Do we need to cast to boolean in an if statement?

src/components/MapView/Legends/index.tsx
```diff
@@ -76,35 +89,113 @@ function Legends({ classes, layers }: LegendsProps) {
...
+  opacity: initialOpacity,
+  children,
+}: LegendItemProps) {
+  const map = useSelector(mapSelector);
+  const [opacity, setOpacityValue] = useState<number | number[]>(
+    initialOpacity || 0,
+  );
+
+  const [toggle, setToggle] = useState(true);
+  const toggleChecked = (prev: Boolean) => {
```
Ry-DS, line 107: Was the captial `Boolean` instead of lowercase `boolean` intentional?

src/components/MapView/Legends/index.tsx
```diff
@@ -76,35 +89,113 @@ function Legends({ classes, layers }: LegendsProps) {
...
+      case 'nso':
+        return [`layer-${id}-fill`, 'fill-opacity'];
+      case 'point_data':
+        return [`layer-${id}-circle`, 'circle-opacity'];
+      default:
+        return ['', ''];
+    }
+  })(type);
+
+  const allLayers = map!.getStyle().layers;
```
Ry-DS, line 130: Instead of assuming these are not null, it would be nice to just render nothing or a message just in case they are

src/components/MapView/Legends/index.tsx
```diff
@@ -76,35 +89,113 @@ function Legends({ classes, layers }: LegendsProps) {
...
+  const isAvailable = allLayers!.find(layer => layer.id === layerId);
```
Ry-DS, line 131: You should use `some` instead so you don't end up casting everywhere. Also, maybe `isLayerActive` is a better name here?

src/components/MapView/Legends/index.tsx
```diff
@@ -76,35 +89,113 @@ function Legends({ classes, layers }: LegendsProps) {
...
+  const [toggle, setToggle] = useState(true);
+  const toggleChecked = (prev: Boolean) => {
+    setToggle(!prev);
+  };
+
+  const handleChangeOpacity = (event: object, newValue: number | number[]) => {
+    setOpacityValue(newValue);
+  };
+
+  const layerTypes = ['wms', 'nso', 'point_data', 'impact'];
```
Ry-DS, line 115: You should type this so that as layer types change, this updates too. I don't remember the codebase too well but I think there might be an enum somewhere with these values. Try looking for it, otherwise just a typing here to remind us to update it if layer types change would be nice.

src/components/MapView/Legends/index.tsx
```diff
@@ -76,35 +89,113 @@ function Legends({ classes, layers }: LegendsProps) {
...
+  const [layerId, opacityType] = (e => {
```
Ry-DS, line 116: Could we type `e` and give it a better name? Similar to the array above I presume.

src/components/MapView/Legends/index.tsx
```diff
@@ -76,35 +89,113 @@ function Legends({ classes, layers }: LegendsProps) {
...
+  const [opacity, setOpacityValue] = useState<number | number[]>(
+    initialOpacity || 0,
+  );
+
+  const [toggle, setToggle] = useState(true);
+  const toggleChecked = (prev: Boolean) => {
+    setToggle(!prev);
+  };
+
+  const handleChangeOpacity = (event: object, newValue: number | number[]) => {
```
Ry-DS, line 111: Type this as SliderProps['onChange'] so we don't need the object type (it's too lenient)

dubwalla:
Progress is looking good. I noticed that the layer toggle on the legend is not sync'd with the toggle in the layer selection menu. I.e. if I toggle the layer off in the legend, the navigation doesn't reflect the new toggle state

src/components/MapView/Legends/index.tsx
```diff
@@ -76,35 +89,113 @@ function Legends({ classes, layers }: LegendsProps) {
...
+  useEffect(() => {
+    if (!!isAvailable && layerTypes.includes(type!)) {
+      map!.setPaintProperty(layerId, opacityType, opacity);
+      map!.setLayoutProperty(
+        layerId,
+        'visibility',
+        toggle ? 'visible' : 'none',
+      );
+    }
+  });
```
Ry-DS, line 141: This should have a dependency array

Ry-DS:
@dubwalla this might be a little hard to do;
There's two states here: 
1. Whether the layer is in the legend sidebar
2. Whether the layer is visible in general. 

To make the switch also turn off the switch in the navbar, we'd have to remove it from the legend sidebar immediately after its off. Think this is ok?

Ry-DS:
BTW: This doesn't support layers added by the Analysis UI. This might be hard to do based off how the analysis UI is written, but it might be worthwhile.

dubwalla:
thanks @Ry-DS for the input on this. sounds like short-term solution is to 'close' the layer (not toggle on/off) from the legend area. in the longer term, some awareness of the layer visibility would be better

dubwalla:
As just discussed in a call, let's have this PR focus on just the opacity issue and create another PR for to collapse a legend. Instead of toggling a layer off/on - we'll just collapse the layer legend entry

src/components/MapView/Legends/index.tsx
hav142: no need after i changed `find` with `some` based on your comment [here](https://github.com/WFP-VAM/prism-frontend/pull/138#discussion_r590468753). thanks

src/components/MapView/Legends/index.tsx
hav142: sorry did't realize this one, but i removed the lines related to toggle feature and will submit it to another pull request

src/components/MapView/Legends/index.tsx
hav142: i think it will not be null since there are many initial layers like boundaries, line, etc and if the map layer is not found in the list, the opacity feature should not be rendered

src/components/MapView/Legends/index.tsx
hav142: I tried to refactor based on your suggestion, but got into trouble trying to display the Analysis map layer. It has no `id` and `type` so I need to keep the optionals here. Once the AnalysisLayer is added to LayerType, I think it can be done

src/components/MapView/Legends/index.tsx
Ry-DS: got it, but as it is now, an exception might be thrown if we somehow end up with a null map. Doing something like:
```ts
if(!map){
return null;
} 
```
would be good enough :+1:

dubwalla:
@Ry-DS and @ericboucher - now that this PR is specifically for the opacity slider, I think we are good to go. Can you confirm?

Ry-DS (commented):
In terms of UI, this PR looks great :tada:! But we need to find a better way to actually display the opacity effect on the map. 

The current way its done - haphazardly accessing the map and changing the opacity in `Legends/index.tsx` isn't great because if we add new layer types, this will break very easily. Instead, we should use the existing `opacity` value we have in redux, because every layer is already written in a way to use this value (except the analysis layer, but I'll talk more about that later).

Every layer uses this variable to set its opacity: https://github.com/WFP-VAM/prism-frontend/blob/08cb71507eb18a5b67ec5244fd94fd0b2c539bf1/src/config/types.ts#L140

Therefore, adding a reducer here:
https://github.com/WFP-VAM/prism-frontend/blob/b556d8ccba86660780b19cfe09675184dfd8de54/src/context/mapStateSlice/index.ts#L43
to update the opacity field for a given layer ID would be perfect. 

The opacity in AnalysisLayer will need refactoring, because right now:
https://github.com/WFP-VAM/prism-frontend/blob/b47470c7fd5a1990fd9fb151cdfa6d4b7f153dbf/src/components/MapView/Layers/AnalysisLayer/index.tsx#L29
It is a fixed value (0.3).

This can be fixed by adding an `opacity` field to the analysisResultStateSlice here:
https://github.com/WFP-VAM/prism-frontend/blob/ca955c5cc5d3088d591b0d05719be0f604116231/src/context/analysisResultStateSlice.ts#L35
and writing the necessary reducers/selectors to update said field.  Then updating `AnalysisLayer` to use the new `opacity` field.

Hopefully this approach makes sense. Let me know if you have any questions :+1:

src/components/MapView/Legends/index.tsx
```diff
@@ -76,24 +88,87 @@ function Legends({ classes, layers }: LegendsProps) {
...
+  classes,
+  id,
+  title,
+  legend,
+  type,
+  opacity: initialOpacity,
+  children,
+}: LegendItemProps) {
+  const map = useSelector(mapSelector);
+  const [opacity, setOpacityValue] = useState<number | number[]>(
```
Ry-DS, line 101: You won't need this state anymore after what I discussed in the previous review (use redux state instead of local state)

hav142:
@Ry-DS I tried to implement the approaches that you mentioned (except for analysis layer, need to check further). overall seems good, but for WMS type layer there is a bit delay since the raster layer is a bit heavy. kindly suggest if there is a good way to fix this, or maybe i am still doing it wrong 🙏

Ry-DS:
image
I noticed on max opacity a scroll bar appears

Ry-DS:
> @Ry-DS I tried to implement the approaches that you mentioned (except for analysis layer, need to check further). overall seems good, but for WMS type layer there is a bit delay since the raster layer is a bit heavy. kindly suggest if there is a good way to fix this, or maybe i am still doing it wrong 🙏

Hm, maybe you can take the approach you had previously on the WMS layer?
So instead of setting it like this: https://github.com/WFP-VAM/prism-frontend/blob/b556d8ccba86660780b19cfe09675184dfd8de54/src/components/MapView/Layers/WMSLayer/index.tsx#L36
You would access the map directly and do it the hacky way. But this code should only live in `WMSLayer/index.tsx` since that would be the only layer which needs this approach. What do you think?

Ry-DS:
Deployed: http://prism-opacity.surge.sh

Ry-DS (approved):

ericboucher (approved):

Issue #19, opened 2020-04-10
Propose new UX for date selection


ericboucher:
The current date selection implementation is not ideal. Let's propose improvements from a UX and UI perspective.

*Current implementation*

<img width="1642" alt="Screen Shot 2020-04-09 at 5 45 31 PM" src="https://user-images.githubusercontent.com/16843267/78952289-ef2c7d80-7a89-11ea-91cd-0c12975a1e8d.png">

*Context*
Layers can have data available daily, once per month or every 10 days. We can only select dated layer at a time.

laveesh:
We have got two new designs for the date picker. 
![Design 1](https://user-images.githubusercontent.com/31915056/84622545-e4172300-aefa-11ea-8425-757ce364b499.png)
![Design 2](https://user-images.githubusercontent.com/31915056/84622514-d366ad00-aefa-11ea-84ef-a072b16a0242.png)

SimplyBallistic:
**UX Behind both designs**
 - We wanted to closely match the feel of Material UI, which is why we believe replacing the old date picker would be best here. 
 - Both designs we posted more closely match the design of Material UI while also allowing easy styling via CSS. 
 - We removed the bottom bar entirely as we believed it introduced unneeded complexity to the date selection. Moving everything to one dialog makes the process smoother.
 - The selector was moved to the top left to make it symmetrical with the Legend.

**UX Behind Design 1**
 - Design 1 aims to make the process centered to the user by taking the whole screen. 
 - The selector is part of Material UI/picker making it the closest match to PRISM's design. 
 - There are arrow buttons on the left and right so you can easily jump between possible dates instead of months, which makes more sense to the user.
 - While not shown here, we can disable dates so the user can see exactly which years and dates are available for them to choose from.

**UX Behind Design 2**
 - Design 2 instead moves everything from the old design to the top right corner.
 - There is instead a month drop down to allow you to see exactly which months are available to choose from. 
 - There is an non-intrusive overlay instead of taking the center of the page.

These designs are more so to provide ideas for the final solution. If you have a plan to mix and match both designs let us know!

SimplyBallistic:
**Other considerations**
We could also look into restyling the legend button so it looks more inline with Material UI, or converting it to a dropdown too.

dubwalla:
Thanks Ryan and Laveesh for taking this on - really appreciated! Some additional context:

[...]

PR #139, opened 2021-02-15
Change datepicker with timeline slider (#19)

hav142:
Based on issue #19 and #100 related to new UI for datepicker

Before:
image

After:
image


ericboucher:
@hav142 nice! How does it look on mobile?

hav142:
> 
> 
> @hav142 nice! How does it look on mobile?

Thanks @ericboucher. It should be responsive like this:

image

ericboucher:
Ok, let's add a bit of spacing maybe to make it look a bit better on mobile?

hav142:
@ericboucher I added some spacing and change the code structure a litle bit to make sure it works on all screen size

image

src/components/MapView/DateSelector/index.tsx
```diff
@@ -1,31 +1,39 @@
...
-import { getMonthStartAndEnd, isAvailableMonth, months } from './utils';
+import { findDateIndex } from './utils';
 import { dateRangeSelector } from '../../../context/mapStateSlice/selectors';
 
 interface InputProps {
   value?: string;
   onClick?: () => void;
 }
 
+const moment = extendMoment(Moment as any);
```
Ry-DS, line 35: Maybe TS has a point complaining about this? Since from the TS typing, importing moment as `import * as moment` makes the error go away. Is the `extendMoment` working without `import  * as moment`?

src/components/MapView/DateSelector/index.tsx
```diff
@@ -43,38 +51,148 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
   const dispatch = useDispatch();
   const { startDate: stateStartDate } = useSelector(dateRangeSelector);
 
-  const stateStartDateYear = moment(stateStartDate).year();
   const [selectedDate, setSelectedDate] = useState(moment(stateStartDate));
-  const selectedMonth = selectedDate.month();
+  const [dateRange, setDateRange] = useState([
```
Ry-DS, line 55: Can we type this? In the generic arguments `<>`. While its inferred, its a little complex to be be left to TS to guess.

src/components/MapView/DateSelector/index.tsx
```diff
@@ -43,38 +51,148 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
...
 
+  // Move the slider automatically so that the pointer always visible
   useEffect(() => {
+    let x = 0;
+    if (pointerPosition.x >= dateRange.length * 10 - timeLineWidth) {
+      // eslint-disable-next-line fp/no-mutation
+      x = timeLineWidth - dateRange.length * 10;
+    } else if (pointerPosition.x > timeLineWidth) {
+      // eslint-disable-next-line fp/no-mutation
+      x = -1 * pointerPosition.x + timeLineWidth / 2;
```
Ry-DS, line 80:
```suggestion
      x = -pointerPosition.x + timeLineWidth / 2;
```
does this work?

src/components/MapView/DateSelector/index.tsx
```diff
@@ -43,38 +51,148 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
...
+    }
+    if (
+      -1 * timelinePosition.x > pointerPosition.x ||
+      -1 * timelinePosition.x + timeLineWidth < pointerPosition.x
+    ) {
+      setTimelinePosition({ x, y: 0 });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Ry-DS, line 88: We should avoid ignoring this eslint rule, since it helps make sure we use `useEffect` as it was meant to be used, keeping its behavior predictable. I recommend [this answer](https://stackoverflow.com/a/58959298/5279269) to help gain a grasp why, it helped for me. Does the code work without this ignore?

src/components/MapView/DateSelector/index.tsx
```diff
@@ -43,38 +51,148 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
...
+  // Create timeline range and set pointer position
+  useEffect(() => {
+    const range = Array.from(
+      moment()
+        .range(
+          moment(stateStartDate).startOf('year'),
+          moment(stateStartDate).endOf('year'),
+        )
+        .by('days'),
+    ).map(e => {
```
Ry-DS, line 100: better name than `e`? e makes me think its an event

src/components/MapView/DateSelector/index.tsx
```diff
@@ -43,38 +51,148 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
...
   function updateStartDate(date: Date) {
     const time = date.getTime();
     dispatch(updateDateRange({ startDate: time, endDate: time }));
   }
 
+  function setDatePosition(date: number, increment: number) {
+    const dates = availableDates.map(d => {
+      return d + USER_DATE_OFFSET;
+    });
+    const selectedIndex = findDateIndex(dates, date as number);
```
Ry-DS, line 128: isn't  date already a number?

src/components/MapView/DateSelector/index.tsx
```diff
@@ -43,38 +51,148 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
...
+    if (dates[selectedIndex + increment]) {
+      updateStartDate(new Date(dates[selectedIndex + increment]));
+    }
+  }
+
+  // move pointer to closest date when change map layer
+  if (!isEqual(dateRef.current, availableDates)) {
+    setDatePosition(stateStartDate as number, 0);
```
Ry-DS, line 136: Its a little unusual to have `setState` outside of a `useEffect` or function. We should also not assume stateStartDate isn't undefined, since the typing tells us it could be.

Ry-DS: Maybe making the function setDatePosition handle undefined dates correctly would remove the need to cast.

src/components/MapView/DateSelector/index.tsx
```diff
@@ -43,38 +51,148 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
...
+  }
+
+  // Click on available date to move the pointer
+  const clickDate = (index: number) => {
+    const dates = availableDates.map(date => {
+      return date + USER_DATE_OFFSET;
+    });
+    const selectedIndex = findDateIndex(dates, dateRange[index].value);
+    if (selectedIndex >= 0 && dates[selectedIndex] !== stateStartDate) {
+      setPointerPosition({ x: index * 10, y: 0 });
```
Ry-DS, line 155: Can we make 10 a constant? It seems important and its used quite a lot.

src/components/MapView/DateSelector/index.tsx
```diff
@@ -43,38 +51,148 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
...
+      updateStartDate(new Date(dates[selectedIndex]));
+    }
+  };
+
+  // Set timeline position after being dragged
+  const onTimelineStop = (e: any, position: { x: number; y: number }) => {
```
Ry-DS, line 161: Try avoiding the any type. What could this be?

src/components/MapView/DateSelector/index.tsx
```diff
@@ -43,38 +51,148 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
...
+    setTimelinePosition(position);
+  };
+
+  // Set pointer position after being dragged
+  const onPointerStop = (e: any, position: { x: number; y: number }) => {
```
Ry-DS, line 166: You have a type definition for `{x: number, y: number}`. Moving it to a type variable and using it in your `useState` definitions and here also would clean things up.

src/components/MapView/DateSelector/index.tsx
```diff
@@ -90,48 +208,105 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
...
+                      TransitionComponent={Fade}
+                      TransitionProps={{ timeout: 0 }}
+                      placement="top"
+                      arrow
+                    >
+                      <Grid
+                        item
+                        xs
+                        className={
+                          date.isFirstday
```
Ry-DS, line 257:
```suggestion
                          date.isFirstDay
```

src/components/MapView/DateSelector/index.tsx
```diff
@@ -90,48 +208,105 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
...
+                  handle="#datePointerSelector"
+                  bounds={{
+                    top: 0,
+                    bottom: 0,
+                    left: 0,
+                    right: dateRange.length * 10,
+                  }}
+                  grid={[10, 1]}
+                  position={pointerPosition}
+                  onStart={(e: any) => e.stopPropagation()}
```
Ry-DS, line 297: possible to type e as a changeevent?

src/components/MapView/DateSelector/index.tsx
```diff
@@ -90,48 +208,105 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
...
+                  onStop={onPointerStop}
+                >
+                  <div className={classes.triangle} id="datePointerSelector" />
```
Ry-DS, line 300: the datepoint selector ID should be a constant string since its used in more than one place. Also, making this a MUI icon may help with making things look more like its from MUI.

src/components/MapView/DateSelector/index.tsx
```diff
@@ -172,6 +356,87 @@ const styles = (theme: Theme) =>
...
+    },
+
+    timeline: {
+      position: 'relative',
+      top: '5px',
+    },
+
+    dateItemFull: {
+      borderLeft: '1px solid white',
+      height: '36px',
```
Ry-DS, line 380: using theme.spacing would be nice where possible. Also, these could just be numbers if you are using pixel measurements.

src/components/MapView/DateSelector/index.tsx
```diff
@@ -172,6 +356,87 @@ const styles = (theme: Theme) =>
...
+    },
+
+    dateItem: {
+      borderTop: '1px solid white',
+      color: 'white',
+      position: 'relative',
+      top: '-5px',
+      cursor: 'pointer',
+      minWidth: '10px',
+      '&:hover': {
```
Ry-DS, line 399: this is a little repetitive from the full version. Making a constant which shares commonality would be nice.

src/components/MapView/DateSelector/utils.ts
```diff
@@ -60,3 +60,48 @@ export function isAvailableMonth(
...
   return findAvailableDayInMonth(month, year, availableDates).length > 0;
 }
+
+/**
+ * Return index of available dates that matched
+ * @param availableDates in millisecond format
+ * @param date in millisecond format
+ * @return
+ */
+export function findDateIndex(
```
Ry-DS, line 70: This *looks* like a binary search algorithm modified to find the closest match, but not too sure. If it is, are we assuming that availableDates is sorted? Might be worthwhile to document if so.

Ry-DS:
> Sorry to nitpick, but is there a Material UI design solution we can use for this? I like the functionality but the UI is not in harmony with other elements

Good first steps to help achieve this might be to choose a different variant for the dates (maybe matching the text in the navbar) and using a MUI icon for the arrow, with a color already used in the app.

dubwalla:
> > Sorry to nitpick, but is there a Material UI design solution we can use for this? I like the functionality but the UI is not in harmony with other elements
> 
> Good first steps to help achieve this might be to choose a different typography variant for the dates (maybe matching the text in the navbar) and using a MUI icon for the arrow, with a color already used in the app.

good recommendations - thanks Ryan. @hav142 - can you make an attempt to implement these suggestions?

src/components/MapView/DateSelector/index.tsx
hav142: is it correct?
`const [dateRange, setDateRange] = useState<DateRangeProps[]>([
    {
      value: 0,
      label: '',
      month: '',
      isFirstDay: false,
    },
  ]);
`
with DateRangeProps as
`type DateRangeProps = {
  value: number;
  label: string;
  month: string;
  isFirstDay: boolean;
};
`

src/components/MapView/DateSelector/index.tsx
hav142: It works if i include all the dependencies and remove the ignore. is that alright?

src/components/MapView/DateSelector/index.tsx
hav142: sorry my bad habit. i changed it to `date`.

src/components/MapView/DateSelector/index.tsx
hav142: i changed it to `DraggableEvent`

src/components/MapView/DateSelector/utils.ts
hav142: you are right, it is a binary search to find closest match. i updated the documentation

hav142:
> 
> 
> > > Sorry to nitpick, but is there a Material UI design solution we can use for this? I like the functionality but the UI is not in harmony with other elements
> > 
> > 
> > Good first steps to help achieve this might be to choose a different typography variant for the dates (maybe matching the text in the navbar) and using a MUI icon for the arrow, with a color already used in the app.
> 
> good recommendations - thanks Ryan. @hav142 - can you make an attempt to implement these suggestions?

is this good enough?

image

dubwalla:
Looks good from the screenshot - thank you!

Ry-DS (approved):
Looks good!
A few things we could consider:
- The DateSelector test is insanely large. Maybe we can mock some things so we can cut it down to something more reasonable?
- All good that we use a FontAwesome arrow icon, but it might be worthwhile to use the MUI icon instead for consistency.  Up to you and what you think looks best :+1:
- I think there's a few comments left from last review that you haven't responded to yet. If you have any quesitons, let me know :)

hav142:
>     * The DateSelector test is insanely large. Maybe we can mock some things so we can cut it down to something more reasonable?

i tried to add one more param to DateSelector component so it become like this `<DateSelector availableDates={selectedLayerDates} dateRangeBy="days" />`, then in the test file, i change it to `dateRangeBy="months"` so that the snapshot only create 12 month as mock instead of hundreds. and maybe it can be useful in the future if we want to display the timeline as monthly for some specific layer instead of daily. is this ok? what do you think?

>     * All good that we use a FontAwesome arrow icon, but it might be worthwhile to use the MUI icon instead for consistency.  Up to you and what you think looks best 👍

it's the same case with the one [here](https://github.com/WFP-VAM/prism-frontend/pull/131#discussion_r604695476). will update if using MUI icons is possible

>     * I think there's a few comments left from last review that you haven't responded to yet. If you have any quesitons, let me know :) They are quite small so I'm ok to overlook them if they don't seem worthwhile.

sorry i haven't replied them yet because i'm not sure about how to solve them:
1. [TS complaint](https://github.com/WFP-VAM/prism-frontend/pull/139#discussion_r593743694). not sure about this because i got no complaint in my local machine
2. [Constant for classes](https://github.com/WFP-VAM/prism-frontend/pull/139#discussion_r593746029). can you give an example on how to implement this one? 🙏

ericboucher:
For the DateSelector, you could spot which lib components take multiple lines and mock them in `setuptTest`

Another thing you can do is break down your component into a few different ones, and test them separately. So that you only test the component within the `.map` once. @Ry-DS any other ideas?

hav142:
> Another thing you can do is break down your component into a few different ones, and test them separately. So that you only test the component within the `.map` once. @Ry-DS any other ideas?

noted, i will try to break down the component first and see how is it going. Thanks @ericboucher

Ry-DS:
> 1. [TS complaint](https://github.com/WFP-VAM/prism-frontend/pull/139#discussion_r593743694). not sure about this because i got no complaint in my local machine
> 2. [Constant for classes](https://github.com/WFP-VAM/prism-frontend/pull/139#discussion_r593746029). can you give an example on how to implement this one? 🙏

1. Should have been more explicit haha. I assumed you were casting to `any` because if you didn't, TS would complain. I was thinking that maybe we could avoid this `any` cast by replacing `import Moment from 'moment'` with `import * as moment from 'moment';`. All good if this doesn't work, but might be worth a shot.
2.  The `&:hover` styles were repeated, so I was thinking something like:
```ts

const DATE_ITEM_STYLES: CreateCSSProperties = {
  borderTop: '1px solid white',
  color: 'white',
  position: 'relative',
  top: -5,
  cursor: 'pointer',
  minWidth: TIMELINE_ITEM_WIDTH,
  '&:hover': {
    borderLeft: '1px solid #5ccfff',
  },
};
const styles = () =>
  createStyles({
    dateItemFull: {
      ...DATE_ITEM_STYLES,
      borderLeft: '1px solid white',
      height: 36,
    },

    dateItem: merge(DATE_ITEM_STYLES, {
      '&:hover': {
        '& $dayItem': {
          borderLeft: 0,
        },
      },
    }),
// continued
```

ericboucher:
Great job @hav142 ! Let's do the two small changes @Ry-DS suggested and we should be good to go :)

hav142:
Thanks @Ry-DS for the clear guideline

>     1. Should have been more explicit haha. I assumed you were casting to `any` because if you didn't, TS would complain. I was thinking that maybe we could avoid this `any` cast by replacing `import Moment from 'moment'` with `import * as moment from 'moment';`. All good if this doesn't work, but might be worth a shot.

it works when i remove the cast and use your approach like this:
```
import * as Moment from 'moment';
...
const moment = extendMoment(Moment);
```
but then there is an error when running `yarn test`

image

Point no. 2 is good, but i need to install `@material-ui/styles` for using  `CreateCSSProperties`, and add new config in eslint: `import/no-extraneous-dependencies`. am i doing it right?

Ry-DS:
> 
> it works when i remove the cast and use your approach like this:
> 
> ```
> import * as Moment from 'moment';
> ...
> const moment = extendMoment(Moment);
> ```
> 
> but then there is an error when running `yarn test`

Ahh I see, you can keep the `any` cast then.
For the second point, I was able to use the type without a new dependency. I removed it in a new commit to see if the build passes. Let me know if you still face issues :+1:

PR #143, opened 2021-03-19
Add configurable AlertForm component

rysavym-amzn:
This is the UI component associated with the "alert" functionality described by [this product requirements doc](https://docs.google.com/document/d/1HiZo1YIVIk97AW2tqGmm0xBfEmWuk6zi9euGDpxlzmA/edit) as well as [this UI-specific doc](https://docs.google.com/document/d/1YUZlNsbSmAz2o4K7KMOIEq-pa8aPWyYBHQZyKVk8bFM/edit#heading=h.vd6eq8ctbhg8).

I am publishing a draft PR to document my progress and get some feedback.

### Outstanding Work

This is an MVP for the alerting UI to send a request to the backend to create an alert.

Some possible improvements for the future:

* Allow selection of "admin 1" level which selects all "admin 2" level zones
* Select regions by clicking on map

### What it does

Generates a JSON object that looks like this:

```json
{
  "email": "prism+test@ovio.org",
  "prism_url": "not-exactly-sure-how-to-get-this-yet",
  "alert_name": "Test",
  "min": 4,
  "max": 14,
  "alert_config": {
    "type": "wms",
    "serverLayerName": "ModisLST",
    "baseUrl": "https://mongolia.sibelius-datacube.org:5000",
    "wcsConfig": {
      "scale": 0.02,
      "offset": -273,
      "pixelResolution": 64
    }
  },
  "zones": {
    "zone_data": "goes_here"
  }
}
```

### Screenshots

alert_button
---
region_list_open
---
Screenshot from 2021-04-19 12-11-50


src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+  MenuItem,
+  Select,
+  TextField,
+  Theme,
+  Typography,
+  withStyles,
+  WithStyles,
+} from '@material-ui/core';
+import { grey } from '@material-ui/core/colors';
+
```
Ry-DS, line 17:
```suggestion
```
Avoid newlines in the import section, unless there's a clear separation between different assets. (e.g images vs components)

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+  const [emailValid, setEmailValid] = useState<boolean>(false);
+  const [email, setEmail] = useState<string>('');
+  const [selectedStat, setSelectedStat] = useState<string>('');
+  const [selectedThreshold, setSelectedThreshold] = useState<number>(0);
+
+  // Very basic regex for detecting emails
+  const emailRegex: RegExp = RegExp('.+@.+');
+
+  // On by default, may want to disable it in some cases
+  const isAlertFormEnabled = true;
```
Ry-DS, line 62: Constants should be at the top, outside of the functional component and (based off personal preference) cased as 'CAPITAL_SNAKE_CASE'

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+    | LayerData<BoundaryLayerProps>
+    | undefined;
+
+  const availableDates = useSelector(availableDatesSelector);
+
+  const [isAlertFormFormOpen, setIsAlertFormFormOpen] = useState(false);
+
+  // form elements
+  const [hazardLayerId, setHazardLayerId] = useState<LayerKey>();
+  const [selectedDate, setSelectedDate] = useState<number | null>(null);
```
Ry-DS, line 51: This variable isn't used. All good if there are plans to use it later :+1:

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+  const [regionsList, setRegionsList] = useState<string[]>(['allRegions']);
+  const [emailValid, setEmailValid] = useState<boolean>(false);
+  const [email, setEmail] = useState<string>('');
```
Ry-DS, line 54: To clean things up, you could get react to infer these types (`useState('')` infers to string)

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+  useEffect(() => {
+    const dates = hazardLayerId
+      ? availableDates[
+          (LayerDefinitions[hazardLayerId] as WMSLayerProps)?.serverLayerName
+        ]
+      : null;
+    if (!dates || dates.length === 0) {
+      setSelectedDate(null);
+    } else {
+      setSelectedDate(dates[dates.length - 1]);
```
Ry-DS, line 74: thoughts on using `lodash.last`?

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+    }
+  }, [availableDates, hazardLayerId]);
+
+  const regionNamesToNsoCodes: Map<string, string> = useMemo(() => {
+    if (!boundaryLayerData) {
+      // Not loaded yet. Will proceed when it is.
+      return new Map();
+    }
+
+    return new Map(
```
Ry-DS, line 84: Would be nice if we used a plain JS object. Since we cannot type a Map very strongly (right now its just <string,string>). Think that's possible? Or are there important reasons we are using a map?

Ry-DS: I'm not too fussed using Map here, since we won't be able to type things better anyway, but maybe `fromPairs` from lodash would work just as well here?

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+      boundaryLayerData.data.features
+        .filter(feature => feature.properties != null)
+        .map(feature => [
+          `${feature.properties?.ADM1_EN} / ${feature.properties?.ADM2_EN}`,
+          feature.properties?.NSO_CODE,
+        ]),
+    );
+  }, [boundaryLayerData]);
+
+  const regionNames: string[] = useMemo(() => {
```
Ry-DS, line 94: If you're using `useMemo` for performance, we likely don't need it here since the code is quite simple. Let me know if I'm wrong. 

If you wish to continue using them,  we should try combine this memo and the one above to output `{regionNames, regionNamesToNsoCodes}` since they share the same dependency array, and produce similar outputs (both produce `${feature.properties?.ADM1_EN} / ${feature.properties?.ADM2_EN}`)

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+        .map(
+          feature =>
+            `${feature.properties?.ADM1_EN} / ${feature.properties?.ADM2_EN}`,
+        )
+        // We can mutate the array since it's a new one anyway
+        .sort()
+    );
+  }, [boundaryLayerData]);
+
+  const onChangeEmail = (event: { target: { value: string } }) => {
```
Ry-DS, line 113: Type this with `React.ChangeEvent<HTMLInputElement>`

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+    const newEmail = event.target.value as string;
```
Ry-DS, line 114: If the above comment is taken into consideration, we won't need this cast.

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+      geotiff_url: '', // ^
+    };
+
+    console.log(apiData);
+
+    // TODO: Make API call
+    // await fetchApiData('http://localhost:80/alarms', apiData);
+  };
+
+  return isAlertFormEnabled ? (
```
Ry-DS, line 140: you should do:
```ts
if(!isAlertFormEnabled)
    return null;
return <...>
```
so this is easier to read.

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+const styles = (theme: Theme) =>
+  createStyles({
+    AlertForm: {
+      zIndex: theme.zIndex.drawer - 1, // position this below the Analysis drawer
+      position: 'absolute',
+      top: 40,
+      left: 2,
+      textAlign: 'left',
+    },
+    AlertFormMenu: {
```
Ry-DS, line 304: We are switching between PascalCase and camelCase here. Maybe we stick to just camelCase? Since I think we use that everwhere.

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+      '&.Mui-focused': { color: 'white' },
+    },
+    selector: {
+      margin: '5px',
+    },
+    numberField: {
+      paddingLeft: '10px',
+      marginTop: '10px',
+      width: '85.5px',
+      '& .Mui-focused': { color: 'white' },
```
Ry-DS, line 345: We should avoid using `& ...Mui class-name` since its a little prone to breaking. Overriding this style in the global theme (where we have access to TS checking) is the best route. If you don't want this style to apply globally, then we could keep this. 
Also applies to the 2 other cases in this file.

Ry-DS (commented):
Looks good! Like how you added comments throughout where one would expect. Let me know if you have any questions :+1:

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+
+  const [isAlertFormFormOpen, setIsAlertFormFormOpen] = useState(false);
+
+  // form elements
+  const [hazardLayerId, setHazardLayerId] = useState<LayerKey>();
+  const [selectedDate, setSelectedDate] = useState<number | null>(null);
+  const [regionsList, setRegionsList] = useState<string[]>(['allRegions']);
+  const [emailValid, setEmailValid] = useState<boolean>(false);
+  const [email, setEmail] = useState<string>('');
+  const [selectedStat, setSelectedStat] = useState<string>('');
```
Ry-DS, line 55: If this value can only take specific options (maximum, minimum), can we type this stronger than `string`?

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,354 @@
...
+                    value="allRegions"
+                  >
+                    All
+                  </MenuItem>
+                  <ListSubheader>Administrative Regions</ListSubheader>
+                  ...
+                  {regionNames.map(region => {
+                    return (
+                      <MenuItem
+                        style={{ color: 'black' }}
```
Ry-DS, line 259: I noticed when I wrote `LayerDropdown` I was doing the same thing, and styling this as black. Maybe we can style this once and for all as black in the global MUI style, and remove all the inline styles both here and in `layerdropdown`?

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,353 @@
...
+  const [hazardLayerId, setHazardLayerId] = useState<LayerKey>();
+  const [selectedDate, setSelectedDate] = useState<number | null>(null);
+  const [regionsList, setRegionsList] = useState<string[]>(['allRegions']);
+  const [emailValid, setEmailValid] = useState<boolean>(false);
+  const [email, setEmail] = useState<string>('');
+  const [selectedStat, setSelectedStat] = useState<string>('');
+  const [selectedThreshold, setSelectedThreshold] = useState<number>(0);
+
+  // Very basic regex for detecting emails
+  const emailRegex: RegExp = RegExp('.+@.+');
```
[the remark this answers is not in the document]
rysavym-amzn: Sure thing. When I use regexes for input validation I try not to overcomplicate things ([which seems to be common guidance for email addresses](https://stackoverflow.com/questions/201323/how-to-validate-an-email-address-using-a-regular-expression)) but I am happy to add something more strict.

src/components/MapView/AlertForm/index.tsx
rysavym-amzn: can do! I'm not super familiar with this so I was being explicit.

src/components/MapView/AlertForm/index.tsx
rysavym-amzn: Sure, I will look at making this an enum or something similar.

src/components/MapView/AlertForm/index.tsx
rysavym-amzn: Hadn't really thought about it - this code originates in `Analyser`. I can clean it up some

src/components/MapView/AlertForm/index.tsx
rysavym-amzn: I will look at combining them; and it's true that the code is quite simple although the underlying dataset is quite large so it is pretty slow (at least anecdotally, on my machine) so I was hoping to only do it once. I'll take another look.

src/components/MapView/AlertForm/index.tsx
rysavym-amzn: Sure thing; this may have been a find-and-replace mistake

src/components/MapView/AlertForm/index.tsx
rysavym-amzn: I did not end up combining them, but I made them (hopefully) easier to understand.

The reason I have a separate list is because one of these is a sorted list of strings, and one is an object whose keys are those strings. Unfortunately I don't think there's really a way to have the object keys sorted in memory and the semantics of `.sort()` (since it's self-mutating) make me think doing this each render is probably wasteful.

If I am understanding `useMemo` and using it correctly, performance-wise it seems pretty good to me how it is. But I'm still happy to change this if you want.

src/components/MapView/AlertForm/index.tsx
rysavym-amzn: I was able to delete 2 of the 3 occurrences here. I think putting this in the global theme is a good idea.

src/components/MapView/AlertForm/index.tsx
Ry-DS: If it helps performance, all good (useMemo's indented use!) :+1:

src/components/MapView/AlertForm/index.tsx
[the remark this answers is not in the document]
Ry-DS: that sounds fair. If there's strong validation (on the API at least) to ensure no errors happen down the line this should be good :+1:

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+import { faBell, faCaretDown } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { useSelector } from 'react-redux';
+import { getBoundaryLayerSingleton } from '../../../config/utils';
+import { BoundaryLayerProps, LayerKey } from '../../../config/types';
+import { LayerData } from '../../../context/layers/layer-data';
+import { layerDataSelector } from '../../../context/mapStateSlice/selectors';
+import LayerDropdown from '../Layers/LayerDropdown';
+
+// Not fully RFC-compliant, but should filter out obviously-invalid emails.
```
Ry-DS, line 25: Linking the source would be great ;)

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+const ALERT_FORM_ENABLED = true;
+
+enum Statistic {
+  Maximum,
+  Minimum,
+  Median,
+  Mean,
+}
+
+enum Comparator {
```
Ry-DS, line 40: I recommend doing:
```ts
enum Comparator {
    GREATER_THAN = '>',
    GREATER_THAN_OR_EQUAL = '>=',
    LESS_THAN = '<'
// etc
}
```

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+  const [isAlertFormFormOpen, setIsAlertFormFormOpen] = useState(false);
+
+  // form elements
+  const [hazardLayerId, setHazardLayerId] = useState<LayerKey>();
+  const [regionsList, setRegionsList] = useState<string[]>(['allRegions']);
+  const [emailValid, setEmailValid] = useState<boolean>(false);
+  const [email, setEmail] = useState('');
+  const [selectedStat, setSelectedStat] = useState<Statistic>();
+  const [selectedComparator, setSelectedComparator] = useState('');
```
Ry-DS, line 62: after changing the enum, this should be `useState<Comparator>()` and be used similar to how to setup `selectedStat`

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+// Not fully RFC-compliant, but should filter out obviously-invalid emails.
+const EMAIL_REGEX: RegExp = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+
+// This should probably be determined on a case-by-case basis,
+// depending on if the downstream API has the capability.
+// For now it can be permanently enabled.
+const ALERT_FORM_ENABLED = true;
+
+enum Statistic {
```
Ry-DS, line 33: Is it possible to use this: https://github.com/WFP-VAM/prism-frontend/blob/6b51b51ba9532a25614fd35395f55c02fb1455e6/src/config/types.ts#L219
And add `max` and `min` to it?

If the analysis tool doesn't support max and min, you could split those into a new enum and make:
```ts
enum MaxMin{
Max='max',Min='min'
}
type Statistic= MaxMin | AggregationOption;
```

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+    setEmailValid(!!newEmail.match(EMAIL_REGEX));
+    setEmail(newEmail);
+  };
+
+  const runAlertForm = async () => {
+    const regionCodes = regionsList
+      .filter(r => r !== 'allRegions')
+      .map(r => regionNamesToNsoCodes[r]);
+
+    const apiData = {
```
Ry-DS, line 99: We should define an API type interface and keep that at the top

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+    // await fetchApiData('http://localhost:80/alarms', apiData);
+  };
+
+  if (!ALERT_FORM_ENABLED) {
+    return null;
+  }
+
+  return (
+    <div className={classes.alertForm}>
+      <Button
```
Ry-DS, line 122: This dropdown code was copied from `Analyser` right? Maybe we can make a new component which allows anyone to make dropdowns like the Analysis menu?

src/components/MapView/AlertForm/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+        variant="contained"
+        color="primary"
+        onClick={() => {
+          setIsAlertFormFormOpen(!isAlertFormFormOpen);
+        }}
+      >
+        <FontAwesomeIcon
+          style={{ marginRight: '10px', fontSize: '1.6em' }}
```
Ry-DS, line 130: Also, when I wrote this I didn't do it that well. I think we could have used `startIcon` and `endIcon` in `Button` instead of these inline styles. When you make a component which combines the dropdown menu from Analyser and AlertForm, you could explore using these props. (startIcon and endIcon explained here: https://material-ui.com/api/button/)

src/components/MapView/AlertForm/index.tsx
rysavym-amzn: Great suggestions! However after refactoring this selector again after talking to @ericboucher I've removed these enums.

src/components/MapView/AlertForm/index.tsx
rysavym-amzn: I updated this to match what is now in the Analyser dropdown.

src/components/MapView/AlertForm/index.tsx
rysavym-amzn: Perhaps, although in my conversations with @ericboucher it sounds like maybe the way to create alerts would ideally not be a dropdown anyway - it might be it's own view entirely. So I'm not sure how useful a refactor like this would be at the moment.

src/utils/analysis-utils.ts
```diff
@@ -138,8 +138,10 @@ export const checkBaselineDataLayer = (
   );
 };
 
+export type ApiRequest = {};
```
ericboucher, line 141: Not sure I get the purpose of these changes?

mrysav: Previously, the fetch method took an ApiData object. ApiData is significantly different than the request I want to make to post an alert, but the fetch method is not specific to ApiData. So I made a type that both ApiData and the new request type inherit from, and changed the fetch method to take that instead.

I'm open to changing this if you think it doesn't make sense.

ericboucher: Hmm, as a first thought I would create the request that you want and give the two request type more explicit names and then have the fetch command take a type of "RequestType1 | RequestType2" ?

mrysav: Good suggestion - I updated this in the latest commit

ericboucher:
Looking great @mrysav , a few improvements and we should be good to go:

- It should be quite rare but let's make sure we handle errors properly
- After submitting the alert, check for success and show a success message to the user
- Fix the glitch/error after submission of the alert, see screenshot below (the alert submits properly but there is an error in the console):

Screen Shot 2021-04-22 at 3 36 44 PM


I deployed on https://staging-prism-frontend.web.app/

rysavym-amzn:
I just pushed some commits that have some minor changes to `fetchApiData` which I believe will fix the alert submit glitch - in addition, I display the response from the server on successful submit. Hopefully that wraps it up!

ericboucher (approved):

PR #145, opened 2021-03-28
Share Analysis and Import Analysis by URL

yapalgit:
My first commit, I hope as you want ! It can be easy to do the same feature for the MapView layers too, but I need some details about some layers to export or import like WMS type.


src/components/MapView/Analyser/index.tsx
```diff
@@ -78,32 +81,128 @@ function Analyser({ classes }: AnalyserProps) {
...
     const dates = hazardLayerId
       ? availableDates[
           (LayerDefinitions[hazardLayerId] as WMSLayerProps)?.serverLayerName
         ]
       : null;
-    if (!dates || dates.length === 0) {
-      setSelectedDate(null);
-    } else {
-      setSelectedDate(dates[dates.length - 1]);
+    if (!urlParams[UrlParamsEnum.DATE]) {
```
Ry-DS, line 198: This useEffect is responsible for selecting a default date on the datepicker when the user changes their hazard layer selection.
This isn't a good idea because now, when the user selects a new layer, the date picker doesn't update with compatible dates. 
Since it'll be hard to make this use effect work with URL Queries, the best way I see is to edit the  `LayerDropdown` `setValue` function and add the functionality there, so when a new value is picked, the date selector is updated as well from before.

src/components/MapView/Analyser/index.tsx
```diff
@@ -153,40 +244,24 @@ function Analyser({ classes }: AnalyserProps) {
...
-      extent: adminBoundariesExtent,
-      threshold: {
-        above: parseFloat(aboveThreshold) || undefined,
-        below: parseFloat(belowThreshold) || undefined,
-      },
-    };
-
-    await dispatch(requestAndStoreAnalysis(params));
+  const shareAnalysis = () => {
+    navigator.clipboard.writeText(
```
Ry-DS, line 248: Remember when I said in a previous comment we should make all the fields live in *one* state object? This is because now, we can construct this url in a scalable way, since we'll have access to the value *and* the key.
```ts
Object.entries(formFields).reduce((search, [key, value])=>search+=key+'='+value+'&','?');
```
This is very rough (doesn't stop adding the `&` at the end but hoping you get the idea ;)

Ry-DS (changes requested):
Good start! We just need to address a few things:
- Try moving all the form fields into one state like: `const [formFields, setFormFields] = useState<FormFields>(/* Initial URL Query here*/)`. This will make it much easier to generate urls and also load the initial values, cause right now what we are doing is accessing these states one by one. This means if we add a new field in the future, updating this code won't be too straightforward. Discussed here: https://github.com/WFP-VAM/prism-frontend/pull/145#discussion_r604640918
- Ensure that if you select a new hazard layer, the date selector updates and automatically selects the closest date. This would mean you have to refactor the `useEffect` that does this right now and possibly move the code to the `setValue` instead. Discussed here: https://github.com/WFP-VAM/prism-frontend/pull/145#discussion_r604649155

Let me know if you have any questions!

src/components/MapView/Analyser/index.tsx
```diff
@@ -360,6 +435,21 @@ function Analyser({ classes }: AnalyserProps) {
                 <Typography variant="body2">Run Analysis</Typography>
               </Button>
             )}
+            {analysisResult && (
+              <Button
+                className={classes.innerAnalysisButton}
+                onClick={shareAnalysis}
+                disabled={
+                  !!thresholdError || // if there is a threshold error
```
Ry-DS, line 443: Wasn't this copied from right above? We should put this large boolean expression in a constant

Issue #137, opened 2021-02-08
Some Features implemented in idn_dev branch


reed1:
Hi everyone,

This post is intended to give you an overview of new features we have implemented in idn_dev:

1. Download Map as CSV / GeoTiff #124 
2. NSO / Choropleth Layer with date dimension #112 
3. Custom Map Popup Content (generated by http backend)
4. Group data layer, for example 1Day,2Days,3Days calculation should be grouped under 1 layer with simple day selection #122 

I squashed to the commit here: aa10796 If something is beneficial to be merged then I can make a pull request. Or if the feature is implemented in master with different or better approach then we can use that too. Feel free to cherry-pick / copy from the branch. I'll try to explain briefly on each point.

**1. Download Map as CSV / GeoTiff**

We've tried to dowload raw geotiff for a Layer using this guide: https://docs.geoserver.org/stable/en/user/community/wps-download/rawDownload.html. But then the resulting file is somewhat different from original tiff uploaded to GeoServer. It's still an ongoing issue, as such we need to make a custom API endpoint to download the original GeoTiff

```
{ ...
"server_layer_name": "rainfall_extreme_daily",
"base_url": "https://vampire.idn.wfp.org:8081/geoserver/vampire",
"download_url": "http://vampire.idn.wfp.org:3002/download/rainfall_extreme?type=daily&numday={DIM_NUMDAYSOURCE}",
... }
```

![a](https://user-images.githubusercontent.com/1814324/107163653-6057fc80-69dd-11eb-9fc1-8e2becca7e1c.png)

The cons is that we need to configure the backend and layer configuration a bit, but the benefit is that we can customize what the resulting file will contain, for example in CSV output we can add more columns

![image](https://user-images.githubusercontent.com/1814324/107164145-3c49ea80-69e0-11eb-9482-ece5857a571f.png)

**2. NSO / Choropleth Layer with date dimension #112**

```
{ ...
"title": "Impact",
"type": "nso",
"source": "https://vampire.idn.wfp.org/home/get_layer2/25?start={YYYY-MM-DD}&end={YYYY-MM-DD}",
"has_date": true,
"date_url": "https://vampire.idn.wfp.org/home/get_date/25?start=%22&end=2020-01-02",
...}
```

There are 2 added properties here:
- has_date: Whether the NSO layer is static
- date_url: Date availability API

The source URL can have date parameter within { }, the format within the bracket will be used by momentjs to determine the resulting value

[...]

PR #147, opened 2021-03-31
Move config into country specific subfolders

ericboucher:
The ultimate goal is to have all country configs in master to avoid merge conflicts as we start working on more countries simultaneously.


README.md
```diff
@@ -7,21 +7,27 @@ This project is the front-end interface for the World Food Programme's [PRISM pr
 ## Functionalities
 
 The new PRISM frontend is built as a static website to minimize cross dependencies and simplify deployments as much as possible. Currently, PRISM frontend provides the ability to:
+
 - Load administrative boundaries as GeoJSON (`src/config/admin_boundaries.json`)
 - Load baseline data as JSON, and link it to administrative boundaries
 - Display WMS layers from Geoserver or Open Data Cube endpoints, with date selection capabilities
 - Display CSV tables in a left side panel
 
+To chose which country to run PRISM for, you can set the environment variable `REACT_APP_COUNTRY`. Currently, available configurations are `mongolia`, `indonesia`. The default is `mongolia`.
```
Ry-DS, line 16: Not really required, but if its a list
- Item 1
- Item 2

Itll be easy to add to in the future.

src/components/MapView/index.tsx
```diff
@@ -40,7 +40,7 @@ import {
   loadAvailableDates,
 } from '../../context/serverStateSlice';
 
-import appConfig from '../../config/prism.json';
+import { appConfig } from '../../config/config';
```
Ry-DS, line 43: What do you think of using `index.ts` inside the config folder so there isn't a double config?

src/config/config.ts
```diff
@@ -0,0 +1,55 @@
...
+    appConfig: myanmarConfig,
+    rawLayers: myanmarRawLayers,
+    rawTables: myanmarRawTables,
+  },
+} as const;
+
+type Country = 'indonesia' | 'mongolia' | 'mozambique' | 'myanmar';
+
+const { REACT_APP_COUNTRY: COUNTRY } = process.env;
+const safeCountry = (COUNTRY && Object.keys(configMap).includes(COUNTRY)
```
Ry-DS, line 49: Maybe we could use https://lodash.com/docs/4.17.15#has

src/config/config.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import {
+  mozambiqueConfig,
+  mozambiqueRawLayers,
+  mozambiqueRawTables,
+} from './mozambique';
+
+import { myanmarConfig, myanmarRawLayers, myanmarRawTables } from './myanmar';
+
+const DEFAULT = 'mongolia';
```
Ry-DS, line 21: `const DEFAULT: Country=...`?

Ry-DS (approved):
Looks good! :+1:

ericboucher:
@hav142 @dubwalla @JorgeMartinezG this is a big change but will make development easier and avoid merge conflicts in the future.

Instead of having a master branch per country, we will use `master` to hold all the configs. The app will default to Mongolia but running another country will be as simple as 

`REACT_APP_COUNTRY=indonesia yarn start` once this PR is merged

dubwalla:
@mwarizmiwafiq @reed1 - please have a look. this change will impact how country level configurations are managed and simplify merging in the future

reed1:
Thanks for notifying Amit, this will indeed makes development easier. If I may, I have several inputs on this:
    
- Country specific subfolders to be added to gitignore and make each country their own git repository to store config. This will make the repository consists of frontend code only. We can also use git-submodule, but I think cloning twice (for frontend and for country config) is simpler.

- Or we can also use dropbox like service to store and serve files, especially for big json and csv. It will help shrinking the frontend repo size.

For now we're still using idn_branch for features like #137. Should I make pull requests for features implemented there? After the features are merged, we can use master and remove idn_dev branch.

dubwalla:
Thanks for the feedback Ridwan. I think the new structure with subfolders per country within the data folder covers this sufficiently. We should continue to have the data folder content available in GitHub if thats what the configuration points to. You can always ignore folders that aren't relevant for your locally. 

I think it's also ok to store large data elsewhere. S3 would probably be the preferred solution. @ericboucher - thoughts?

dubwalla:
> For now we're still using idn_branch for features like #137. Should I make pull requests for features implemented there? After the features are merged, we can use master and remove idn_dev branch.

Yes, you should be creating PRs for your work as a draft PR if still in progress, or as a PR if ready for review.

ericboucher:
@reed1 thanks for your input. Yes, please transition all future PRs to point towards the master branch.

I think having one repo with all the configs will be useful to let developers check that their features are working in other countries as well.

If some data files are/get too big, we will move them to S3 like we are doing for some of the admin boundaries.

Let me know if you have any other questions :)

PR #152, opened 2021-04-02
Adam integration with grouped layers

JorgeMartinezG:
This PR includes WFP ADAM layers into PRISM.

## **Changes**
- Added field group on layers type schema.
- Included adam earthquake layer.
- Included adam tropical storms layers, displayed as single layer.
- Modified date parser to remove time values.

image

You can see demo of this PR here: http://homeless-giraffe.surge.sh/

### DIFF
--- a/src/context/mapStateSlice/index.ts
+++ b/src/context/mapStateSlice/index.ts
@@ -1,6 +1,7 @@
 import { createSlice, PayloadAction } from '@reduxjs/toolkit';
 import { Map as MapBoxMap } from 'mapbox-gl';
 import { LayerType } from '../../config/types';
+import { LayerDefinitions } from '../../config/utils';
 import { LayerData, LayerDataTypes, loadLayerData } from '../layers/layer-data';
 
 interface DateRange {
@@ -41,10 +42,30 @@ export const mapStateSlice = createSlice({
   name: 'mapState',
   initialState,
   reducers: {
-    addLayer: ({ layers, ...rest }, { payload }: PayloadAction<LayerType>) => ({
-      ...rest,
-      layers: layers.filter(layer => keepLayer(layer, payload)).concat(payload),
-    }),
+    addLayer: ({ layers, ...rest }, { payload }: PayloadAction<LayerType>) => {
+      // Check if a layer belongs to a group.
+      if (!payload.group) {
+        return {
+          ...rest,
+          layers: layers
+            .filter(layer => keepLayer(layer, payload))
+            .concat(payload),
+        };
+      }
+
+      const { name } = payload.group;
+      // Get all layers that belong to a group.
+      const groupedLayer = Object.values(LayerDefinitions).filter(
+        l => l.group?.name === name,
+      );
+
+      return {
+        ...rest,
+        layers: layers
+          .filter(layer => keepLayer(layer, payload))
+          .concat(groupedLayer),
+      };
+    },
 
     removeLayer: (
       { layers, ...rest },
--- a/src/components/MapView/index.tsx
+++ b/src/components/MapView/index.tsx
@@ -154,6 +154,10 @@ function MapView({ classes }: MapViewProps) {
             .map(date => moment(date).format('YYYY-MM-DD'))
             .includes(moment(selectedDate).format('YYYY-MM-DD'))
         ) {
+          if (layer.group && layer.group.main === false) {
+            return;
+          }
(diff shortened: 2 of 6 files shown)


ericboucher:
When clicking on a past month, we get the following errors:

Screen Shot 2021-04-05 at 4 04 50 PM

Are they avoidable? Could we directly select a valid date?

JorgeMartinezG:
@ericboucher Fixed issue. Basically it was popping alerts for each layer. Just modified the group schema which specified if a layer is the main one. Branch was rebased wrt master.

Also, used a subset of the global layer to be consistent with a given country. For this example, I used mozambique. You can see it within the demo page: http://homeless-giraffe.surge.sh/

dubwalla:
Looks good, thanks Jorge. One thing I'm seeing is that when you click on a month, it requests a layer for the first day of the month. If a tropical storm happens on a day besides the first of the month, clicking on a month results in the mismatched date popup alert. It would be good if interaction by month fetches whatever the first event within a month is instead of just the first of the month. I'll put this into a separate issue

dubwalla:
Added issue https://github.com/WFP-VAM/prism-frontend/issues/153 for the date mismatch error @ericboucher 
 @JorgeMartinezG

ericboucher:
That is already the expected behavior if I recall correctly so something else must be happening here. Or maybe it's because we select multiple layers at once and maybe their dates are not matching?

JorgeMartinezG:
Master has the same behaviour. the logic for this is set here: https://github.com/WFP-VAM/prism-frontend/blob/master/src/components/MapView/index.tsx#L148

Related to ADAM tropical storms layers, tracks and wind buffers are generated from nodes layers, using the same date.

ericboucher:
Hmm I think you're right, maybe it's just because I am used to the Mongolia version which has all layers with data on the first

https://prism-mongolia.org/

src/context/mapStateSlice/index.ts
```diff
@@ -41,10 +42,34 @@ export const mapStateSlice = createSlice({
   name: 'mapState',
   initialState,
   reducers: {
-    addLayer: ({ layers, ...rest }, { payload }: PayloadAction<LayerType>) => ({
-      ...rest,
-      layers: layers.filter(layer => keepLayer(layer, payload)).concat(payload),
-    }),
+    addLayer: ({ layers, ...rest }, { payload }: PayloadAction<LayerType>) => {
+      // Validate if layer belongs to a group.
```
ericboucher, line 46: Validate -> Check if a layer

src/context/mapStateSlice/index.ts
```diff
@@ -41,10 +42,34 @@ export const mapStateSlice = createSlice({
...
+          layers: layers
+            .filter(layer => keepLayer(layer, payload))
+            .concat(payload),
+        };
+      }
+
+      const { name } = payload.group;
+      // Get all layers that belong to a group.
+      const items = Object.values(LayerDefinitions).filter(l => {
+        const groupName = l.group ? l.group.name : undefined;
```
ericboucher, line 59: You could simply do a one line:

`l.group?.name === name`

ericboucher (approved):
Just a few cleanups, looks good!

src/context/mapStateSlice/index.ts
```diff
@@ -41,10 +42,32 @@ export const mapStateSlice = createSlice({
...
+      }
+
+      const { name } = payload.group;
+      // Get all layers that belong to a group.
+      const items = Object.values(LayerDefinitions).filter(l => {
+        if (l.group?.name === name) {
+          return true;
+        }
+
+        return false;
```
ericboucher, line 63: I meant that you can go even one step further and just:

`Object.values(LayerDefinitions).filter(l => (l.group?.name === name))`

src/context/mapStateSlice/index.ts
```diff
@@ -41,10 +42,32 @@ export const mapStateSlice = createSlice({
...
+          ...rest,
+          layers: layers
+            .filter(layer => keepLayer(layer, payload))
+            .concat(payload),
+        };
+      }
+
+      const { name } = payload.group;
+      // Get all layers that belong to a group.
+      const items = Object.values(LayerDefinitions).filter(l => {
```
ericboucher, line 58: maybe rename `items` to `groupedLayer` or similar?

PR #154, opened 2021-04-08
Ignore secondary layers available dates

JorgeMartinezG:
This PR fixes date conflicts when a group of layers is selected.

The main change corresponds to remove secondary layers from the **selectedLayerswithDates** array. This disables the duplicate dates check and also the update of the **selectedDate** variable for secondary layers, avoiding dates conflicts.

Since dates are not considered for secondary layers, another change is to use buffers and tracks from global Adam tropical storms layers. Mapbox by default fetches tiles given a bounding box. and time value is still specified for all layers within a group. 

Deployment of this PR is found here: http://homeless-giraffe.surge.sh/


dubwalla (approved):
Looks good, thanks!

src/config/mozambique/layers.json
```diff
@@ -166,36 +166,36 @@
...
     "type": "wms",
-    "server_layer_name": "wld_gdacs_tc_events_buffers_moz",
+    "server_layer_name": "wld_gdacs_tc_events_buffers",
     "opacity": 0.9,
     "legend": [],
     "legend_text": "",
     "group": {
       "name": "adam",
-      "main": true
+      "main": false
```
ericboucher, line 178: Maybe just keep it as undefined and default to "false" later on?

ericboucher (approved):

PR #163, opened 2021-04-21
Form feature for WMS Layer

reed1:
image

This will add form for WMS Layer for additional Dimension ([AdditionalDomainAttributes](https://docs.geoserver.org/latest/en/user/data/raster/imagemosaic/configuration.html)). In Rainfall Extreme we want to see different result for different daily simulation(1, 2, 3, 4, 5days) and also different release hour for the prediction. To avoid adding one layer for each day we can use a simple selection form. Also for uniform hour availability on each day, we can use this instead of using full timestamp support as the XML text for date availability will get multiplied 

I'm using Mongolia - Pasture anomaly layer to demonstrate the config and impact on WMS image request, it will be removed upon PR passes the review

### DIFF
--- a/src/components/MapView/Layers/WMSLayer/index.tsx
+++ b/src/components/MapView/Layers/WMSLayer/index.tsx
@@ -1,9 +1,12 @@
 import React from 'react';
+import { useSelector } from 'react-redux';
+import { fromPairs } from 'lodash';
 import moment from 'moment';
 import { Layer, Source } from 'react-mapbox-gl';
 import { WMSLayerProps } from '../../../../config/types';
 import { getWMSUrl } from '../raster-utils';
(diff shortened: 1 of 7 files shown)


dubwalla:
@reed1 - can you please link this back to specific issues this PR will address?

@JorgeMartinezG & @ericboucher - can you have a look at this PR?

reed1:
Hi Amit, it was on issue #137. The demo can be seen here : http://3.134.83.146:3001/ on Flood Alert layer

This feature is to enable parameterizing GeoServer url for custom dimension (just like elevation dimension) via simple selection in Legend Box. Without the feature, we'll need to create separate layers:
- flood alert 1 day
- flood alert 2 days
- flood alert 3 days
- flood alert 4 days
- flood alert 5 days

PR #164, opened 2021-04-21
NSO Layer with date support

reed1:
This will enable NSO Layer to have date support, but only when `date_url` exists on Layer configuration. The path url can now have date parameter within `{ }` brackets:

```
    "title": "Poverty headcount (aimag)",
    "type": "nso",
    "path": "../data/mongolia/nso/NSO_Poverty_Headcount_Admin1.json?date={YYYY-MM-DD}",
    "date_url": "../data/mongolia/nso/nso_sample_dates.json",
```

I'm using Mongolia / Poverty headcount layer to demonstrate the functionality and impact on data fetch request, it will be removed on merge


dubwalla:
@ericboucher and/or @Ry-DS - can you please review this PR?

@reed1 - where is the issue this PR is linked to?

src/context/layers/nso.ts
```diff
@@ -33,9 +34,15 @@ export const fetchNSOLayerData: LazyLoader<NSOLayerProps> = () => async (
     throw new Error('Boundary Layer not loaded!');
   }
   const adminBoundaries = adminBoundariesLayer.data;
+  const datedPath = path.replace(/{.*?}/g, match => {
```
Ry-DS, line 37: Comments explaining this with sample input-output would help here

src/utils/server-utils.ts
```diff
@@ -225,6 +231,16 @@ async function getPointDataCoverage(layer: PointDataLayerProps) {
   return possibleDates;
 }
 
+async function getNSOAvailableDates(layer: NSOLayerProps) {
+  const url = layer.dateUrl!;
```
Ry-DS, line 235: why can we assume this is not null? IMO might be safer to try stick to the TS definitions (add a null check)

Ry-DS: Also, is it possible to link this with the `getPossibleDatesForLayer` which we already have?

src/utils/server-utils.ts
```diff
@@ -225,6 +231,16 @@ async function getPointDataCoverage(layer: PointDataLayerProps) {
   return possibleDates;
 }
 
+async function getNSOAvailableDates(layer: NSOLayerProps) {
```
Ry-DS, line 234: We should add a comment or rename this function in such a way that other devs won't get confused and use this to get the dates instead of `getPossibleDatesForLayer`, since by how its framed, it could be used independently. 

Maybe a comment highlighting the data this outputs is being saved in the server state slice as `availableDates`?

Ry-DS (commented):
Nice! Left some comments to improve readability

reed1:
Updated, thanks for your input.

Renamed `getNSOAvailableDates` to `getNSOCoverage` to look like the others. Also it's not an exported function.

I chose the `{"dates": ["YYYY-MM-DD", ...]}` format for backend because from what I heard top level JSON Array could lead to [vulnerability](https://medium.com/@rm.dev/securing-json-apis-with-wrapper-objects-top-level-json-array-security-issue-1cf855468b8c), not really applicable here, but better be safe

reed1:
feature for IDN branch

PR #165, opened 2021-04-21
Extra Download url(s) for Layers

reed1:
image

```
...
    "title": "Pasture anomaly",
    "type": "wms",
    "server_layer_name": "ModisAnomaly",
    "base_url": "https://mongolia.sibelius-datacube.org:5000",
    "downloads": [
      {
        "label": "GeoTiff",
        "url": "https://example.com/download?layer=poverty&format=tiff"
      },
      {
        "label": "CSV",
        "url": "https://example.com/download?layer=poverty&format=csv"
      }
    ],
    "legend": [
...
```

We've tried to dowload raw geotiff for a Layer using [this guide](https://docs.geoserver.org/stable/en/user/community/wps-download/rawDownload.html). But then the resulting file is somewhat different from original tiff uploaded to GeoServer. It's still an ongoing issue, as such we need to make a custom API endpoint to download the original GeoTiff

The cons is that we need to configure the backend and layer configuration a bit, but the benefit is that we can customize what the resulting file will contain, for example in CSV output we can add more columns

image

I'm using Mongolia / Pasture Anomaly to demonstrate the functionally, it will be removed on merge


reed1:
feature for IDN branch

PR #181, opened 2021-05-06
Frontend dockerized for development

erwagasore:
I saw that both `alerting` and `api-flask` are dockerized and being accustomed to using Docker in development. I opted on dockerizing the frontend for development and maybe in the future we even set up for `production`.

The directory arrangement for this `mono-repo` can be re-arranged for better separation, check the image below.

directory_structure


erwagasore:
@ericboucher please review

web/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM node:12-alpine
+
+WORKDIR /home/app
+COPY . .
+
+RUN addgroup -S appgroup
+RUN adduser -S -D -h /home/app app appgroup
+RUN chown -R app:appgroup /home/app
```
ericboucher, line 8: Do we need these?

web/.dockerignore
```diff
@@ -0,0 +1,4 @@
+**/node_modules
+
+alerting
```
ericboucher, line 3: these are not needed anymore with this folder structure

web/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: '3.7'
```
ericboucher, line 1: How about having just one docker-compose file in the main folder?

web/Dockerfile
ericboucher: Open quetsion, do we even need a Dockerfile?

How about having a named volumes for node modules and running yarn on launch?

Your approach works but it means that we will have to re-build the docker whenever we change a library

ericboucher (commented):
It's a good start! I left a few comments inline, plus:

Don't move the README for now or split it into a main README and a more specific web README

web/.dockerignore
erwagasore: True

web/docker-compose.yml
erwagasore: Do you mean for all services "web", "alerting", and "api-flask"?

web/Dockerfile
erwagasore: For development, we don't need them.

web/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+      context: .
+      dockerfile: Dockerfile
+    image: prism_web:latest
+    container_name: prism_web
+    env_file:
+        - .env
+    volumes:
+      - '.:/home/app'
+    ports:
+      - ${PORT}:${PORT}
```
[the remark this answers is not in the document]
erwagasore: Got it.

ericboucher:
Can I propose this structure:
- website
- api
- alerting

wadhwamatic:
As discussed today @erwagasore , let's separate the directory restructure from dockerization.

ericboucher:
Let's close this PR, it has diverged too much anyways and it will be easier to re-open a new one.

PR #196, opened 2021-05-19
Fix missing wfs_config and use .tif extension for rasters

ericboucher:
Better error handling around geotiffs + fix a small issue with missing wcs_config arg in the demo endpoint


api-flask/app/caching.py
```diff
@@ -24,16 +28,33 @@ def cache_file(url, prefix, extension='cache'):
...
     if os.path.isfile(cache_filepath):
         logger.info('Returning cached file for {}.'.format(url))
-        return cache_filepath
+
+        # if the file is a geotiff, confirm that we can open it.
+        if extension == 'tif':
+            try:
+                rasterio.open(cache_filepath)
+                return cache_filepath
+            except Exception:
```
JorgeMartinezG, line 37: We would need to specify the type of exception we are silencing. For this case, we could wrap all possible errors from rasterio with **rasterio.errors.RasterioError** or *rasterio.errors.RasterioIOError*

api-flask/app/caching.py
```diff
@@ -24,16 +28,33 @@ def cache_file(url, prefix, extension='cache'):
...
+                pass
+        else:
+            return cache_filepath
+
     # If the file does not exist, download and return path.
-    else:
-        r = requests.get(url, verify=False)
+    response = requests.get(url, verify=False)
+
+    try:
```
JorgeMartinezG, line 45: For this case, we could validate the status code we are receiving and then specify the status code we are getting from the remote server. Something like

```
if response.status_code != 200:
  raise InternalServer.. 
```

api-flask/app/zonal_stats.py
```diff
@@ -161,13 +163,17 @@ def calculate_stats(
...
-    )
+    try:
+        stats = zonal_stats(
+            stats_input,
+            geotiff,
+            stats=stats,
+            prefix=prefix,
+            geojson_out=geojson_out
+        )
+    except Exception as e:
```
JorgeMartinezG, line 174: Similar to the previous comment. We could wrap this exception into a RasterioError one.

JorgeMartinezG (changes requested):
Added a few comments related on how the errors are handled in this PR.

api-flask/app/caching.py
ericboucher: The error status will be specified in the error already. I switched to using the HTTPError exception

JorgeMartinezG (approved):

PR #200, opened 2021-05-25
Use below/above language in the alert form. Add default boundaries file

ericboucher:
- Re-order above / below thresholds
- transform missing min/max to undefined before sending the alert to the API
- Add "true" default to alert `active` field

### DIFF
--- a/src/context/analysisResultStateSlice.ts
+++ b/src/context/analysisResultStateSlice.ts
@@ -61,6 +61,10 @@ function getAdminBoundariesURL() {
   if (adminBoundariesPath.startsWith('http')) {
     return adminBoundariesPath;
   }
+  // do not send a local path to the API, use a fixed boundary file instead.
+  if (window.location.origin.includes('localhost')) {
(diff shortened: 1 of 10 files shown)


Ry-DS (approved):
Looks good :+1:

src/context/analysisResultStateSlice.ts
```diff
@@ -61,6 +61,10 @@ function getAdminBoundariesURL() {
   if (adminBoundariesPath.startsWith('http')) {
     return adminBoundariesPath;
   }
+  // do not send a local path to the API, use a fixed boundary file instead.
+  if (window.location.origin.includes('localhost')) {
```
Ry-DS, line 65: For this, would it make more sense to check the environment and see if its in development mode? And then refer to this url as a constant? Trying to think of ways to make it consistent to how we load development mode in general

Ry-DS: maybe in src/config/index.ts somewhere

ericboucher: This breaks if the country changes but the window includes localhost is a safe test that we are in development mode, no?

PR #202, opened 2021-05-27
Apply WCS GetCoverage request with version 2.0.0

JorgeMartinezG:
This PR updates wcs request for analysis to version 2.0.0. With this version it will not be necessary to specify returning image WIDTH and HEIGHT.

**Changes**:
- Removed WCSTileUrls function. It is not used in the project.
- Modified getWCSUrl to version 2.0.0. Time parameter is now considered a subset request parameter.

### DIFF
--- a/src/components/MapView/Layers/raster-utils.ts
+++ b/src/components/MapView/Layers/raster-utils.ts
@@ -60,16 +60,6 @@ export type GeoTiffImage = {
   }) => Promise<Rasters>;
 };
 
-function numberOfTiles(
-  min: number,
-  max: number,
-  resolution: number,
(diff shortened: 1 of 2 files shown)


dubwalla:
Thanks @JorgeMartinezG . One quick input from me (and I'll look in more detail tomorrow) - should we allow for backwards compatibility of WCS versions? While we know our ODC implementation and GeoServer are using WCS 2.x, I'm not sure if we can assume all systems PRISM will retrieve layers from will have the same

JorgeMartinezG:
Closing in favor of https://github.com/WFP-VAM/prism-frontend/pull/206

Issue #201, opened 2021-05-26
Update WMS version to 1.3.0


valpesendorfer:
We might have had a discussion on this before - but why is PRISM using WMS `v1.1.1`?

I've had a talk with the devs @ odc-ows, and at least on paper the datacube-ows stack only supports `v1.3.0`:

https://github.com/opendatacube/datacube-ows/blob/master/datacube_ows/protocol_versions.py#L59-L61

PR #204, opened 2021-05-27
update WMS version to 1.3.0

dubwalla:
The WMS version we were using is very out of date. 1.3.0 has been active since 2006. Tested locally for ODC OWS layers and GeoServer layers as well without issue.

This will close https://github.com/WFP-VAM/prism-frontend/issues/201


dubwalla:
I inadvertently added Myanmar layer fixes here too. Those address an update on the GeoServer side for how we are managing tropical storm location filtering

ericboucher (approved):

Issue #124, opened 2021-01-13
Download data layer


wadhwamatic:
Allow a user to download data as a CSV for tabular / vector data and as a GeoTIFF for raster data. The same UI element should allow a user to download the selected data layer regardless of the type of data. 

Suggestion: The UI should include a download link / button within the context of the current legend area. At a later date, we may revisit this and create a more extensive UI for interacting with layers.

erwagasore:
I have realized that there are many places where the user can download either data or images. We can try to consolidate those user's actions to one central location. I had a discussion with @dubwalla suggesting having a layers sidebar. Below is a mockup of how that would look like.

![Prism - Layers hidden](https://user-images.githubusercontent.com/1290048/117296404-4d868400-ae75-11eb-9e1f-bdcc8ee0ecd9.png)

![Prism - Layers Mockup](https://user-images.githubusercontent.com/1290048/117296456-5e36fa00-ae75-11eb-8016-1305c98a0fd9.png)

I took inspiration from some implementations of ![TerriaJS](https://nsw.digitaltwin.terria.io/)
Please tell me how you see it ;)

erwagasore:
This suggestion also took into consideration of some already created branches `idn_dev` and another one for `layer opacity`

wadhwamatic:
@ericboucher - temporarily assigning this to you (Ovio) but feel free to close this and refer to a new issue or the related ones ([618](https://github.com/WFP-VAM/prism-app/issues/618), [209](https://github.com/WFP-VAM/prism-app/issues/209))

Issue #191, opened 2021-05-17
Set geometry type for vector layers


JorgeMartinezG:
In order to validate that a layer is able to perform population exposure analysis, a geometry type  needs to be included within its configuration options. Also, there needs to be a check in order to remove vector layers for common analysis.

wadhwamatic:
I believe you're working on something similar here @DanielJDufour ?

wadhwamatic:
closed via https://github.com/WFP-VAM/prism-frontend/pull/192

PR #208, opened 2021-06-03
🎉 Add ability to download layer raw data: #124

erwagasore:
On every active layer, the end-user can now download raw data as `GeoTiff` or `GeoJSON`.
We could push to download `csv` but I thought that dedicating this button to only `raw` data is best.

Screen Shot 2021-06-03 at 5 20 11 PM


erwagasore:
@ericboucher please review

src/components/MapView/Legends/index.tsx
```diff
@@ -16,14 +16,31 @@ import {
...
+  PointDataLayerProps,
+} from '../../../config/types';
 import { formatWMSLegendUrl } from '../../../utils/server-utils';
+import { getWCSLayerUrl } from '../../../context/layers/wms';
 import {
   analysisResultSelector,
   isAnalysisLayerActiveSelector,
 } from '../../../context/analysisResultStateSlice';
+import { Extent } from '../Layers/raster-utils';
+// import { layer } from '@fortawesome/fontawesome-svg-core';
```
ericboucher, line 40: 👣

wadhwamatic:
Good start Eugene! A couple of things (first one we talked about already):
1. It would make sense to have 'nso' layers downloadable as a CSV or JSON. I created an issue for that: https://github.com/WFP-VAM/prism-frontend/issues/209 
2. Currently, the nso download uses a .txt extension instead of .json. Please update that
3. 'wms' layers can have different data types (rasters, points, polygons) which needs to be handled in the download. Jorge started to work on this (https://github.com/WFP-VAM/prism-frontend/issues/191). At present, when I test a non-raster WMS download, it get forwarded to the source page (GeoServer in this case). You can see this if you test with Myanmar and try to download any of the tropical storm layers

src/components/MapView/Legends/index.tsx
```diff
@@ -154,6 +174,104 @@ function LegendItem({
...
+    const extent = bbox(boundaryLayerData?.data) as Extent;
+
+    switch (legendLayer.type) {
+      case 'wms': {
+        const dataUrl = getWCSLayerUrl({
+          layer: legendLayer,
+          extent,
+          date: selectedDate,
+        });
+        downloadToFile(
```
Ry-DS, line 221: does download related code belong in `Legends`? Moving this to a new util file would be nice for organization and if we want to support downloading through other sources.

Ry-DS (commented):
Looks good overall :+1:

src/components/MapView/Legends/index.tsx
erwagasore: This makes perfect sense. Since it can be used somewhere else

ericboucher: I agree, let's move these util functions somewhere else.

wadhwamatic:
Good progress, but I found a couple of issues when testing. 

1) For the tropical storm layers, it seems we're not adding a bounding box and date to the requested layer so when you download, you get all wind buffers, nodes and tracks. It should instead correspond to what is being visualized. Here's a screenshot:
Screen Shot 2021-06-10 at 22 04 19

2) For flood data, it seems we're not getting the right resolution in the download. Check with @JorgeMartinezG on how this was resolved with population raster data. Here are screenshots showing the view in PRISM compared to the downloaded geotiff

Screen Shot 2021-06-10 at 22 05 51
Screen Shot 2021-06-10 at 22 05 44

erwagasore:
Hi @wadhwamatic the last commit solves both issues you raise.

Below are resulting data plotted in QGIS:

Screen Shot 2021-06-16 at 1 42 39 PM
Screen Shot 2021-06-16 at 1 53 07 PM
Screen Shot 2021-06-16 at 1 53 13 PM

wadhwamatic:
Nice work, but still facing some issues unfortunately. 
1. The flood raster data loses information when downloading. See screenshot from a flood layer on 1-Jan-2021 in PRISM compared to the downloaded raster in QGIS

Screen Shot 2021-06-16 at 12 42 30
Screen Shot 2021-06-16 at 12 42 19

2. When I try to download the most recent tropical storm layers (26-May-2021), the returned .json files have no data (i.e. empty features). I expect this is because the storm is not within country's boundaries. I'm wondering if the download query would need to consider the events visualized rather than those within a country boundary.  

Here's the json from the tropical storm node layer:
{"type":"FeatureCollection","features":[],"totalFeatures":0,"numberMatched":0,"numberReturned":0,"timeStamp":"2021-06-16T19:25:44.575Z","crs":null}

Screenshot of the same storm:
Screen Shot 2021-06-16 at 13 46 09

erwagasore:
@wadhwamatic for vector it is working perfectly and I have increased the `resolution` and `maxPixel` values to enable download to have the higher image resolution.

wadhwamatic:
> @wadhwamatic for vector it is working perfectly and I have increased the `resolution` and `maxPixel` values to enable download to have the higher image resolution.

Thanks Jorge. This looks good now from a functionality perspective. Requesting @JorgeMartinezG and/or @ericboucher to do a code review

JorgeMartinezG:
Looks Good to me. One thing to note is that the downloaded raster might not have original resolution, since wcs request is using v1.0.0. Talking with @erwagasore, we made tests with 2.0 version to ODC, receiving a mix of 502 and 504 status codes.

For instance: https://odc.ovio.org/?service=WCS&request=GetCoverage&version=2.0.0&coverageId=hfs1_sfw_mask&subset=Long(92.17274709800029,101.17001505562291)&subset=Lat(9.67125210200021,28.545538862)&subset=time(%222021-05-03T10:00:00.000Z%22)

cc @wadhwamatic 

Also, I have the geojson/csv check #209 in another branch. Once this PR is merged I'll make a new one on top of master

wadhwamatic:
Need to address the resolution issue. https://github.com/WFP-VAM/prism-frontend/issues/216

src/components/MapView/Legends/index.tsx
```diff
@@ -231,58 +236,59 @@ function LegendItem({
...
+                  content: JSON.stringify(data),
+                  isUrl: false,
+                },
+                legendLayer.title,
+                'application/json',
+              );
+            })
+            .catch(err => {
+              throw new Error(err);
+            });
```
Ry-DS, line 272:
```suggestion
            .catch(console.error);
```
I'm not too sure if this is needed since the error will be thrown regardless if there's no catch case. If you want you can pass it into `console.error` which will be neater than throwing the error (unhandled errors in promises are discouraged: https://developers.redhat.com/blog/2016/11/09/node-7-and-promise-rejections-please-handle-them)

src/components/MapView/Legends/index.tsx
```diff
@@ -398,6 +398,30 @@ function LegendItem({
...
+          )}
+
+          {exposure && analysisResult && (
+            <AnalysisButton
+              variant="contained"
+              color="secondary"
+              size="small"
+              onClick={() => dispatch(clearAnalysisResult())}
+            >
+              clear analysis
```
Ry-DS:
```suggestion
              Clear Analysis
```
Nice to case these well even if they are in all caps just in case we revert back to Normal case ;)

Ry-DS (commented):
Nice! Few suggestions to help with quality :+1:

src/components/MapView/Layers/raster-utils.ts
```diff
@@ -155,6 +156,7 @@ export function WCSRequestUrl(
   layer: WMSLayerProps,
   date: string | undefined,
   extent: Extent,
+  resolution: number | undefined,
```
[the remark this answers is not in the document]
erwagasore: Solved. Check below
```typescript
export function WCSRequestUrl(
  layer: WMSLayerProps,
  extent: Extent,
  date?: string,
  resolution?: number,
  maxPixels = 5096,
) {...}
```

src/components/MapView/Layers/raster-utils.ts
```diff
@@ -194,6 +200,46 @@ export function WCSRequestUrl(
   );
 }
 
+export function getWFSUrl(
+  baseUrl: string,
+  layerName: string,
```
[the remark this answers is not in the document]
erwagasore: No. We exect a string here. The name `layerName` is wrong I have replaced it with proper name `serverLayerName` which is always of type `string`. The updated implementation is below

```typescript
export function getWFSUrl(
  baseUrl: string,
  serverLayerName: string,
  override: { [key: string]: string } = {},
) {...}
```

src/components/MapView/Layers/raster-utils.ts
```diff
@@ -194,6 +200,46 @@ export function WCSRequestUrl(
...
+    outputFormat: 'application/json',
+    ...override,
+  };
+
+  return formatUrl(`${baseUrl}/prism/ows`, params);
+}
+
+export function WFSRequestUrl(
+  layer: WMSLayerProps,
+  date: string | undefined,
```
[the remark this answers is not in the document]
erwagasore: Solved. New code below

```typescript
export function WFSRequestUrl(
  layer: WMSLayerProps,
  extent: Extent,
  override: { [key: string]: string } = {},
  date?: string,
) {
```

src/components/MapView/Legends/index.tsx
erwagasore: You are right. Thank you for the link was not aware. Fix below as you exactly suggested.

```typescript
...catch(console.error);
```

src/components/MapView/Legends/index.tsx
erwagasore: Good point. Done.

ericboucher:
@erwagasore can you fix the conflicts? It would be nice to merge this soon.

As a quick note: as the PR owner, it is your responsibility to push it to the finish line, or decide to scrap it if is not relevant anymore. PRs should be left stale like this for 4 months ;)

wadhwamatic:
> @erwagasore can you fix the conflicts? It would be nice to merge this soon.
> 
> As a quick note: as the PR owner, it is your responsibility to push it to the finish line, or decide to scrap it if is not relevant anymore. PRs should be left stale like this for 4 months ;)

@erwagasore  - suggesting we use what has been completed to download vector datasets but remove the functionality for rasters. The raster approach needs to be revisited as WCS is not a good solution. See Valentin's comments here: https://github.com/WFP-VAM/prism-frontend/issues/216#issuecomment-876226161

PR #220, opened 2021-06-24
scale statistic value given parameters

### LINKED ISSUE #219: Exposure analysis is not using the scale factor set in layer configurations
Exposure analysis is not using the scale factor set in layer configurations. Please check if this affects stats API broadly. We'll need to validate for the Mongolia deployment before making any backend changes.

It also seems the layer config for Myanmar and Mozambique are specifying the scale factor in the wrong place (wind buffer instead of population raster).

JorgeMartinezG:
Closes #219 

This PR enables the value update given scale and offset parameters for each feature returned from the API. These parameters are set within the wcsConfig variable within the layers configuration file.

### DIFF
--- a/src/components/MapView/Legends/index.tsx
+++ b/src/components/MapView/Legends/index.tsx
@@ -228,7 +228,11 @@ function LegendItem({
                 legend.map(({ value, color }: any) => (
                   <ColorIndicator
                     key={value}
-                    value={value.toLocaleString('en-US')}
+                    value={
+                      typeof value === 'number'
+                        ? Math.round(value).toLocaleString('en-US')
+                        : value
+                    }
                     color={color as string}
                     opacity={opacity as number}
                   />
--- a/src/components/MapView/Legends/index.tsx
+++ b/src/components/MapView/Legends/index.tsx
@@ -228,7 +228,7 @@ function LegendItem({
                 legend.map(({ value, color }: any) => (
                   <ColorIndicator
                     key={value}
-                    value={value as string}
+                    value={value.toLocaleString('en-US')}
                     color={color as string}
                     opacity={opacity as number}
                   />
--- a/src/components/MapView/Legends/index.tsx
+++ b/src/components/MapView/Legends/index.tsx
@@ -24,6 +24,7 @@ import {
   LayerType,
   GeometryType,
   ExposedPopulationDefinition,
+  LegendDefinitionItem,
 } from '../../../config/types';
 import { formatWMSLegendUrl } from '../../../utils/server-utils';
 import {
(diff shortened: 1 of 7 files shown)


wadhwamatic (approved):
Tested from my side, looks good. Thanks Jorge

src/components/MapView/Legends/index.tsx
```diff
@@ -228,7 +228,11 @@ function LegendItem({
                 legend.map(({ value, color }: any) => (
```
Ry-DS, line 228: know this isn't in scope for this PR, but would love if we could investigate why there's an `any` here. Possibly a comment or fix would be good :+1:

src/context/analysisResultStateSlice.ts
```diff
@@ -217,11 +218,23 @@ export const requestAndStoreExposedPopulation = createAsyncThunk<
...
+  const { scale, offset } = populationLayer.wcsConfig ?? {
+    scale: undefined,
+    offset: undefined,
+  };
+
+  const features =
+    !scale && !offset
+      ? apiFeatures
+      : apiFeatures.map(f =>
+          scaleFeatureStat(f, scale || 1, offset || 0, statistic),
```
Ry-DS, line 235: we could technically just give `apiFeatures.map(f =>
          scaleFeatureStat(f, scale || 1, offset || 0, statistic)` here without a ternary right?
Since the `|| 1` and `|| 0` would anyway negate the effect of undefined scale or offset. The only improvement I see over this is a small performance boost. Let me know what you think :+1:

src/components/MapView/Legends/index.tsx
```diff
@@ -228,7 +228,11 @@ function LegendItem({
...
                   <ColorIndicator
                     key={value}
-                    value={value as string}
+                    value={
+                      typeof value === 'number'
+                        ? Math.round(value).toLocaleString('en-US')
```
Ry-DS, line 233: is `number.toString()` ok here? Or is `toLocaleString` better for this case?

Ry-DS (commented):
Nice! Generally G2G, just a few small things we could look into:

src/components/MapView/Legends/index.tsx
JorgeMartinezG: toLocateString includes the thousand separator. For instance 2000 as number will give you '2,000'

src/components/MapView/Legends/index.tsx
JorgeMartinezG: Included in 56e98bb6b12cf499da432ececb7b027b616a31d9

src/context/analysisResultStateSlice.ts
JorgeMartinezG: Yes. Basically is to avoid the loop if scale and offset are not defined

PR #232, opened 2021-07-01
🎉 Feature/pop analysis data table

erwagasore:
- Updated the toggle switch to activate `showing the table` instead of the map
- Massaged pop-exposure data from `object` to arrays for table consumption
- Render Pop Exposure Analysis in a drawer and showed the data in table

MicrosoftTeams-image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ yarn-error.log*
 # Editors
 .vscode/
 .idea/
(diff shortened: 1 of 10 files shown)


erwagasore:
CI Tests were failing after merging master into the branch. I decided to re-base with intention to retrace issue and solve it.

Issue #224, opened 2021-06-28
Display population exposure analysis outputs in a table view which can be downloaded


wadhwamatic:
With the completion of PR https://github.com/WFP-VAM/prism-frontend/pull/214 we now get a response from the API where the statistics (max, mean, median, min, std, and sum), a categorical label, and the admin name are wrapped within a properties field. However each element is a custom polygon based on the intersection of a wind buffer and admin area. We need to take these elements and aggregate them (the sum specifically) to the admin level which is then displayed as a table.

The table should look like this:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/amit/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/amit/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">

</head>

<body link="#0563C1" vlink="#954F72">

Township | 60 km/h exposure | 90 km/h exposure | 120 km/h exposure | Total
-- | -- | -- | -- | --
Ramree | 1714 | 691 | 250 | 2655
Toungup | 2677 | 951 | 440 | 4068
Kyaukpyu | 2084 | 794 | 396 | 3274
Labutta | 2494 | 572 | 371 | 3437
Ngapudaw | 1790 | 537 | 332 | 2659
Bogale | 2208 | 911 | 302 | 3421
Thandwe | 1310 | 514 | 37 | 1861
Kungyangon | 4194 | 820 | 31 | 5045

</body>

</html>

As @JorgeMartinezG said on Slack: 

> The geometry contains the polygon of the intersection between the wind buffer and admin boundary. There is a possibility of having more than one adm repeated, since two or more wind buffers might intersect. One quick way to generate the data table is to group by admin boundary to get all possible wind buffers values and population affected. And then transform it into some sort of array of arrays which is the type that the DataTable accepts

An example response can be found here: https://gist.github.com/wadhwamatic/34a7953ee7a751a982ef20c8319d9a61

PR #233, opened 2021-07-01
🎉 Feature/pop analysis data table

erwagasore:
- Updated the toggle switch to activate showing the table instead of the map
- Massaged pop-exposure data from object to arrays for table consumption
- Render Pop Exposure Analysis in a drawer and showed the data in table

MicrosoftTeams-image


wadhwamatic:
Good progress. A couple of things:
1. Let's round the data to the nearest whole number and add a thousands separator (i.e. add a comma)
2. When there's no data returned (i.e. the storm doesn't affect any population), can we have text that says: "No population exposed" in the table area? Right now it's just a big empty pane (screenshot below)

Screen Shot 2021-07-01 at 15 38 38

wadhwamatic:
> * Let's round the data to the nearest whole number and add a thousands separator (i.e. add a comma)

@JorgeMartinezG - can you point to where you addressed this already?

JorgeMartinezG (commented):
In my opinion a better option would be to reuse the DataDrawer component which handles its own state and style.

The workflow would be to update the tableState within the MapView or AnalysisLayer component, doing a review of the type of AnalysisResult. Something similar like:
```
if (analysisResult instanceof ExposedPopulationResult) {
   //
   const tableData: TableData = someFunctionThatBuildsTableData (analsysisResult.featureCollection);
   dispatch (addTableData (tableData));
}
```
Also, please check that the name fields within the DataTable component are hardcoded for Tropical cyclone data only. One option would be to pass the necessary parameters to group and extract the values. Earthquake and shakemap layers are planned to be included and have different fields.

I want to know your opinion about this @ericboucher

wadhwamatic:
On a couple of attempts, I got an error if I load an ODC hazard layer (for example, 10-day rainfall estimate) before loading anything else, saying that no date was found. Then, when I tried to load the tropical storm layers, I got this error. However, after this happened twice, I couldn't recreate it. Have you seen this error? Was anything relating to dates changed in this branch?

Screen Shot 2021-07-07 at 08 33 59

erwagasore:
> On a couple of attempts, I got an error if I load an ODC hazard layer (for example, 10-day rainfall estimate) before loading anything else, saying that no date was found. Then, when I tried to load the tropical storm layers, I got this error. However, after this happened twice, I couldn't recreate it. Have you seen this error? Was anything relating to dates changed in this branch?
> 
> Screen Shot 2021-07-07 at 08 33 59

I tried to replicate this issue but couldn't. To conduct `Population Exposure Analysis` we benefit from `selectedDate` which is always present as `global state`.

erwagasore:
@JorgeMartinezG and @ericboucher new changes are ready to be reviewed

- Improved the table component to be more dynamic and benefit from already available table rendering functionality fixing issue #224 
- Disable ability to show table if there are no `pop-exposure analysis` data instead show a message
- Hide `pop-exposure legend` no features are available.
- Allow user to download `pop-exposure` data as `CSV` which fixes the issue #235 

Below are different state previews:

Screen Shot 2021-07-14 at 5 50 29 PM

Screen Shot 2021-07-14 at 6 04 38 PM

Screen Shot 2021-07-14 at 6 05 53 PM

src/components/DataDrawer/index.tsx
```diff
@@ -9,18 +9,27 @@ import {
...
+    dispatch(hideDataTableDrawer());
   };
 
   return (
-    <Drawer anchor="left" open={tableIsShowing} onClose={handleClose}>
+    <Drawer anchor="left" open={isShowing} onClose={handleClose}>
       <div className={classes.drawerContent}>
-        <DataTable maxResults={1000} />
+        {tableIsShowing ||
+          (analysisTableIsShowing && <DataTable maxResults={1000} />)}
```
Ry-DS, line 32: This is somewhat confusing. Think we can combine this to one conditional expression?

src/components/MapView/Legends/exposedPopulationAnalysis.tsx
```diff
@@ -80,20 +83,49 @@ const ExposedPopulationAnalysis = ({
...
-      <AnalysisFormControlLabel
-        control={
-          <Switch
-            color="primary"
-            checked={isMapLayerActive}
-            onChange={e => dispatch(setIsMapLayerActive(e.target.checked))}
+  const ResultSwitches = () => {
+    const data = useSelector(analysisResultSelector);
+    const features = data?.featureCollection.features;
+    const hasData = features?.length === 0 || false;
```
Ry-DS, line 89: is the || false needed?

src/components/MapView/Legends/exposedPopulationAnalysis.tsx
```diff
@@ -80,20 +83,49 @@ const ExposedPopulationAnalysis = ({
...
+
+    const handleTableViewChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+      dispatch(setIsDataTableDrawerActive(e.target.checked));
+      dispatch(
+        setCurrentDataDefinition({
+          id: TableDefinitions.animal_deaths.id,
+          title: data?.getTitle() as string,
+          table: '',
+          legendText: data?.legendText as string,
```
Ry-DS, line 98: it isn't great to cast like this and ignore TS's concerns that these could be undefined. Maybe we can do an if check to make sure they are indeed not undefined? Otherwise, if we're 100% certain these can't be undefined we can leave as is

src/components/MapView/utils.ts
```diff
@@ -1,7 +1,10 @@
+import _ from 'lodash';
```
Ry-DS, line 1: we should import the exact functions we use with `{uniq, ...} from ...` since it's easier to "tree-shake" in the future (reduce bundle size)

src/components/MapView/utils.ts
```diff
@@ -46,3 +49,83 @@ export const getFeatureInfoParams = (
...
+  });
+  const rowData = _.mapValues(_.groupBy(featureProperties, groupBy), x =>
+    _.chain(x)
+      .keyBy(key)
+      .mapValues(statistic)
+      .mapValues(z => _.parseInt(z))
+      .value(),
+  );
+
+  const groupedRowData = _.map(rowData, (x, i: number) => {
```
Ry-DS, line 78: try avoid using lodash methods if a vanilla approach would work just as well (reduces complexity)
Possibly every `map` used in this file could be replaced with the array.map

src/components/MapView/utils.ts
```diff
@@ -46,3 +49,83 @@ export const getFeatureInfoParams = (
...
 };
+
+export const convertToTableData = (
+  result: ExposedPopulationResult,
+  groupBy: string,
+) => {
+  const { key, statistic } = result;
+  const { features } = result.featureCollection;
+  const fields = _.uniq(
+    _.map(features, f => f.properties && f.properties[key]),
```
Ry-DS, line 60: optional chaining?

Ry-DS (commented):
Nice! Left a few comments let me know what you think :+1:

src/components/DataDrawer/index.tsx
erwagasore: Did just that. Check the new implementation below

```javascript
const isShowing = tableIsShowing || analysisTableIsShowing;
...
{isShowing && <DataTable maxResults={1000} />}
```

src/components/MapView/Legends/exposedPopulationAnalysis.tsx
erwagasore: Solved. Check the implementation below

```javascript
const hasData = features?.length === 0;
```

src/components/MapView/Legends/exposedPopulationAnalysis.tsx
erwagasore: You are correct. There is potential for it to be `undefinded`. Solved with implementation below

```javascript
{
          id: TableDefinitions.animal_deaths.id,
          title: data ? data.getTitle() : '',
          table: '',
          legendText: data ? data.legendText : '',
}
```

src/components/MapView/utils.ts
erwagasore: Yeah. I have solved it. Check below

```javascript
import { values, zipObject, assign, difference, mapValues, groupBy, keysIn, keyBy, uniq } from 'lodash';
```

src/components/MapView/utils.ts
erwagasore: Yeah. New to Typescript, I am loving it :) Check correction below

```typescript
const { key, statistic, featureCollection: { features } } = result;
```

src/components/MapView/utils.ts
erwagasore: Good point. I have updated the entire function to use vanilla functions as much as possible. Reduced complexity as well.

JorgeMartinezG (approved):

src/components/MapView/utils.ts
```diff
@@ -52,64 +62,53 @@ export const getFeatureInfoParams = (
...
   return { columns, rows };
 };
 
 export const exportDataTableToCSV = (data: TableData) => {
   const { rows } = data;
-  return _.join(
-    _.map(rows, r => _.map(_.values(r), x => x)),
-    '\n',
-  );
+  return rows.map(r => values(r).map(x => x)).join('\n');
```
Ry-DS, line 111: I don't understand `map(x => x)` here, wouldn't it just return the same array? Maybe we can remove it, but if it does do something feel free to add a comment :)

Ry-DS (approved):
Thank you for taking the time to comment the exact code you changed! Really helps speed up the review process :)
Overall this looks g2g! Just a few small comments but not too important :)

src/components/MapView/utils.ts
erwagasore: You are right. This is doesn't do anything more. Already removed.

src/components/MapView/Legends/exposedPopulationAnalysis.tsx
```diff
@@ -86,16 +86,16 @@ const ExposedPopulationAnalysis = ({
...
-    const hasData = features?.length === 0 || false;
+    const hasData = features?.length === 0;
 
     const handleTableViewChange = (e: React.ChangeEvent<HTMLInputElement>) => {
       dispatch(setIsDataTableDrawerActive(e.target.checked));
       dispatch(
         setCurrentDataDefinition({
           id: TableDefinitions.animal_deaths.id,
-          title: data?.getTitle() as string,
+          title: data ? data.getTitle() : '',
```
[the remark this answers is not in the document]
erwagasore: Works perfectly. Thanks

src/components/MapView/Legends/exposedPopulationAnalysis.tsx
```diff
@@ -80,20 +83,49 @@ const ExposedPopulationAnalysis = ({
...
+  const ResultSwitches = () => {
+    const data = useSelector(analysisResultSelector);
+    const features = data?.featureCollection.features;
+    const hasData = features?.length === 0;
+
+    const handleTableViewChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+      dispatch(setIsDataTableDrawerActive(e.target.checked));
+      dispatch(
+        setCurrentDataDefinition({
+          id: TableDefinitions.animal_deaths.id,
```
ericboucher, line 95: @erwagasore why is referencing `animal_deaths` which is a country-specific dataset? Not sure what we are trying to achieve here?

erwagasore: Looking into it...

erwagasore: @ericboucher just created a fix https://github.com/WFP-VAM/prism-frontend/pull/373

Issue #238, opened 2021-07-26
Failure rendering exposed population table for earthquake shakemaps


JorgeMartinezG:
Since commit https://github.com/WFP-VAM/prism-frontend/commit/346569624ea2c190e5bdc11cd386aace7adc7dd2, PRISM generates a table which displays exposed population per admin boundary specified within the **groupBy** key and wind buffer polygon label. 

This approach works perfectly for tropical Storms layer. However, it fails to render correctly for different countries and layers. The reason is that the **groupBy** key is hardcoded here: https://github.com/WFP-VAM/prism-frontend/commit/346569624ea2c190e5bdc11cd386aace7adc7dd2#diff-7ae5c288cb15a212f72f749c04fc8d4ed41fe529e99af7287a89af076502b178R232 and it is not found in all countries geojson files.

Furthermore, not all layers have the **label** column. This parameter, together with the **groupBy** one, need to be passed to the **ExposedPopulationResult** object for correct grouping and filtering.

Also, there is a possibility that multiple geometries might contain same label and admin boundary. So, PRISM should be able to sum the grouped by values per boundary and label.

PR #239, opened 2021-07-28
Tajikistan earthquake shakemaps

JorgeMartinezG:
This PR includes shakemap layer served from geoserver as WMS/WFS request:

**Additional features:**
- Included getFeatureinfo props to get mmi value.

image

**Note:** Exposed population analysis is currently disabled. There is a bug within the table rendering step. Documented here: https://github.com/WFP-VAM/prism-frontend/issues/238


JorgeMartinezG:
This branch has been updated and includes a fix for legend generation. Also, validated the exposed population analysis. cc @erwagasore @wadhwamatic 

image

erwagasore (approved):
Seems very straight forward. Ship it.

src/utils/analysis-utils.ts
```diff
@@ -524,15 +524,17 @@ export function createLegendFromFeatureArray(
     (a, b) => a - b, // This function is required since JS assumes unicode values.
   );
 
-  const statsLen = stats.length;
+  const uniqueStats = [...new Set(stats)]; // Remove repeated elements.
```
ericboucher, line 527: Can you explain a bit more when this is used and why it is important to remove the repeated elements?

*Technically speaking, the values below are not quintiles anymore if you pre-filter to remove duplicate values.

ericboucher: Yes I understand, however:
- these are not quintiles anymore by definition (not a huge deal - my request was mostly to add comments in the code)

- this problem is a bit more complex than it seems, and this solution might not work exactly as you expect. I recommend adding some tests for this function with a few edge cases. Eg. what happens if there are less than 5 unique values? Then we would have 1,734,633 multiple times as well. -> Using unique values also risks adding its own edge cases.
- We could also envision using an equal value split between the min/max for example.

wadhwamatic (approved):
Nice work on this Jorge. I think the issue raised by Eric about equal interval vs quintile deserves a bit more thinking. For now, let's use the solution you've created, but I'll create another issue on this.

wadhwamatic:
> Nice work on this Jorge. I think the issue raised by Eric about equal interval vs quintile deserves a bit more thinking. For now, let's use the solution you've created, but I'll create another issue on this.

See issue https://github.com/WFP-VAM/prism-frontend/issues/246

PR #240, opened 2021-08-02
🐛 Fix Exposure Population Analysis Minor issues

erwagasore:
Here is a list of minor issues and improvement added with regard to `exposure population analysis` table rendering
- Fixed `table` showing `index` number instead of `Group By` value
- Improved `table rendering` to be more dynamic and remove had-coded `Group By` and `label`
- Fixed `map` rendering to always show the `boundaries` layer before all other layers

### DIFF
--- a/src/components/MapView/utils.ts
+++ b/src/components/MapView/utils.ts
@@ -4,7 +4,7 @@ import {
   assign,
   difference,
   mapValues,
-  groupBy,
+  groupBy as _groupBy,
   keysIn,
   keyBy,
   uniq,
@@ -60,12 +60,10 @@ export const getFeatureInfoParams = (
   return params;
 };
 
-export const convertToTableData = (
-  result: ExposedPopulationResult,
-  groupedBy: string,
-) => {
+export const convertToTableData = (result: ExposedPopulationResult) => {
   const {
     key,
+    groupBy,
     statistic,
     featureCollection: { features },
   } = result;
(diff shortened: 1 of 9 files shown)


src/components/MapView/utils.ts
```diff
@@ -97,10 +100,11 @@ export const convertToTableData = (
   });
 
   const headlessRows = groupedRowDataWithAllLabels.map(row => {
-    const total = fields.reduce((acc, o) => row[acc] + row[o]);
+    const total = fields.map(f => row[f]).reduce((a, b) => a + b);
```
Ry-DS, line 103: does something like `sumBy(fields, key=>row[key]);` from lodash work? Didn't look into this too far, so all good if it doesn't work :+1:

Ry-DS (approved):
This is great!
Quick question, do you think this can solve this issue (see comments): https://github.com/WFP-VAM/prism-frontend/blob/2e44cef77702dc167d9bbd591083aac12eeddd6f/src/components/MapView/index.tsx#L138

It's a hacky fix to load the boundary layer super early since so much code depends on it, but we were running into issues where mapbox would load the layer way too late for anything else to use it beforehand. If this `before` fix works, then maybe we can move the loading to the BoundaryLayer file (where it belongs)

I recommend trying to move it quickly to see if it works. If not, all good probably not worth putting too much time into this ;)

JorgeMartinezG:
image


Tested locally with TJK shakemaps layer and works fine

JorgeMartinezG (approved):

src/components/MapView/utils.ts
erwagasore: I opted on using `vanilla` Javascript as you suggested previously. Thus reduce the bundle size

Ry-DS: that's fair! My stance on it is to only use lodash if it makes the code easier to read.
For example, the `map` and `reduce` functions by lodash don't serve much use for most cases since vanilla JS can do it and looks basically the same, but the use of `sumBy` here would make the code easier to read. If lodash can make the code look cleaner, I think it's worthwhile. However up to personal preference :+1:

PR #245, opened 2021-08-27
Modify regular expression to replace more than one dot

### LINKED ISSUE #243: Stats API gives a 500 error on prism.mmr.wfp.org
When running an analysis or exposure analysis, the API gives a 500 error. 
image

JorgeMartinezG:
This Pull Request closes #243 

On production environments, PRISM builds the admin boundaries URL for stats analysis using PATH parameter from layers config and server hostname. However, the regular expression removes only a single dot if the path is relative.

You can verify the PR here by running exposed population analysis and checking the zones_url within payload.

image

deployment site: http://demo-myanmar.surge.sh/

Also, please check this comment: https://github.com/WFP-VAM/prism-frontend/issues/243#issuecomment-907178795

### DIFF
--- a/src/context/analysisResultStateSlice.ts
+++ b/src/context/analysisResultStateSlice.ts
@@ -89,9 +89,9 @@ function getAdminBoundariesURL() {
   if (isLocalhost) {
     return defaultBoundariesFile;
   }
-  // the regex here removes the dot at the beginning of a path, if there is one.
+  // the regex here removes the dot(s) at the beginning of a path, if there is at least one.
   // e.g the path might be ' ./data/xxx '  instead of ' /data/xxx '
-  return window.location.origin + adminBoundariesPath.replace(/^\./, '');
+  return window.location.origin + adminBoundariesPath.replace(/^\.+/, '');
 }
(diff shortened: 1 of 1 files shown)


Ry-DS (approved):
Looks good! I'm a little concerned however why we need to go up a level with `../` when accessing admin boundaries since all other layers use a single dot `./` e.g here: 
https://github.com/WFP-VAM/prism-frontend/blob/ff4f63af99388c400f8f109c01105f5bdec55be0/src/config/indonesia/layers.json#L4
https://github.com/WFP-VAM/prism-frontend/blob/ff4f63af99388c400f8f109c01105f5bdec55be0/src/config/indonesia/layers.json#L13

In the future this could break if we go up two levels with `../..`. Maybe we could investigate here? Otherwise if there's a rational reason we can leave as is :+1:

JorgeMartinezG:
@Ry-DS all other layers are fetched within the react application. Relative paths are accepted. However, within a runanalysis, the frontend created a url for admin boundaries which is sent to the backend. In this case is the full path.

PR #249, opened 2021-09-01
Render baseline layer analysis button on an already executed exposed population

### LINKED ISSUE #248: Clear exposure analysis results when toggling off a hazard layer
If I run an exposure analysis on a storm, and then load another hazard layer (like rainfall), we continue to keep the exposure analysis. This would be ok except that if I then try to run an analysis (hazard + vulnerability), the 'run analysis' button doesn't appear

cc @JorgeMartinezG

JorgeMartinezG:
Closes #248 

This PR contains the following fixes:
- Render the run analysis button even an exposure analysis has been executed before.
- Clear any result when a new analysis is requested.

**Note.** Analysis are not cleared on a layer toggle event, but is possible if needed.

analysis


wadhwamatic (commented):
Looks great, thanks for the quick action on this Jorge. Let's have @erwagasore provide a review as well before we merge

erwagasore (approved):
Straight forward. Good to go @JorgeMartinezG

PR #255, opened 2021-09-10
Url parsing feature

JorgeMartinezG:
This PR has the following changes:
- On a layer selection or date change event, an update of the prism url is executed including or modifying query parameters.
- When loading prism from a url with hazardLayerId, baselineLayerId or date parameters, PRISM automatically loads the layers which are displayed on the map.
- For the case of hazard layer removal, the date url parameter is also removed.

**Note.** This feature does not include baselineLayer or exposureAnalysis url updates yet.

### DIFF
--- a/src/components/MapView/index.tsx
+++ b/src/components/MapView/index.tsx
@@ -35,12 +35,17 @@ import {
 import {
   DiscriminateUnion,
   LayerType,
+  LayerKey,
   BoundaryLayerProps,
 } from '../../config/types';
 
 import { Extent } from './Layers/raster-utils';
+import { useUrlHistory } from '../../utils/url-utils';
 
-import { getBoundaryLayerSingleton } from '../../config/utils';
+import {
+  getBoundaryLayerSingleton,
+  LayerDefinitions,
+} from '../../config/utils';
 
 import DateSelector from './DateSelector';
 import { findClosestDate } from './DateSelector/utils';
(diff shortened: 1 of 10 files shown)


wadhwamatic (commented):
Tested from a functionality perspective. This is great. I am noticing an increasing lag from the time you click on a layer's toggle to activate to the time that the toggle moves and the layer loads. I am not sure if this is related to this update or not.

Let's also have a code review from @ericboucher and/or @Ry-DS

src/utils/url-utils.ts
```diff
@@ -0,0 +1,102 @@
+import { useHistory } from 'react-router-dom';
+import moment from 'moment';
+
+export type URLParams = {
+  hazardLayerId?: string;
```
Ry-DS, line 5: If these are layer keys might as well type them as so? This will avoid unneeded casts down the line.

Ry-DS: You could also consider adding an assertion to ensure the values are indeed layer keys before storing them in the object

src/utils/url-utils.ts
```diff
@@ -0,0 +1,102 @@
...
+  baselineLayerId?: string;
+  date?: number;
+};
+
+/*
+  transforms the string value according to the field specified in the
+  type URLParams
```
Ry-DS, line 12: This function and the one below are clearly doing something with dates but they avoid the topic entirely. Maybe give them better names and docs like `parseIfDate` and consider combining the two to one function with a boolean to change between valueOf and format

src/utils/url-utils.ts
```diff
@@ -0,0 +1,102 @@
...
+      return moment(value).format('YYYY-MM-DD');
+    default:
+      return value as string;
+  }
+};
+
+/*
+  Serialized value to string depending on the parameter specified in the
+  type URLParams
+*/
```
Ry-DS, line 42: Pretty sure there exists existing util to do this for us. Check out `new URLSearchParams`

src/utils/url-utils.ts
```diff
@@ -0,0 +1,102 @@
...
+            ...obj,
+            [key]: parseValue(key, value),
+          }),
+          {} as URLParams,
+        );
+
+/*
+  Serialize urlParams object to string.
+*/
+const urlParamsToString = (params: URLParams): string =>
```
Ry-DS, line 61: Same as above, you can use URLSearchParams

src/utils/url-utils.ts
```diff
@@ -0,0 +1,102 @@
...
+    .map(([key, value]) => `${key}=${valueToString(key, value)}`)
+    .join('&');
+
+/*
+  This custom hook tracks the browser url string, which is defined by the useHistory hook.
+  We created additional functions to update the url based on user events, such as select date
+  or select layer.
+*/
+export const useUrlHistory = () => {
+  const { replace, location } = useHistory();
```
Ry-DS, line 72: This entire function can be simplified quite a lot by using existing tools like URLSearchParams. Better to use existing infrastructure so we can guarantee it works without bugs!

Ry-DS (commented):
Looking good! A few comments to try use existing tools more to simplify what we have to make/test ourselves.

src/components/MapView/index.tsx
```diff
@@ -246,7 +263,9 @@ function MapView({ classes }: MapViewProps) {
       );
     }
     // let users know if their current date doesn't exist in possible dates
-    if (selectedDate && urlParams.date !== selectedDate) {
+    const urlDate = urlParams.get('date');
```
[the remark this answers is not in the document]
JorgeMartinezG: @Ry-DS using `moment(undefined)` returns current timestamp. This might make issues and would need to redefine logic, since there is a check to include the date value in the url if it is not defined. I've included previous comment in commit a9c4a64f84eea8295bd99f6a40dfad80aef22bf4

wadhwamatic:
@JorgeMartinezG - it looks like we don't handle kobo / point_data layers in this PR. can we add this here?

PR #269, opened 2021-09-28
initial configuration for west africa regional prism (RBD)

wadhwamatic:


JorgeMartinezG (approved):
Just tested locally and layers are being displayed correctly. However, baseline analysis cannot be run since it requires a baselinelayer, and rbd does not have any. Check figure below

image

Maybe this should be reported as separate issue to disable baseline analysis button in case a country has 0 baseline layers

wadhwamatic:
> Just tested locally and layers are being displayed correctly. However, baseline analysis cannot be run since it requires a baselinelayer, and rbd does not have any. Check figure below
> 
> image
> 
> Maybe this should be reported as separate issue to disable baseline analysis button in case a country has 0 baseline layers

Thanks Jorge. This would not likely happen in any actual deployment, but good to address it when we have time. Here's an issue to track it: https://github.com/WFP-VAM/prism-frontend/issues/270

PR #276, opened 2021-10-15
Kobo toolbox forms integration

### LINKED ISSUE #274: Additional changes for Kobo Integration
- [x] Fetch and parse datetime fields **end** and **_submission_time**
- [x] Use kobo form parameters as layer parameters instead of including them within the URL request.
**Example:** Instead of setting the Kobo layer format.
```
  "kobo_test": {
    "title": "Kobo data",
    "type": "point_data",
    "data": "http://localhost/kobo/forms?nameField=Test MMR&dateTimeField=_submission_time&geomField=Location&measureField=The_number"
}
```
Use:
```
  "kobo_test": {
    "title": "Kobo data",
    "type": "point_data",
    "data": "http://localhost/kobo/forms",
   "nameField": "Test MMR",
   "dateTimeField": "_submission_time",
   "geomField": "Location",
   "measureField": "The_number"
}
```

- [x] Include validation status label
- [ ] Look into issue where timezone difference between form data and PRISM user seem to cause the timeline ticker to fail.

JorgeMrtnzG:
closes #274 

This PR enables prism backend to fetch form responses from Kobo toolbox API.

### DIFF
--- a/api-flask/docker-compose.yml
+++ b/api-flask/docker-compose.yml
@@ -10,5 +10,5 @@ services:
     environment:
       - DATABASE_URL
       - WORKERS_PER_CORE=1
-      - KOBO_USERNAME=jorgemartinezg
-      - KOBO_PW=Jorgemg_0327
+      - KOBO_USERNAME=changeme
+      - KOBO_PW=changeme


src/components/MapView/Layers/PointDataLayer/index.tsx
```diff
@@ -51,18 +64,23 @@ function PointDataLayer({ layer }: { layer: PointDataLayerProps }) {
       circleLayout={circleLayout}
       circlePaint={circlePaint}
       circleOnClick={(evt: any) => {
-        dispatch(
```
erwagasore, line 54: In line with a converstation I had with Amit. The default `data_field` or `mesure` in this case should remain there and add `feature_info_props` as extra fields.

PR #288, opened 2021-10-28
Expand filtering capabilities for kobo form responses

### LINKED ISSUE #287: Expand filtering capabilities for form / point_data
There are use cases where form data that is returned through the API will need to be filtered before rending the layer in PRISM. Currently, we have the ability to filter based on form status, but we should expand this to also filter based on the value of a given field. As an example, we could modify the current filter to be a list which would be something like the following, where disaster_type is a form field with strings:
        "filters":{
          "status":"approved",
          "disaster_type":"flood"
        }

JorgeMartinezG:
closes #287


src/config/myanmar/layers.json
```diff
@@ -1197,7 +1197,10 @@
       "datetime_field": "date_time",
       "geom_field": "location",
       "measure_field": "number",
-      "kobo_url": "https://kobo.humanitarianresponse.info/api/v2/assets.json"
+      "kobo_url": "https://kobo.humanitarianresponse.info/api/v2/assets.json",
+      "filters": {
+        "status": "Approved"
+      }
```
erwagasore, line 1203: Tested Myanmar `Kobo data` and since you are filtering by `status` to equal `Approved` we should only see approved `point_data` only. I still am seeing `On hold`.

erwagasore: I have added `status` into `feature_info_props` to visual see if on the map I will only see `Approved` and still I had `On hold`. Correct me if my assumption on `filters` is wrong.

Here is the screenshot for clarity
Screen Shot 2021-11-02 at 6 24 33 PM

erwagasore (approved):
Looks good @JorgeMartinezG

PR #295, opened 2021-11-11
Fix kobo form dates and prioritize WMS layers when layer dates inters…

### LINKED ISSUE #293: Activating ADAM storm layer, followed by Kobo layer causes app to hang
This is happening consistently on master for Myanmar. When I activate the ADAM tropical storm layer and then activate a Kobo layer, the application hangs. This does not occur when testing ODC WMS layers, only ADAM layers.

JorgeMartinezG:
…ection is empty array

This PR closes #293 

Reviewing the behavior of prism from master branch, the application enters an infinite date updating loop. This is due to the API response for the kobo endpoint contains time values (YYYY-MM-DDTHH: MM: ss) which causes a PRISM status update with incorrect dates.

This PR solves the previous issues. In addition, changes were made where the WMS dates are used at the time when, having multiple layers where dates do not intersect.

**Note** Once the PR is merged, we need to update the API


wadhwamatic:
I came across an issue when there is a mismatch between the WMS layer's date and the Kobo layer. If I activate a Kobo layer and choose a date that doesn't match a WMS layer, the timeline and the calendar picker no longer show available dates beyond the currently active date.

For example: Myanmar, activate Kobo layer on 18-10-2021. Then activate 10-day rainfall estimate. 21-10-2021 rainfall layer will load but no other dates are available. The same happens if you activate a WMS layer first and then activate the Kobo layer.

I'm not sure if there is a short term solution here or if we need to address the larger question around conflicting dates first. If possible, I'd suggest just having the most recently activated layer takeover the timeline for now while we figure out the longer term solution on dates.

PR #299, opened 2021-11-12
Parse kobo form response fields

### LINKED ISSUE #290: Editing a form in Kobo changes the data type and breaks numeric data field in PRISM
This seems to be a Kobo issue more than anything but noting it to see if we can explore some way to handle it. Steps to create:

1. Create a form in Kobo with a numeric input field which is required with the condition of accepting any value >=0
2. Configure PRISM to use this form and use the numeric field as the measure attribute
3. Collect data in Kobo, with a '0' for the numeric field
4. Form data should appear as expected
5. Now edit a form entry, specifically change the numeric field to any other valid value
6. From what I see, the form field is now text with the numeric value wrapped in single quotes. This breaks the functionality on the PRISM side with no form entries displayed as there is no valid numeric data

JorgeMartinezG:
closes #290


wadhwamatic:
This works for Myanmar as expected, but when I tried the Cambodia version, I get a 500 error:

"error": "Internal server error: '<=' not supported between instances of 'datetime.datetime' and 'str'"

Here's the form configuration in layers.json:

"incident_report": {
    "title": "Disaster impact report",
    "type": "point_data",
    "data": "http://localhost/kobo/forms",
    "date_url": "http://localhost/kobo/forms",
    "additional_query_params": {
      "form_name": "PRISM-KHM-Disaster-Report-v1",
      "datetime_field": "disaster_date",
      "geom_field": "location",
      "kobo_url": "https://kobo.humanitarianresponse.info/api/v2/assets.json",
      "filters": {
        "status": "Approved"
      }
    },
    "feature_info_props": {
      "disaster_type": {
        "type": "text",
        "label": "Type of disaster"
      },
      "num_ppl_affected": {
        "type": "number",
        "label": "Number of people affected"
      }
    },
    "opacity": 0.9,
    "measure": "num_ppl_affected",
    "legend_text": "Number of people affected",
    "legend": [
      {"value": "0", "color": "#909090"},
      {"value": "< 100", "color": "#ffeda0"},
      {"value": "100 - 500", "color": "#feb24c"},
      {"value": "500 or more", "color": "#f03b20"}
    ]
  }

wadhwamatic (changes requested):
shouldn't 'measure' be included in the parsing?

JorgeMartinezG:
@wadhwamatic now we can get from kobo the type for each field with the goal of parsing each value. Because of this, we do not need to send the `measure` field to the PRISM API. 
```
"error": "Internal server error: '<=' not supported between instances of 'datetime.datetime' and 'str'"
```

This error happens because Cambodia forms have a `date` field instead of `datetime`. Fix included in 07a8d40365922a8a0f17c695e6f17063f97a7306

wadhwamatic (approved):

wadhwamatic:
> @wadhwamatic now we can get from kobo the type for each field with the goal of parsing each value. Because of this, we do not need to send the `measure` field to the PRISM API.
> 
> ```
> "error": "Internal server error: '<=' not supported between instances of 'datetime.datetime' and 'str'"
> ```
> 
> This error happens because Cambodia forms have a `date` field instead of `datetime`. Fix included in [07a8d40](https://github.com/WFP-VAM/prism-frontend/commit/07a8d40365922a8a0f17c695e6f17063f97a7306)

Got it. Thanks for the explanation and changes. All looks good from my side.

PR #305, opened 2021-11-18
Feature/admin level time dimension

erwagasore:
Added support for `time-dimension` to `admin-level-data` and allowed date range with `startDate` and `endDate`.


erwagasore:
@wadhwamatic there was an issue with `wms` layer with unmatching dates with `admin-level-data` was causing the app to freeze due to `infinite` loop. I had a discussion with @JorgeMartinezG and he told me that we are aware of the issue and plan to have a discussion on how to better handle it. In the mean time I just implemented his fix.

wadhwamatic:
When I load a WMS layer (10-day rainfall) with the latest data (11-Nov-2021) and then activate the 'Admin Level With Date Range' layer, the message 'No data was found for the layer 'Admin Level With Date Range' on 2021-11-11' appears. The WMS layer changes to 1-April-2021 but the related admin level data does not load. The timeline and calendar picker are also empty. See screenshot #1 below.

When I do the same as above but choose the 'Admin Level With Dates' layer instead, the layer does load but the user isn't made aware of the change in date to 1-April-2021. In addition, the timeline and calendar picker are empty. See screenshot #2

Screen Shot 2021-11-18 at 11 51 27

Screen Shot 2021-11-18 at 12 11 46

wadhwamatic:
Closing this PR as we'll revisit dates more holistically in the near future

Issue #292, opened 2021-11-08
Add a dropdown with a list of available admin areas for AOI navigation


wadhwamatic:
As we prepare for regional deployments, the ability to specifically choose and AOI to zoom in on is a requirement. This would require an admin selection feature that we could include a hierarchy of admin areas available, and after selecting, zoom in on that admin area.

We have some components of this addressed in the alerts module where there is a list of available admin areas shown to a user. However, it doesn't have a hierarchy implemented and it doesn't modify the display.

wadhwamatic:
Coordinate with https://github.com/WFP-VAM/prism-frontend/issues/278

wadhwamatic:
This was addressed with PR 329 https://github.com/WFP-VAM/prism-frontend/pull/329

Issue #284, opened 2021-10-25
Add style parameters to admin boundary files


wadhwamatic:
The styles used for admin boundaries are currently universal. We'd like to be able to allow each deployment to configure how boundaries are displayed. This includes increasing thickness for admin 1 boundary in relation to admin 2 boundaries, and so forth for each level. 

See also https://github.com/WFP-VAM/prism-frontend/issues/169

wadhwamatic:
See also  https://github.com/WFP-VAM/prism-frontend/issues/293

PR #306, opened 2021-11-19
Style and include multiple admin boundaries as separate layers

JorgeMartinezG:
This Pull Request allows PRISM to read and render multiple admin boundary layers within mapbox. Also, **line** and **fill** mapbox styles are now defined from layers.json configuration file

image


src/config/utils.ts
```diff
@@ -129,21 +129,29 @@ export const LayerDefinitions: LayersMap = (() => {
...
-export function getBoundaryLayerSingleton(): BoundaryLayerProps {
+export function getBoundaryLayers(): BoundaryLayerProps[] {
   const boundaryLayers = Object.values(LayerDefinitions).filter(
     (layer): layer is BoundaryLayerProps => layer.type === 'boundary',
   );
-  if (boundaryLayers.length > 1) {
+  if (boundaryLayers.length === 0) {
     throw new Error(
-      'More than one Boundary Layer defined! There should only be one boundary layer in layers.json',
+      'No Boundary Layer found! There should be exactly one boundary layer defined in layers.json.',
```
wadhwamatic, line 138: This is no longer true.

wadhwamatic:
Tested and this works as expected. We should next introduce the concept of levels / order. See https://github.com/WFP-VAM/prism-frontend/issues/292. 

Minor request to update the message around 'only 1 admin boundary file'. See in-line comment https://github.com/WFP-VAM/prism-frontend/pull/306/files#r753441710

src/config/utils.ts
wadhwamatic: Suggested alternative text:
"No boundary layer found. There should be at least one boundary layer defined in layers.json"

wadhwamatic (approved):
Good to go. Thanks

ericboucher:
Sorry I am seeing this PR a bit late but it came up while chatting about #338

wadhwamatic:
> Sorry I am seeing this PR a bit late but it came up while chatting about #338
> 
> Can I get a bit more context on why this is needed? What is the excpected behavior for `getBoundaryLayerSingleton`? Should it grab one of the layers? a merge of the layers? something else?

@ericboucher - this was to address https://github.com/WFP-VAM/prism-frontend/issues/284; we need unique styles depending on the admin level

PR #308, opened 2021-11-22
Improve UI to select AOI for an alert

### LINKED ISSUE #226: Improve UI to select AOI for an alert
The current UI to choose which admin areas to include is cumbersome. Two improvements to consider:

1. Use the hierarchy of admin areas to allow a single select of an admin level-1 to apply to all admin level-2 below it
image

2. Allow the map to become a multi-select tool (i.e. click on admin areas of interest)	

https://staging-prism-frontend.web.app

OLD UI:
image

Doc: https://docs.google.com/document/d/1bgPC1WeThqjPE32t9jj8eubiym2mfGTc_oqd2k5nFeM/edit#heading=h.igzdohvfar19

Ry-DS:
Closes #226
Deploy: http://prism-selectable-map.surge.sh

### TODO: 
- [x] Cleanup code
- [x] Fix a bug with how overlays are rendered (the way it's done right now prevents interaction with the map)
- [x] Find a way to support selecting all.
- [x] Maybe search if it's easy?
- [x] Decide on a select color (currently green)

### Features:
- Fixed the heading overlapping at the top for both layer dropdown and boundary dropdown
- Made the layerdropdown a new component so future forms which need boundary selection will be very easy to make (map syncing is handled for you)
- If you click the region header, the entire region is selected automatically.

### Dev Notes:
- ~I moved from `@turf/bbox` to `@turf/turf`. This might significantly increase our bundle size, however this might not happen if webpack is smart and doesn't bundle the entire library (only the parts we use). We should investigate to see if there are any performance drops.~ after testing, confirmed tree shaking doesn't happen, so moved back to individual packages.
- Refactored `selectedDate` in DateSelector significantly to avoid unnecessary duplication of state, and avoid using refs as a bubbling method of state (an anti-pattern in react). We need to test to ensure this refactoring doesn't break anything)

image


src/components/MapView/Layers/BoundaryDropdown.tsx
```diff
@@ -0,0 +1,166 @@
...
+                {category.title}
+              </Typography>
+            </ClickableListSubheader>,
+            ...category.children.map(({ label, value }) => (
+              <MenuItem key={value} value={value}>
+                {label}
+              </MenuItem>
+            )),
+          ],
+          [] as ReactElement[],
```
Ry-DS, line 154: this type should be in a generic

src/components/MapView/Layers/BoundaryDropdown.tsx
```diff
@@ -0,0 +1,166 @@
...
+        )}
+      </Select>
+    </FormControl>
+  );
+}
+
+interface BoundaryDropdownProps {
+  // a way to type this and every other default thing in? e.g style
```
Ry-DS, line 162: remove comment

src/components/MapView/Layers/SelectionLayer/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const selectedBoundaries = useSelector(getSelectedBoundaries);
+  const boundaryLayerState = useSelector(
+    layerDataSelector(boundaryLayer.id),
+  ) as LayerData<BoundaryLayerProps> | undefined;
+  const { data } = boundaryLayerState || {};
+  if (!data || !isSelectionMode) {
+    return null;
+  }
+
+  // We use the legend values from the config to define "intervals".
```
Ry-DS, line 32: Incorrect comment

src/config/index.ts
```diff
@@ -116,10 +103,13 @@ const configMap = {
   },
 } as const;
 
+type Country = keyof typeof configMap;
+
+const DEFAULT: Country = 'mozambique';
```
Ry-DS, line 108: remember to change default country back

src/config/mozambique/prism.json
```diff
@@ -1,5 +1,6 @@
 {
   "country": "Mozambique",
+  "alertFormActive": true,
```
Ry-DS, line 3: remember to undo these if they shouldn't be on

src/context/store.ts
```diff
@@ -3,34 +3,34 @@ import {
...
-  // serialize the state
   middleware: getDefaultMiddleware({
+    // TODO: Instead of snoozing this check, we might want to
+    // serialize the state
     serializableCheck: false,
-    immutableCheck: {
-      ignoredPaths: ['mapState.layersData', 'analysisResultState.result'],
-    },
+    // TODO this is incredibly slow in development
+    immutableCheck: false,
```
Ry-DS, line 33: undo after development finishes

wadhwamatic:
Great work so far on this. A few inputs from my side:
1. Can we make the visual distinction between admin levels in the list of locations more clear? Maybe color the background of provinces in addition to the differing font?
2. A tooltip on the region selection list that notes that locations can be chosen from the list or the interactive map would be helpful
3. The suggestion to add a search feature is a good one. If this is relatively simple to do, we'd like to see it included.

wadhwamatic:
@Ry-DS - two additional inputs:
1. We should consider either hiding data layers that are active when we switch to the region selector, or change the visual feedback on region select to a boundary line color change rather than polygon fill. Otherwise we can end up with a confusing outcome (see screenshot below)
2. We might need a more explicit way to exit the region selection mode. Currently, you have to minimize the alert dropdown to achieve this. Any suggestions on where in the UI we could do this?


Screen Shot 2021-11-24 at 10 48 17

ericboucher:
1. Can you think of a color that would make it less confusing because it is unlikely to overlap with data layers? Purple maybe?
2. For the alert box, we could hide it when people click outside of the box and outside of the clickable map. What do you think @wadhwamatic ?

wadhwamatic:
> 1. Can you think of a color that would make it less confusing because it is unlikely to overlap with data layers? Purple maybe?
> 2. For the alert box, we could hide it when people click outside of the box and outside of the clickable map. What do you think @wadhwamatic ?
> 
> Also, would you want the alerts to be active by default or should we turn them off after testing?

1. See my earlier comment - an outline instead of fill could be a better option. If we must use a fill, I'd suggest black. 
2. Yes, that could work
3. We should have them off by default

wadhwamatic:
This is looking great, thanks Ryan. I found that the search input lost focus on a few occasions when testing. It seems to happen when I use the search feature to find a region, select it, and then return to search to try another location. 

It also seems that if I search, then move over to the interactive map selection, and then return to search, I can't clear out the previous search.

Ry-DS:
@wadhwamatic Fixed! Let me know if it works :+1:

wadhwamatic:
issues with search are fixed, thanks @Ry-DS. I believe @JorgeMartinezG had some thoughts on the technical implementation that he'll share with you soon

src/components/MapView/index.tsx
```diff
@@ -331,38 +334,57 @@ function MapView({ classes }: MapViewProps) {
...
+                              data: result[k],
+                              coordinates: evt.lngLat,
+                            },
+                          }),
+                        );
+                      });
+                    }
+                    dispatch(setWMSGetFeatureInfoLoading(false));
+                  },
+                );
```
ericboucher, line 384: let's extract this into a separate function to keep it clean

Ry-DS: Yup definitely good since this is getting quite complex

src/components/MapView/DateSelector/index.tsx
```diff
@@ -124,11 +120,12 @@ function DateSelector({
       x: dateIndex * TIMELINE_ITEM_WIDTH,
       y: 0,
     });
-    setSelectedDate(moment(stateStartDate));
```
Ry-DS, line 127: I was really confused why we are managing two variables both tracking the same thing (stateSelectedDate). I removed the state in favor of redux to reduce coupling and make things easier to read/track. Let me know if this was done purposefully.

src/components/MapView/Layers/BoundaryDropdown.tsx
```diff
@@ -0,0 +1,260 @@
...
+            )),
+          ],
+          [],
+        )}
+      </Select>
+    </FormControl>
+  );
+}
+
+interface BoundaryDropdownProps {
```
Ry-DS, line 256: New prop ideas for @JorgeMartinezG to make this support single selection:
- `menuMode` if given makes the component act more like a dropdown menu than a selector. This can mean replacing the current Select parent with Menu (children should be the same)
- `onSelect` listener which could override default behavior (saving to redux state)

Or we could follow a HOC (higher order component) pattern where we have a super simple base component, and 2 extra components (one which provides multi select functionality and saves to redux state, and another which handles the goto feature). Might need more work however.

src/components/MapView/Legends/index.tsx
```diff
@@ -193,6 +193,13 @@ function LegendItem({
   const map = useSelector(mapSelector);
   const analysisResult = useSelector(analysisResultSelector);
   const dispatch = useDispatch();
+  useEffect(() => {
+    // should this be here? Or somewhere more related to analysis?
```
Ry-DS, line 197: Confused why this is here. Also confused why we left what looks like an expensive operation without a useEffect.

src/components/MapView/DateSelector/index.tsx
Ry-DS: (should test thoroughly to make sure nothing is broken with this refactoring)

ericboucher:
@Ry-DS does this PR change anything? https://github.com/WFP-VAM/prism-frontend/pull/333

alerting/src/utils/analysis-utils.ts
```diff
@@ -1,6 +1,6 @@
 import fetch from 'node-fetch';
 import { get, isNil } from 'lodash';
-import bbox from '@turf/bbox';
+import { bbox } from '@turf/turf';
```
JorgeMartinezG, line 3: it is recommended to use subpackages. Turf library itself might increase bundle size drastically, and currently prism does not make use of a lot of turf functions

src/components/MapView/Layers/BoundaryDropdown.tsx
```diff
@@ -0,0 +1,260 @@
...
+      !searchIncludes(label) &&
+      !searchIncludes(code) &&
+      !searchIncludes(parentCategory)
+    ) {
+      return;
+    }
+    // add to categories if exists
+    const category = categories.find(c => c.title === parentCategory);
+    if (category) {
+      // eslint-disable-next-line fp/no-mutating-methods
```
JorgeMartinezG, line 125: We could generate the output without disabling eslint. I made a similar function here: https://github.com/WFP-VAM/prism-frontend/pull/329/files#diff-a7f8ac7902cdbf14643da46189b14ee847a5169165fbf0f8fbfb9248fb96f62dR180

src/components/MapView/Layers/BoundaryDropdown.tsx
JorgeMartinezG: Agree with second option. The base component should receive the event handler

JorgeMartinezG:
@Ry-DS not sure if redux for managing globally the state for selected admin boundaries is needed, since not many components make use of it. Another option might be to read from redux the mapbox map object and handle interaction within the AlertForm.

alerting/src/utils/analysis-utils.ts
ericboucher: 👍

src/components/MapView/Layers/BoundaryDropdown.tsx
Ry-DS: I used `reduce` to remove one eslint disable, but to remove both I may need to use some intermediary data structures like an object to only switch back to the 2d array needed at the end. IMO it's not worth it to complicate things too much when the current solution can make sense (sacrificing mutability for readability)

Ry-DS:
> @Ry-DS not sure if redux for managing globally the state for selected admin boundaries is needed, since not many components make use of it. Another option might be to read from redux the mapbox map object and handle interaction within the AlertForm.

What do you mean exactly? I don't see another way to share state without a redux state to allow the selection layer and boundary dropdown to communicate with each other. Maybe we could use `useContext` instead since it's simpler? Not too sure about calling methods inside the mapbox object:
1. Not too sure what we could do in the first place, would the mapbox map object have methods which can color certain cells on a layer? If you have any examples that would be awesome :eyes:
2. Not too comfortable with calling methods to mutate internal state we don't track, since that break's react's declarative nature (the map holds onto state we don't control). Or maybe I'm misunderstanding here, let me know how this would work exactly.

src/components/MapView/Layers/BoundaryDropdown.tsx
Ry-DS: I did e7eea9159c884d458aaf392b5b3ec707d57907e7 to convert the component to HOC. You can extend on it in #329 for it to behave more like a menu ;)

JorgeMartinezG:
@Ry-DS my mistake. Since this task requires to create additional layers on demand, is correct to use a new layer type and state tracking. My idea required to use mapSelector and interact with mapbox map object, but it means losing consistency across the project.

I agree with using redux for state management. However, since `AlertForm` is an optional component, I do not think we should have a state created for this globally. As you mentioned, having a context only for `AlertForm` and `SelectionLayer` created within `MapView` component is a good idea.

src/components/MapView/Legends/index.tsx
ericboucher: Did you find out? Or should I have a look @Ry-DS ?

src/components/MapView/Layers/BoundaryDropdown.tsx
```diff
@@ -101,8 +102,11 @@ const getCategories = (
     }>
   >((ret, feature) => {
     const parentCategory =
-      feature.properties?.[boundaryLayer.adminLevelNames[1]];
-    const label = feature.properties?.[boundaryLayer.adminLevelNames[0]];
+      feature.properties?.[boundaryLayer.adminLevelNames[0]];
+    const label =
+      feature.properties?.[
+        last(boundaryLayer.adminLevelNames) || boundaryLayer.adminLevelNames[1]
```
ericboucher, line 108: can you explain what's happening here? when would `last` return a nullish value here but `boundaryLayer.adminLevelNames[1]` would not?

Ry-DS: I took a shortcut here to make TS happy haha. `last` returns `string | undefined` which I needed to convert to `string`. I'll make this cleaner...

ericboucher: You could say you were just testing my reviewing skills ;)

Ry-DS:
> @Ry-DS my mistake. Since this task requires to create additional layers on demand, is correct to use a new layer type and state tracking. My idea required to use mapSelector and interact with mapbox map object, but it means losing consistency across the project.
> 
> I agree with using redux for state management. However, since `AlertForm` is an optional component, I do not think we should have a state created for this globally. As you mentioned, having a context only for `AlertForm` and `SelectionLayer` created within `MapView` component is a good idea.

IMO on second thought, I'm not too sure about introducing useContext when we already use redux. If for some reason we do want to use this state somewhere else, it would be confusing initially if a dev can't access it wherever they want. I think the slice separation is good enough to keep this state distinct. Or maybe we can include it inside an existing slice. Overall not a fan of having 2 separate ways of doing the same thing. What are your thoughts here?

wadhwamatic:
@Ry-DS - a few comments:
1. If you select a higher level admin area after having selected a lower one within it, it clears out all areas and doesn't select the higher admin area. i.e. select a township in myanmar, then select the district above. nothing is selected. I think this was working differently before?
2. Is it possible to order alphabetically? It seems the second level is, but not the first in the case of Mozambique (or any two level country)
3. For Myanmar where we have 3 levels, I think there's the 2nd level is not exposed to the user. That's ok but the order within the group would maybe make more sense if it was also alphabetical, as noted above

Ry-DS:
> @Ry-DS - a few comments:
> 
>     1. If you select a higher level admin area after having selected a lower one within it, it clears out all areas and doesn't select the higher admin area. i.e. select a township in myanmar, then select the district above. nothing is selected. I think this was 

Ah I initially made it so you deselect if even one is selected, but the approach you mentioned would be better for the user. Updated all three items the latest commit :+1:

Also keep in mind, for this feature to work, the `adminLevelNames` array in the layers.json file needs to be ordered from largest (states) to small (towns/districts). See commit 98d58e2 where I revered the names to make sure the boundary dropdown works.

wadhwamatic:
Thanks @Ry-DS ! Looks good on the surge deployment, but I'm not seeing changes on the branch so can't verify this for another country. I'm testing with Mozambique FYI.

Ry-DS:
> Thanks @Ry-DS ! Looks good on the surge deployment, but I'm not seeing changes on the branch so can't verify this for another country. I'm testing with Mozambique FYI.

Sorry, forgot to push. Should be testable now on your end.

wadhwamatic:
> Sorry, forgot to push. Should be testable now on your end.

Looks good now, thanks @Ry-DS !

wadhwamatic (approved):
approved from my side. @ericboucher - ok to merge if this also looks good to you

PR #309, opened 2021-11-22
Feat/load default layer

erwagasore:
Added support for default layer.
The layer is being loaded as URL search parameter only if the URL doesn't have params already set

### DIFF
--- a/src/components/MapView/index.tsx
+++ b/src/components/MapView/index.tsx
@@ -13,7 +13,7 @@ import {
   WithStyles,
   withStyles,
 } from '@material-ui/core';
-import { countBy, pickBy } from 'lodash';
+import { countBy, get, pickBy } from 'lodash';
 import moment from 'moment';
 // map
 import ReactMapboxGl from 'react-mapbox-gl';
@@ -148,6 +148,17 @@ function MapView({ classes }: MapViewProps) {
 
     const hazardLayerId = urlParams.get('hazardLayerId');
     const baselineLayerId = urlParams.get('baselineLayerId');
+
+    /*
+      In case we don't have hazard or baseline layers we will use the default
+      layer provided in the appConfig defined within `prism.json` file.
+     */
+    if (!hazardLayerId && !baselineLayerId) {
+      const defaultLayer = get(appConfig, 'defaultLayer', '');
+      if (defaultLayer.length !== 0) {
+        updateHistory('baselineLayerId', defaultLayer);
+      }
+    }
 
     if (
       (!hazardLayerId && !baselineLayerId) ||
--- a/src/config/cambodia/prism.json
+++ b/src/config/cambodia/prism.json
@@ -1,6 +1,7 @@
 {
   "country": "Cambodia",
   "alertFormActive": false,
+  "defaultLayer": "hydra_s1",
   "icons": {
     "vulnerability": "icon_vulnerable.png",
     "exposure": "icon_basemap.png",
--- a/src/config/global/prism.json
+++ b/src/config/global/prism.json
@@ -1,6 +1,7 @@
 {
   "country": "Global",
   "alertFormActive": false,
+  "defaultLayer": "rainfall_dekad",
   "icons": {
     "rainfall": "icon_rain.png",
     "vegetation": "icon_veg.png",
--- a/src/config/indonesia/prism.json
+++ b/src/config/indonesia/prism.json
@@ -9,6 +9,7 @@
     "tables": "icon_table.png"
   },
   "alertFormActive": false,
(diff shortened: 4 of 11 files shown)


wadhwamatic:
An admin_level_layer as default seems to work without issue. But when testing with a WMS layer as default, the boundary file never loads and then the application hangs. Did these work on your side?

Let's also avoid trying to merge a PR that changes the configuration of multiple deployments with a test scenario.

erwagasore:
@wadhwamatic solved the issue with `wms` layers you experienced and cleaned up `prism.json` from other countries remained with only `Myanmar` as test country with `defaultLayer` inside `prism.json`. Handled the case where the country don't have `defaultLayer` setup to just show the normal `admin_boundaries`

- Test country: Myanmar
- Default Layer: `inform` can be changed to any layerId
- Tested with at least one layer in each categories in Myanmar.

wadhwamatic:
Thanks, this is working now for all layer types (WMS, admin, point). @JorgeMartinezG - please provide code review

src/components/MapView/index.tsx
```diff
@@ -150,6 +150,22 @@ function MapView({ classes }: MapViewProps) {
     const hazardLayerId = urlParams.get('hazardLayerId');
     const baselineLayerId = urlParams.get('baselineLayerId');
 
+    /*
+      In case we don't have hazard or baseline layers we will use the default
+      layer provided in the appConfig defined within `prism.json` file.
+     */
+    if (!hazardLayerId && !baselineLayerId) {
+      const defaultLayer = get(appConfig, 'defaultLayer', '');
+      const layer = LayerDefinitions[defaultLayer as LayerKey];
```
Ry-DS, line 159: Would be nice if we showed a notification using the notification context if the defaultLayer param is invalid.

src/components/MapView/index.tsx
```diff
@@ -150,6 +150,22 @@ function MapView({ classes }: MapViewProps) {
...
+      if (layer) {
+        const urlLayerKey =
+          layer.type === 'admin_level_data'
+            ? 'baselineLayerId'
+            : 'hazardLayerId';
```
Ry-DS, line 164: these two strings should be constants so we don't lose track of them.

Ry-DS (commented):
Small comment about adding a notification if there's an invalid default layer. Thoughts?

src/components/MapView/index.tsx
erwagasore: Solved

```
    const HAZARD_LAYER_PARAM = 'hazardLayerId';
    const BASELINE_LAYER_PARAM = 'baselineLayerId';
```

src/components/MapView/index.tsx
erwagasore: Solved as well

```
if (Object.keys(LayerDefinitions).includes(defaultLayer)) {
        const layer = LayerDefinitions[defaultLayer as LayerKey];
        const urlLayerKey =
          layer.type === 'admin_level_data'
            ? BASELINE_LAYER_PARAM
            : HAZARD_LAYER_PARAM;
        updateHistory(urlLayerKey, defaultLayer);
      } else {
        dispatch(
          addNotification({
            message: `Invalid default layer identifier: ${defaultLayer}`,
            type: 'error',
          }),
        );
      }
```

JorgeMartinezG (commented):
@erwagasore either we could add case for missing default_layer within prism.json file, or include the parameter for each country

wadhwamatic:
> @erwagasore either we could add case for missing default_layer within prism.json file, or include the parameter for each country

I'd suggest we allow for missing default layers rather than including the parameter for each country

src/components/MapView/index.tsx
wadhwamatic: @erwagasore - the notification for an invalid default layer shows 5x. See screenshot below.

Screen Shot 2021-11-30 at 18 04 05

erwagasore: Solved. Good catch @wadhwamatic. This was being caused by state change in `useEffect` added a local state as a boolean to only allow one attempt to set the default layer.

erwagasore:
> > @erwagasore either we could add case for missing default_layer within prism.json file, or include the parameter for each country
> 
> I'd suggest we allow for missing default layers rather than including the parameter for each country

@JorgeMartinezG to empasize what @wadhwamatic said. We try to avoid changes which will force us to re-configure every layer unless it is absolutely un-avoidable

wadhwamatic:
looks good from my side. @JorgeMartinezG or @Ry-DS - can you take a final look at the last commit?

Ry-DS (approved):
Looks great!

PR #321, opened 2021-11-30
Added support for unique admin layer boundary

erwagasore:
- Added support for specifying a unique boundary for an `admin-level-data` layer by just providing `boundary` as a valid `boundary` layer.

- Changed how we define boundaries by adding `display` boolean property. With the `display` property we load only boundaries which have `true` as display value

- Test country: Myanmar all boundaries are now set with `display: true`
- Test admin-level-data layer: `Vulnerability Index` with `boundary: inform`

### DIFF
--- a/src/components/MapView/Layers/AdminLevelDataLayer/index.tsx
+++ b/src/components/MapView/Layers/AdminLevelDataLayer/index.tsx
@@ -3,7 +3,11 @@ import { useSelector, useDispatch } from 'react-redux';
 import { get } from 'lodash';
 import { GeoJSONLayer } from 'react-mapbox-gl';
 import * as MapboxGL from 'mapbox-gl';
-import { AdminLevelDataLayerProps, LayerKey } from '../../../../config/types';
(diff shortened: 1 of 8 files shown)


wadhwamatic:
A few comments:
1. Don't make the display attribute required. It should default to false
2. I added another layer to test which uses admin 1 boundaries but it gives errors when it loads. Please try with the new layer I added. 
3. The scenario we should be thinking about has a new layer loaded which references a different set of admin boundaries and codes

wadhwamatic:
@erwagasore - i'm not still seeing admin1 level data projected onto the map layer as expected when i test the 'test_admin1_data' layer in the myanmar config

Screen Shot 2021-12-06 at 12 46 11

erwagasore:
PR pointing to a wrong branch. Re-creating a new PR and omitting this one

erwagasore:
👉🏾  https://github.com/WFP-VAM/prism-frontend/pull/338

Issue #262, opened 2021-09-21
Need to handle null values on WCS requests


wadhwamatic:
When we run an analysis, the WCS response currently provides data which may include values that are 'no value' or null in terms of how they should be utilized. We're not handling this on the PRISM side. This can be seen when we run analysis on a flood layer. The flood layer is binary (0=not flooded, 1=flooded), but the values returned range from 0 - 255, where 255 is meant to be 'no value'. See an example of results below:
<img width="581" alt="Screen Shot 2021-09-21 at 16 38 22" src="https://user-images.githubusercontent.com/3343536/134261443-fdc6c3bf-ddf0-464b-9ffa-e3ce1e891764.png">

We should add a parameter to the wcsConfig layer property for 'no-data' to specify values which should be disregarded during an analysis. The handling of no-data values will need to be implemented in the Stats API. 

@valpesendorfer - are there 'no-data' values we need to handle for other OWS layers? It seems the case for xlh_dekad and dlx_dekad. Any others you're aware of?

valpesendorfer:
@wadhwamatic looks like there was a misconfiguration of the nodata attribute in the OWS layer config, wrongly setting it to `0`. This has now been fixed and set to `255` for the hfs layers. 

In general, the GeoTiffs returned should have the appropriate nodata value set, which can be used for downstream processing.

For example, running gdalinfo on a `hfs1_sfw_mask` tiff returns:

```
...
Band 1 Block=256x256 Type=Byte, ColorInterp=Gray
  Description = band
  NoData Value=255
...
```

PR #332, opened 2021-12-03
Filter out nullish values during analysis

ericboucher:
Filter out nullish values during an analysis.

This way, when running flood analysis, we only get matching admin boundaries when the boundary has been affected:

Screen Shot 2021-12-03 at 9 44 29 PM


wadhwamatic (approved):
Tested for flood scenarios in Cambodia and Myanmar. This does what is needed. We still have some rounding issues but that is not a major concern at the moment. This PR addresses https://github.com/WFP-VAM/prism-frontend/issues/262

Issue #311, opened 2021-11-23
mVAM / hungermap monitoring data integration


wadhwamatic:
For three of our upcoming deployments (Namibia, Zimbabwe, and Mozambique), we need to support admin_level_data which is retrieved from the Hunger Monitoring unit's API. The API includes a few metrics which can vary from country to country. On the PRISM side, we would want to fetch metrics of interest for a given country and make them accessible as individual layers. 

There are a few challenges:
1. The API uses admin boundaries and codes from a common tool called the GeoEnabler managed by WFP's Emergencies Division
2. The API currently only exposes the last 100 days of data. 
3. Notably, every day in the last 100 days has valid day. But there is not a dates available request from the API as of yet

Starting with Namibia has an example, we want to bring in the 'fcs' and 'rcsi' measures from the API. This is an example of the API request / endpoint: https://api.hungermapdata.org/v1/foodsecurity/country/NAM/region?date_start=2021-10-01

See the related issue https://github.com/WFP-VAM/prism-frontend/issues/277 which will get a unique boundary file per admin level layer. This integration will also require a unique boundary file which comes from the GeoEnabler.

wadhwamatic:
Adding more details here following a call with HMU team today. 
1. We'll need to get the boundary file and host it directly. Disregard earlier notes about GeoEnabler. Awaiting boundary data / polygons instead
2. End date can be added to the API request. For example, this request will get 5 days of data. https://api.hungermapdata.org/v1/foodsecurity/country/NAM/region?date_start=2021-10-01&date_end=2021-10-05
3. We're still limited to the 100 most recent days of data. There's no endpoint to query available dates. But we can assume there's daily data starting 100 days ago and until today. 
4. There are multiple indicators in the endpoint. We would create a separate admin_level_data layer for each indicator by only need to get the response from the API once

Anything I missed @erwagasore and @JorgeMartinezG ?

wadhwamatic:
@erwagasore - I've added IDs which match those used by the HungerMap API to a boundary json and placed it here (WFP internal at the moment): https://wfp.sharepoint.com/:u:/s/PRISM/ESApcqY4oq9NmKMMEvgLKEUBfBY0m7vIUsp-FuX2qMJxcg?e=cAyXgr 

The field 'mvam-Code' matches with the region id used in the API as seen here: https://api.hungermapdata.org/v1/foodsecurity/country/MOZ/region?date_start=2021-12-01&date_end=2021-12-08 

[...]

PR #338, opened 2021-12-08
Feat/layers can specify a different boundary layer

erwagasore:
Closes: #321
Deploy: https://unique-boundary-layer.surge.sh

TODO:
- [x] Allow Admin Level Data layers to use unique boundary layers
- [x] Allow prism to include multiple boundary layers and only allow few to be displayed

How-to Configure:
- In `layers.json` all layers of type `boundary` must be present into this file
- In `layers.json` optionally you can specify `isPrimary` into one layer of type`boundary`.
- In `prism.json` add `adminBoundaryDisplayOverrides` as an array for boundaries to display.

Features:
- Add boundary layer visibility flag inside of layer.json
- Add boundary layer primary flag inside of layer.json
- Add adminBoundaryDisplayOverrides as a list of boundary layers instead of using multiple flags
- Add boundary property within Admin Level Data layer to signal which boundary layer to use
- Add getDisplayBoundaryLayers utility function to act in place of getBoundaryLayers used previously
- Improved getBoundaryLayerSingleton to get the first boundary in layers returned by getDisplayBoundaryLayers

image

image


src/config/types.ts
```diff
@@ -219,10 +219,13 @@ type LayerStyleProps = {
 export class BoundaryLayerProps extends CommonLayerProps {
   type: 'boundary';
   path: string; // path to admin_boundries.json file - web or local.
+  adminLevel: number;
   adminCode: string;
   adminLevelNames: string[]; // Ordered (Admin1, Admin2, ...)
   adminLevelLocalNames: string[]; // Same as above, local to country
   styles: LayerStyleProps; // Mapbox line and fill properties.
+  @optional
```
ericboucher, line 227: do we really need this?

erwagasore: Yes. This is to align with `RBD` request to have unique `boundaries` per `admin_level_data` layer. The scenario is that by default if `visibility` is not set we imply it as `true` by default to avoid forcing it on all boundaries in all countries. There are cases where we may have a non-default `boundary` which will be visible for a specific `admin_level_data` layer thus such boundary will have `"visibility": false`

erwagasore: e.g.
```json
{

Comment (ericboucher) [reply src/config/types.ts:227]: why not just assume that you want it visible when you are loading the `x_admin_level_data` that requests it, but just use the default admin_boundary otherwise?

File: src/components/MapView/Layers/AdminLevelDataLayer/index.tsx [hunk 2]
```diff
@@ -61,6 +65,16 @@ function AdminLevelDataLayers({ layer }: { layer: AdminLevelDataLayerProps }) {
...
 
+  if (
+    layer.boundary &&
+    !map
+      ?.getStyle()
+      .layers?.map(l => l.id)
+      .includes(`${boundaryId}-line`)
+  ) {
+    return null;
+  }
```
ericboucher, line 76: I would remove the `layer.boundary &&` condition, since you also want this check to work when you switch back to a normal layer which uses the default admin boundary.

If this returns null, will the data load eventually?

src/components/MapView/Layers/AdminLevelDataLayer/index.tsx
```diff
@@ -61,6 +65,16 @@ function AdminLevelDataLayers({ layer }: { layer: AdminLevelDataLayerProps }) {
     return null;
   }
 
+  if (
+    layer.boundary &&
+    !map
```
ericboucher, line 70: it might be worth creating a function `isLayerOnView` to do this type of check

src/components/MapView/Layers/AdminLevelDataLayer/index.tsx
erwagasore: Yeah. The data loads eventually. The case for `layer.boundary` is to only do this check for layers with defined `boundary`. Without it normal layers would return null which is the reverse of what we want.

src/config/utils.ts
```diff
@@ -144,18 +141,22 @@ export function getBoundaryLayers(): BoundaryLayerProps[] {
   return boundaryLayers;
 }
 
-export function getBoundaryLayerSingleton(): BoundaryLayerProps {
-  const boundaryLayers = getBoundaryLayers().filter(
+export function getVisibleBoundaryLayers(): BoundaryLayerProps[] {
+  return getBoundaryLayers().filter(
     l => l.visibility === true || l.visibility === undefined,
```
ericboucher, line 146: visibility doesn't exist anymore, here, does it?

src/components/MapView/Layers/AdminLevelDataLayer/index.tsx
erwagasore: You were correct. It is no longer needed since we will always have `boundaryID`

wadhwamatic:
@erwagasore - double loading of the boundary layer is still pretty pronounced on my side. Hoping during a code review either @ericboucher or @Ry-DS can give some inputs on how to address this. If it's going to be a major task, we can create a separate issue as discussed

erwagasore:
@ericboucher  and @wadhwamatic solved the issue of `double` loading layer. Please, check it out

wadhwamatic:
@erwagasore - this looks good. I added Mozambique config as well. One thing I noticed though is that the 'run analysis' feature doesn't utilize the boundary file associated with the admin level data config. I.e., if you 'run analysis' on a layer with admin1 level data, the result still uses admin2. See screenshots below. I'll create a separate issue for this.

admin1 level data correctly displayed
Screen Shot 2021-12-17 at 10 37 12

incorrectly displays admin2 boundary when 'run analysis' is executed using admin1 level data
Screen Shot 2021-12-17 at 10 38 08

wadhwamatic:
@ericboucher and @erwagasore - I'm hoping you two can get on a brief call soon so Eric can share some feedback and get a better understanding of what we're trying to do.

Quick summary from my side on the intention:
1. There's now a concept of multiple boundary files. A default boundary layer would work the same as currently. But some admin level layers would use an alternative boundary file. This is because either the boundary has changed over time and the default doesn't apply to the dataset, or for future cases where we would use boundaries for non-administrative cases such as agro-ecological or livelihood zones
2. We would use the default boundary file on initial load and likely for all WMS and point data layers. For admin level layers, which boundary file could change depending on the dataset. 
3. This also gives us the ability to show admin level data at the appropriate 'level'. I.e. for some cases, the admin level data is at the province level (admin 1), and others at district (admin 2). We're not handling that at the moment and instead just applying the same value for each district within a province, which isn't correct

Hope that helps clarify things for your discussion. Please post any specific questions here if you have them

src/config/mozambique/layers.json
```diff
@@ -21,9 +22,9 @@
     "type": "boundary",
     "path": "../data/mozambique/moz_admin1_boundaries.json",
     "opacity": 0.8,
-    "admin_code": "ADM2_PCODE",
-    "admin_level_names": ["ADM1_PT", "ADM2_PT"],
-    "admin_level_local_names": ["ADM1_PT", "ADM2_PT"],
+    "admin_code": "ADM1_PCODE",
```
JorgeMartinezG, line 25: Is ok to display only admin1 level data? 

image

src/utils/map-utils.ts
```diff
@@ -1,7 +1,7 @@
 import { Map as MapBoxMap } from 'mapbox-gl';
 import { LayerKey } from '../config/types';
 
-export function isMapBoundaryAvailable(
+export function isLayerOnView(
   map: MapBoxMap | undefined,
   boundaryId: LayerKey,
```
ericboucher, line 6: Let's either augment this function to catch other layers and rename boundaryId to layerId, not just lines

Or add a comment explaining that this only works for boundaries at the moment. Indeed, we are currently checking for `${boundaryId}-line`

src/config/mozambique/layers.json
erwagasore: @JorgeMartinezG my mistake. Correcting it.

src/utils/map-utils.ts
wadhwamatic: @erwagasore - has eric's comment above been addressed?

src/config/mozambique/layers.json
```diff
@@ -2,6 +2,7 @@
   "admin_boundaries": {
     "type": "boundary",
     "path": "../data/mozambique/admin_boundaries_MOZ2.json",
+    "primary": true,
```
ericboucher, line 5: is this needed?

src/config/mozambique/layers.json
ericboucher: why do we have changes in this config file?

src/config/types.ts
```diff
@@ -223,6 +223,10 @@ export class BoundaryLayerProps extends CommonLayerProps {
...
+  primary?: boolean; // Flag for general default layer a replacement for `admin_boundaries`
```
ericboucher, line 227: Do we need this new flag? Why not just keep `admin_boundaries` or maybe change it to `default_admin_boundaries` if you prefer?

src/utils/map-utils.ts
erwagasore: Yes.

src/config/types.ts
ericboucher: This way we don't need to add any new flag

src/components/NavBar/MenuSwitch/index.tsx
```diff
@@ -35,10 +46,29 @@ function MenuSwitch({ classes, title, layers, tables }: MenuSwitchProps) {
 
     if (checked) {
       updateHistory(urlLayerKey, layer.id);
+      const primary = getBoundaryLayerSingleton();
+      if (!('boundary' in layer)) {
+        if (!isLayerOnView(map, primary.id)) {
```
ericboucher, line 51: you can do that in one line I think

ericboucher:
@erwagasore much better. Just be careful, you sometimes get errors such as `Error: Layer with id "admin_boundaries-line" does not exist on this map` that kill the app.

Eg. steps:
- turn on admin test data
- turn on storms (this will load the default boundaries)
- turn off admin test data
-> crash

@wadhwamatic question for you, what is the expected behavior when you activate "admin test data" and then another layer that relies on a different boundary?

wadhwamatic:
> @wadhwamatic question for you, what is the expected behavior when you activate "admin test data" and then another layer that relies on a different boundary?

@ericboucher - it should switch to the boundary of the last layer that was selected. this will require that we continue to only allow one type of layer to be displayed at a time

wadhwamatic:
> @erwagasore much better. Just be careful, you sometimes get errors such as `Error: Layer with id "admin_boundaries-line" does not exist on this map` that kill the app.

@erwagasore - I'm seeing the same issue, which consistently fails to load a WMS layer

erwagasore:
> > @wadhwamatic question for you, what is the expected behavior when you activate "admin test data" and then another layer that relies on a different boundary?
> 
> @ericboucher - I guess you are specifically asking about a WMS layer? The WMS layer shouldn't update the boundary; we should maintain the 'admin test data' boundary

@ericboucher and @wadhwamatic: If we load a WMS layers which depends on a boundary which is not present we use the `getBoundaryLayerSingleton` function to get the default boundary layer associated with 'WMS'.

1. `Admin Level Data` Layer with unique boundary
image

2. `WMS` and `Admin Level Data` Layer
image

package.json
```diff
@@ -9,9 +9,7 @@
     }
   },
   "lint-staged": {
-    "src/**/*.{js,jsx,ts,tsx}": [
```
ericboucher, line 12: not sure why changes to this file happened?

src/config/utils.ts
```diff
@@ -133,25 +133,57 @@ export function getBoundaryLayers(): BoundaryLayerProps[] {
...
-  if (!boundaryLayer) {
+  if (boundariesCount === 0) {
     throw new Error(
-      'No admin_boundaries Layer found! There should be exactly one boundary layer defined in layers.json with id admin_boundaries.',
+      'No boundary layer found. There should be at least one boundary layer defined in layers.json',
     );
   }
 
-  return boundaryLayer;
+  if (boundariesCount > 1) {
```
ericboucher, line 150: Add comments explaining what we are doing in this section

src/components/NavBar/MenuSwitch/index.tsx
```diff
@@ -11,34 +11,80 @@ import {
...
+        if (Object.keys(LayerDefinitions).includes(boundaryId)) {
+          const displayBoundaryLayers = getDisplayBoundaryLayers();
+          const uniqueBoundaryLayer = LayerDefinitions[boundaryId as LayerKey];
+
+          if (
+            !displayBoundaryLayers
+              .map(l => l.id)
+              .includes(uniqueBoundaryLayer.id)
+          ) {
+            dispatch(removeLayer(uniqueBoundaryLayer));
```
ericboucher, line 78: what happens if we switched on two layers which share the same uniqueBoundary?

src/context/layers/admin_level_data.ts
```diff
@@ -23,10 +27,13 @@ export const fetchAdminLevelDataLayerData: LazyLoader<AdminLevelDataLayerProps>
   { layer }: LayerDataParams<AdminLevelDataLayerProps>,
   api: ThunkApi,
 ) => {
-  const { path, adminCode, dataField, featureInfoProps } = layer;
+  const { path, adminCode, dataField, featureInfoProps, boundary } = layer;
   const { getState } = api;
 
-  const adminBoundaryLayer = getBoundaryLayerSingleton();
+  const adminBoundaryLayer =
```
ericboucher, line 33: Add a comment explaining what is happening here

src/utils/map-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+import { getBoundaryLayers } from '../config/utils';
+
+export function isLayerOnView(map: MapBoxMap | undefined, layerId: LayerKey) {
+  return map
+    ?.getStyle()
+    .layers?.map(l => l.source)
+    .includes(`layer-${layerId}`);
+}
+
+export function boundaryOnView(map: MapBoxMap | undefined): BoundaryLayerProps {
```
ericboucher, line 12: Add a docstring for this function.
Also, shouldn't this be named: `boundariesOnView` instead?

src/config/mozambique/layers.json
ericboucher: @erwagasore it's still happening. But if you toggle the admin level 1 test data and then back, it works as expected.
So maybe something to look at there.

src/config/utils.ts
```diff
@@ -133,25 +133,57 @@ export function getBoundaryLayers(): BoundaryLayerProps[] {
...
+    }
+
+    return adminOverrideLayers;
+  }
+
+  return boundaryLayers;
+}
+
+export function getBoundaryLayerSingleton(): BoundaryLayerProps {
+  return getDisplayBoundaryLayers()[0];
```
ericboucher, line 186: I don't think that's actually what we want though, no? Don't we want the default admin boundary here? not the for one of the displayed boundaries?

src/components/NavBar/MenuSwitch/index.tsx
ericboucher: This will not happen with the current limitations of our app but might happen in the future.

Overall, I think it might be worth creating two util functions:
- `safeDispatchAddLayer`, that checks that the layer is not already loaded
- `safeDispatchRemoveLayer`, that checks that the layer is actually loaded before removing it

Or update addLayer / removeLayer to make them safer if that's worth it

ericboucher:
@erwagasore nice work, we're getting there. A few small things to clean up and make the code more future-proof and we should be gtg.

package.json
erwagasore: My editor decided to format it that way. Going to return to normal.

src/config/utils.ts
erwagasore: I added some comments within the function to clarify a little bit.

src/utils/map-utils.ts
erwagasore: Yeah. Better have `boundariesOnView` instead of picking just one. Especially that this util function will definitely be used somewhere else. Thus, just getting one was limiting

erwagasore:
@ericboucher when free take a look at new changes

src/components/MapView/Layers/AdminLevelDataLayer/index.tsx
```diff
@@ -3,35 +3,81 @@ import { useSelector, useDispatch } from 'react-redux';
...
+    ] as BoundaryLayerProps;
+
+    if ('boundary' in layer) {
+      if (Object.keys(LayerDefinitions).includes(boundaryId)) {
+        boundaryLayers.map(l => dispatch(removeLayer(l)));
+        dispatch(addLayer(boundaryLayer));
+
+        // load unique boundary only once
+        // to avoid double loading which proven to be performance issue
+        if (!isLayerOnView(map, boundaryId)) {
```
ericboucher, line 56: Now you can use `safeDispatchAddLayer` here I think

src/config/utils.ts
```diff
@@ -133,25 +133,60 @@ export function getBoundaryLayers(): BoundaryLayerProps[] {
...
+      id => !boundaryLayers.map(l => l.id).includes(id),
+    );
+
+    if (invalidOverrides.length > 0) {
+      throw new Error(
+        'Some of `adminBoundaryDisplayOverrides` layer Ids are not valid. You must provide valid ids from `layers.json`',
+      );
+    }
+
+    // get override layers from override names without
```
ericboucher, line 169: This is actually not what we want I think.
We need to differentiate between the displayOverride (for styling) and the boundary layer that we use to get admin boundary names by default.

By construction, the union of the displayOverrides should be the same as the defaultAdminBoundary. But this is not enforced, so it is important to treat them as two different entities.

src/config/utils.ts
ericboucher: See my comment above, you actually want to implement a different function, which takes the default admin boundary. Or what you called `primary` above.

@JorgeMartinezG what do you think? See https://github.com/WFP-VAM/prism-frontend/pull/338/files#r782654969 as well for context

public/data/myanmar/nso/poverty_st.json
```diff
@@ -0,0 +1,64 @@
+{
```
ericboucher, line 1: @wadhwamatic do we have a real example that we could add to the app instead of a test dataset?

wadhwamatic: @ericboucher - the short answer is we don't have one yet. but as we're integrating data from another source (mVAM) in the near term which may not have matching boundary files, it's eminent. see my comment on this issue: https://github.com/WFP-VAM/prism-frontend/issues/311#issuecomment-989330140

we also will have unique boundary files per round of data in the 'cadre harmonise' datasets for the upcoming regional deployment for west africa. not easy to see distinction in these two examples, but we know that the boundaries will change in some rounds of data collection:
https://github.com/WFP-VAM/RBD_CHIPC_dataprocessing/blob/main/data/geo/finalized_CHgeofiles/wca_CHIPC_nov2021_current_nov2021_simple.geojson
https://github.com/WFP-VAM/RBD_CHIPC_dataprocessing/blob/main/data/geo/finalized_CHgeofiles/wca_CHIPC_nov2021_projected_jun2022_simple.geojson

src/components/MapView/Layers/AnalysisLayer/index.tsx
```diff
@@ -46,10 +47,11 @@ function AnalysisLayer() {
     analysisData instanceof ExposedPopulationResult
       ? (analysisData.statistic as string)
       : 'data';
+  const boundaryId = getBoundaryLayerSingleton().id;
```
ericboucher, line 50: If the active layer has a unique boundary, `getBoundaryLayerSingleton` should return that unique boundary.

src/components/NavBar/MenuSwitch/index.tsx
```diff
@@ -11,34 +12,94 @@ import {
...
+  };
+
+  const safeDispatchRemoveLayer = (
+    _map: MapBoxMap | undefined,
+    layer: LayerType,
+  ) => {
+    if (isLayerOnView(_map, layer.id)) {
+      dispatch(removeLayer(layer));
+    }
+  };
```
ericboucher, line 56: can we move this to layer-utils?

src/components/NavBar/MenuSwitch/index.tsx
```diff
@@ -11,34 +12,94 @@ import {
...
+            !displayBoundaryLayers
+              .map(l => l.id)
+              .includes(uniqueBoundaryLayer.id)
+          ) {
+            safeDispatchRemoveLayer(map, uniqueBoundaryLayer);
+          }
+
+          displayBoundaryLayers.forEach(l => {
+            safeDispatchAddLayer(map, l);
+          });
```
ericboucher, line 100: I am not sure I understand what we are trying to achieve here? Can you explicit and/or simplify the logic?

src/components/MapView/Layers/AnalysisLayer/index.tsx
ericboucher: @wadhwamatic can you confirm that we want the analysis to use the "unique boundary" when mixing a raster and admin level data with a unique boundary?

wadhwamatic: @ericboucher - yes. whichever boundary associated with the 'admin level data' layer should be used for the analysis. I created an issue for this a few weeks ago: https://github.com/WFP-VAM/prism-frontend/issues/341

src/components/NavBar/MenuSwitch/index.tsx
erwagasore: Done.

erwagasore:
@ericboucher and @wadhwamatic please take a look at the latest changes.
Now the `analysis` layer respects `unique boundary layer` if available.

- https://moz-unique-boundary-layer.surge.sh showcasing`analysis`
- https://unique-boundary-layer.surge.sh is updates as well.

wadhwamatic:
Good progress @erwagasore. A few things to follow up on:
1. As discussed, let's remove admin level data layers that were previously in view when we display the 'run analysis' results. Similarly, if we activate an admin level data layer _after_ 'run analysis' we should hide the 'run analysis' results
2. Please look into the issue we saw when running an analysis on admin1 data in Myanmar. There are two regions (Bago West and Bago East) that have 'no data' in the result. They also are missing the local name. Maybe there's an issue with the boundary file?
3. When I tested the 'rbd' deployment and run an analysis, the results in the table are by admin0 instead of admin1 as expected. Please look into why that is happening
4. I don't see how we specify which boundary layer to use by default ('primary' layer). Are we dependent on naming a layer 'admin_boundaries'? I'd like to avoid that if we can. cc: @ericboucher

wadhwamatic:
@erwagasore - two things:
1. We need to update the readme file to make it clear that if you have more than one boundary file, you must add a configuration in prism.json with adminBoundaryDisplayOverrides
2. As mentioned, I'd like to change the term 'adminBoundaryDisplayOverrides' to 'defaultDisplayBoundaries' in the configuration of prism.json. The former name is unclear. Is this a simple change?

I've added changes to the Mongolia deployment to this branch as they have a a mix of admin 1 and admin 2 datasets

wadhwamatic:
> * As discussed, let's remove admin level data layers that were previously in view when we display the 'run analysis' results. Similarly, if we activate an admin level data layer _after_ 'run analysis' we should hide the 'run analysis' results

@erwagasore - it looks like we didn't implement hiding the 'run analysis' result when we activate a new admin level layer. Let's leave it like that for now though as we need to think about expected behavior of the results layer after a user starts doing other things; the same issue exists for WMS layers. I've captured this here: https://github.com/WFP-VAM/prism-frontend/issues/364

wadhwamatic:
@erwagasore - I tried to run an exposure analysis for Myanmar and got an error:
Unhandled Rejection (TypeError): Cannot read properties of undefined (reading 'boundary')

I thought it might relate to a missing 'isPrimary' layer but adding that didn't make a difference. I also tried an exposure analysis for Tajikistan, which for the moment just has 1 admin boundary layer but received the same error.

cc: @ericboucher

erwagasore:
> > * As discussed, let's remove admin level data layers that were previously in view when we display the 'run analysis' results. Similarly, if we activate an admin level data layer _after_ 'run analysis' we should hide the 'run analysis' results
> 
> @erwagasore - it looks like we didn't implement hiding the 'run analysis' result when we activate a new admin level layer. Let's leave it like that for now though as we need to think about expected behavior of the results layer after a user starts doing other things; the same issue exists for WMS layers. I've captured this here: #364

From a user prespective when you "Run Analysis" there is a need to compare the analysis layer with other layer that is why I think there is value to leave the `analysis layer` active. On the other end. We need to keep the application behaviour consistant. Lets continue the discussion on the new issue #364 on what we should do.

wadhwamatic:
> From a user prespective when you "Run Analysis" there is a need to compare the analysis layer with other layer that is why I think there is value to leave the `analysis layer` active. On the other end. We need to keep the application behaviour consistant. Lets continue the discussion on the new issue #364 on what we should do.

The comparison would be for a hazard layer as the vulnerability layers are simply being filtered. Anyway, we'll cover that separately as we have already discussed in issue 364. 

The exposure analysis is fixed, though another issue still needs to be addressed separately. Issue https://github.com/WFP-VAM/prism-frontend/issues/367 assigned to @JorgeMartinezG 

Let's merge this PR after @ericboucher reviews and approves.

ericboucher (approved):

PR #340, opened 2021-12-17
Trending data feature

### LINKED ISSUE #328: Add the capability to display a dataset with trend data when a user clicks on an admin area
This issue is meant to initiate the steps needed to visualize time-series data associated with an admin area for a selected indicator. For example, if a user has activated the 1-month rainfall anomaly (a WMS layer), and then they click on an admin area in the map, we want to be able to display the average 1-month rainfall anomaly for that place over time. 

The display would work similarly to what we see in the [DataViz platform](https://dataviz.vam.wfp.org/seasonal_explorer/rainfall_vegetation/visualizations#):
Mozambique_Gaza_2021_RainfallAnomaly

We could use the data drawer (same area used for tables.json display and the exposure analysis results), or opt for a pop-up on click.

See quick mock ups below:
1. Pop up
timeseries-popup

2. Data drawer
timeseries-datadrawer

We don't yet have a way to query this data from ODC. However, we will receive it as a static dataset (as a CSV or json) that we can then add to PRISM in the same format as tables or an admin_level_data layer. The idea would then be to add a config parameter in the WMS layer that notes that the table or admin_level_data layer should be displayed when a user clicks on an admin area.

JorgeMartinezG:
This PR closes #328. On a boundary click event, PRISM will display information-related to the sample csv file.

**TODO:**
- Set trigger. A button rendered inside the popup could be a solution. Discussion needed.
- Additional metadata: Dialog box title, plot title, etc.
- Boundary layer configuration.

record


wadhwamatic:
Good progress @JorgeMartinezG. re: trigger, we might consider modifying the general behavior so that the admin name shows on hover, and additional data shows on click. For WMS layers, the onclick would load the data view if available, for other layers, it would maintain the current behavior.

Let's discuss on Monday

JorgeMartinezG:
updated branch. Fixed conflicts with master and applied hover to display popup.

You can find deployment instance here: http://prism-hover.surge.sh/

wadhwamatic:
Great work @JorgeMartinezG. For the metadata to include in this example:
- Dialog box title: Data Viewer: $region-name; Great if we can concatenate admin 1, admin 2 (same as we do for the hover action on the map)  
- Plot title: 1-month rainfall aggregate (mm)

Some additional suggestions (nice to have for v1):
1. Remove admin code from the label that shows up when you hover on the chart 
2. Reduce precision of rainfall values to two decimal points (this should be a config parameter)
3. Add unit after the value that shows up on hover over the chart. In this example it would be 'mm'
4. Add a caption (below chart): CHIRPS 1 month dekadal rainfall aggregations (mm)
5. Add x axis lines 
6. Remove 'district' label on the y axis

erwagasore:
After a discussion with @wadhwamatic on the potential of showing the `Chart` but also allow the user to continue interacting with the `map` below is the design proposal on how we can achieve it.

Keep in mind that this is the not the prefect solution, this is a short term solution which doesn't involve many design changes to the current state of the UI/UX of PRISM. Please share you opinion on this as well.

Show-chart-and-map

wadhwamatic:
> After a discussion with @wadhwamatic on the potential of showing the `Chart` but also allow the user to continue interacting with the `map` below is the design proposal on how we can achieve it.
> 
> Keep in mind that this is the not the prefect solution, this is a short term solution which doesn't involve many design changes to the current state of the UI/UX of PRISM. Please share you opinion on this as well.
> 
> Show-chart-and-map

This looks fine and covers what we've discussed. Please go ahead and implement as part of this PR. Keep @JorgeMartinezG  in the loop while you develop the component

wadhwamatic:
This feature has been committed to master via https://github.com/WFP-VAM/prism-frontend/pull/453

PR #329, opened 2021-12-03
Move map on boundary selected

JorgeMartinezG:
This PR closes #292 .

The changes include a **Go to** button which displays a dropdown menu with the list of boundary names at level 1 and 2. On a click event. the map is centered to the selected boundary

record

PR #347, opened 2022-01-07
Center map on selected admin boundary

JorgeMartinezG:
This PR is an enhancement of #329, which reuses select menu developed on #308.

image


src/components/MapView/Layers/BoundaryDropdown.tsx
```diff
@@ -37,12 +48,31 @@ const ClickableListSubheader = styled(ListSubheader)(({ theme }) => ({
...
     '&>div': {
       color: 'black',
     },
   },
+  dropdownMenu: {
+    backgroundColor: theme.palette.primary.main,
+    color: 'white',
+    display: 'inline-flex',
+    marginTop: '1em',
+    padding: '6px 16px',
```
Ry-DS, line 62: try use `theme.spacing(1,2)` when possible

src/components/MapView/Layers/BoundaryDropdown.tsx
```diff
@@ -13,19 +13,30 @@ import {
   Theme,
```
Ry-DS, line 13: We have BoundaryDropdown and LayerDropdown both in the `Layers` folder which doesn't make too much sense. Think we can move it out to some place better?

src/components/MapView/Layers/BoundaryDropdown.tsx
```diff
@@ -299,8 +333,91 @@ function BoundaryDropdown({
...
+    root: {
+      '& label': {
+        textTransform: 'uppercase',
+        letterSpacing: '3px',
+        fontSize: '11px',
+      },
+      '& svg': { color: 'white', fontSize: '1.25rem' },
+    },
+  };
+})(SimpleBoundaryDropdown);
```
Ry-DS, line 352:
```suggestion
export const SelectMenu = withStyles(() => ({
    root: {
      '& label': {
        textTransform: 'uppercase',
        letterSpacing: '3px',
        fontSize: '11px',
      },
      '& svg': { color: 'white', fontSize: '1.25rem' },
    }
  }))(SimpleBoundaryDropdown);
```
May not be syntactically perfect, but hope you're getting the idea ;)

src/components/MapView/index.tsx
```diff
@@ -441,6 +442,7 @@ function MapView({ classes }: MapViewProps) {
       >
         <Grid item>
           <Analyser extent={adminBoundariesExtent} />
+          <BoundaryDropdownMap />
```
Ry-DS, line 445: A better name may be `<GotoBoundaryDropdown/>`

src/components/MapView/Layers/BoundaryDropdown.tsx
```diff
@@ -299,8 +333,91 @@ function BoundaryDropdown({
       setSelectedBoundaries={newSelectedBoundaries => {
         dispatch(setSelectedBoundariesRedux(newSelectedBoundaries));
       }}
+      labelMessage={labelMessage}
     />
   );
 }
 
+export const SelectMenu = withStyles(() => {
```
Ry-DS, line 341: Is there a better name that summarizes what exactly you're changing here? Or a comment would help since this isn't too straight forward from an initial read.

src/components/MapView/Layers/BoundaryDropdown.tsx
```diff
@@ -299,8 +333,91 @@ function BoundaryDropdown({
...
+
+  return (
+    <div className={styles.dropdownMenu}>
+      <CenterFocusWeak fontSize="small" className={styles.icon} />
+      <SelectMenu
+        selectedBoundaries={selectedBoundaries}
+        labelMessage="Go to"
+        className={styles.formControl}
+        setSelectedBoundaries={newSelectedBoundaries => {
+          dispatch(setSelectedBoundariesRedux(newSelectedBoundaries));
```
Ry-DS, line 378: Is this the same redux used for selecting boundaries that I made? I think it should be fine to use it, just want to make sure it won't break anything if you're using both drop-downs simultaneously/cause highlights if you're using the alert window.

Ry-DS (commented):
Nice! Have a deploy link handy?

wadhwamatic:
Nice work on this @JorgeMartinezG. A few requests / comments:
1. Can we have the height of the 'Go to' button match the 'run analysis' button?
2. Though multi-select by default makes sense in the alerts module, I don't think it does in this case. Can we implement this so that when a user clicks on a new region, it chooses only that one?
3. Related to point 2, if a user wants to select more than one region, can we use shift+click to select another region?
4. Last - and this can be done later - can we include an expand / contract feature at the higher admin levels? i.e., just show the top level layers by default, but allow a user to expand level two, and three if needed. See the original issue on this for an example: https://github.com/WFP-VAM/prism-frontend/issues/226

src/components/MapView/Layers/BoundaryDropdown.tsx
JorgeMartinezG: included in cfa705fe46213f209d51a49cef06e81a6ad3a0c8

src/components/MapView/Layers/BoundaryDropdown.tsx
JorgeMartinezG: Included in d5d3c84a2aecd2a652255af256f06d124848e901

src/components/MapView/Layers/BoundaryDropdown.tsx
JorgeMartinezG: Included in a087321458770c698a09ed29d8823566cbaf9e01 and also in previous comment

src/components/MapView/Layers/BoundaryDropdown.tsx
JorgeMartinezG: I created an internal state here d5d3c84a2aecd2a652255af256f06d124848e901

JorgeMartinezG:
- @Ry-DS there is a deployed version of this branch here: https://prism-goto.surge.sh/
- @wadhwamatic Items from 1-3 have been resolved within this branch. Please check link above

wadhwamatic:
Works great @JorgeMartinezG. I just have the same feedback as Eric regarding select all / de-select all

JorgeMartinezG:
> Works great, small feedback @JorgeMartinezG:
> 
> * we can probably remove the "select all" option in this case
> * when clicking on "deselect all", show "GO TO" again instead of a blank line

Fix for first item is in 01a3475f9e55089893d7762dd8d52b82f1612832. WRT second item, the Go To label disappears on focus event. If we keep the label in the same position MUI will display the label and current value of the Select component at the same time. 

image

Tried multiple options with css but no success yet.

wadhwamatic (dismissed):
looks good @JorgeMartinezG! I added a small commit for the RBD config as well. @ericboucher - any final input?

ericboucher:
As mentioned yesterday I think we need to wait on #338 to be merged first :)

wadhwamatic:
> As mentioned yesterday I think we need to wait on #338 to be merged first :)

@JorgeMartinezG - PR 338 has been merged. Can you now revisit this?

JorgeMartinezG:
this is branch is ready. Also included optional flag within prism.json file. cc @wadhwamatic @ericboucher

wadhwamatic (approved):
Works as expected. @ericboucher - can you give a quick final review before we merge?

src/config/cambodia/prism.json
```diff
@@ -1,6 +1,7 @@
 {
   "country": "Cambodia",
   "alertFormActive": false,
+  "navigationDropdown": false,
```
ericboucher, line 4: @wadhwamatic should the default be false and we only activate it for specific countries? Or the opposite?
Not sure we need to set it explicitly in the config for each counyty. What do you think?

wadhwamatic: agreed with @ericboucher - let's remove the flag and make the default behavior false. we would specify true when required for a deployments. thanks for implementing and letting us know when ready @JorgeMartinezG

JorgeMartinezG: Fix included in 7eae209b4048ba6f4869f4732b13ff75ee0806de

JorgeMartinezG: @wadhwamatic fixed in ad4a5d5c780cbad40d4aa1fc5bd4f2cf09090997

src/components/MapView/Layers/BoundaryDropdown.tsx
```diff
@@ -369,9 +369,10 @@ export const GotoBoundaryDropdown = () => {
     | undefined;
   const { data } = boundaryLayerData || {};
 
+  const navigationDropdown = get(appConfig, 'navigationDropdown', false);
```
ericboucher, line 372: another option would be to extend the type of appConfig

ericboucher: or move this logic to `prism-frontend/src/config/index.ts` wdyt?

ericboucher (approved):

src/components/MapView/Layers/BoundaryDropdown.tsx
JorgeMartinezG: I moved it to the config index file here: fa5e2eb57e893887d3dec610d1b192fe1b797247

ericboucher:
Nice work @JorgeMartinezG, you can go ahead and merge.

Just being picky but you might want to rename the config to something like `enableNavigationDropdown`

PR #355, opened 2022-01-19
Add intersect option to the flask API

ericboucher:
When a user provides the parameter `intersect_comparison`, the API will calculate and return a new stat `intersect_percentage`.

The parameter accepts standard operators such as `<, >, <=, =, !=` in the form of `intersect_comparison=>=40`

### DIFF
--- a/api-flask/app/zonal_stats.py
+++ b/api-flask/app/zonal_stats.py
@@ -175,7 +175,7 @@ def calculate_stats(
     prefix='stats_',
     geojson_out=False,
     wfs_response=None,
-    intersect_threshold=None,
+    intersect_comparison=None,
 ):
     """Calculate stats."""
     if group_by:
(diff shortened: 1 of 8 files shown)


api-flask/app/zonal_stats.py
```diff
@@ -193,14 +194,27 @@ def calculate_stats(
         stats_input = [s.get('geom') for s in zones]
         prefix = None
 
+    # Add function to calculate overlap percentage.
+    add_stats = None
+    if intersect_threshold:
+        def percentage_over_threshold(masked):
```
JorgeMartinezG, line 200: IMO, we could refactor the function like this.
```python
        def percentage_over_threshold(masked):
            # Get total number of elements in the array.
            total = masked.size
            if total == 0:  # Avoid division by zero.
                return total

            # Count number of elements higher than threshold.
            over_threshold = (masked > intersect_threshold).sum()

            # Compute percentage over total.
            return over_threshold / total
```

ericboucher: I think size does not take into account the shape of the admin boundary and returns the size of the complete unmasked area, no?

ericboucher: https://numpy.org/doc/stable/reference/generated/numpy.ma.MaskedArray.size.html

ericboucher: @JorgeMartinezG please let me know what you think about the above.
And have a look at the new format which now accepts parameters of the form `>=10`

wadhwamatic (approved):
Works well from a functional perspective. @ericboucher and @JorgeMartinezG - are you in agreement on the implementation?

JorgeMartinezG (approved):

PR #362, opened 2022-01-24
Replace mapbox-gl with maplibre-gl

DanielJDufour:
This PR does two things: (1) gets maplibre-gl running and (2) removes the access token requirement.  Here's more details:

**in package.json**
- `"mapbox-gl": "npm:maplibre-gl@1.15.2",` basically installs maplibre-gl in the place of mapbox-gl, so the mapbox-gl folder in node_modules actually includes maplibre-gl
- `"**/mapbox-gl": "npm:maplibre-gl@1.15.2"` uses [yarn selective resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/), so that react-mapbox-gl looks for maplibre-gl in place of mapbox-gl
- "setup:maplibre" script basically renames the css file in maplibre-gl to the mapbox-gl name, because react-mapbox-gl looks for it

**in src/components/MapView/index.tsx**
usage of ``` process.env.REACT_APP_DEFAULT_STYLE``` makes it so you can specify a default style in your .env that is open and doesn't require an access token, such as `"https://maputnik.github.io/osm-liberty/style.json"`

**in yarn.lock**
the changes are what happened when I ran `yarn install` after the other changes.

Let me know if you have any questions.

** alternatives **
- I originally tried a full replacement of all references to mapbox-gl in the source code, but ran into a wall because react-mapbox-gl still need something named "mapbox-gl" installed.

Open to your thoughts :-)

Also, here's a screenshot using `REACT_APP_DEFAULT_STYLE="https://maputnik.github.io/osm-liberty/style.json"`
Screen Shot 2022-01-23 at 8 12 28 PM


src/components/MapView/index.tsx
```diff
@@ -410,7 +410,10 @@ function MapView({ classes }: MapViewProps) {
       )}
       <MapboxMap
         // eslint-disable-next-line react/style-prop-object
-        style="mapbox://styles/eric-ovio/ckaoo00yp0woy1ipevzqnvwzi"
+        style={
+          process.env.REACT_APP_DEFAULT_STYLE ||
+          'mapbox://styles/eric-ovio/ckaoo00yp0woy1ipevzqnvwzi'
```
ericboucher, line 415: does this style work with maplibre?

ericboucher: If we are switching completely to maplibre, it might be worth creating/finding a similar style to `mapbox://styles/eric-ovio/ckaoo00yp0woy1ipevzqnvwzi` that is not hosted on mapbox

package.json
```diff
@@ -24,7 +24,9 @@
     "deploy:staging": "firebase deploy --only hosting:staging-target",
     "eject": "react-scripts eject",
     "lint": "eslint --ext .js,.jsx,.ts,.tsx ./src --ignore-path .gitignore",
-    "precommit": "lint-staged"
+    "postinstall": "npm run setup:maplibre",
```
ericboucher, line 27: the default manager for this project is "yarn" so let's use that here as well :)

ericboucher (commented):
@DanielJDufour thanks for your contribution! It looks great. Let's make sure our defaults are compatible with maplibre and this should be gtg.

package.json
DanielJDufour: Good catch!

DanielJDufour: fixed

DanielJDufour:
Just recapping action items:
- @ericboucher , could you download a zip of the default style from Mapbox (mapbox://styles/eric-ovio/ckaoo00yp0woy1ipevzqnvwzi)?  And then email it to me at daniel.j.dufour@gmail.com?
- @DanielJDufour will set up a personal Maptiler acct to test with and see if we can easily transfer the style and report back within a few days

src/components/MapView/index.tsx
DanielJDufour: I updated this to use the temporary Maptiler style on my account.

DanielJDufour:
There appears to be some sort of issue with marked import.  Probably something I introduced when I tried to resolve merge conflicts.  I'll fix it:

Screen Shot 2022-02-02 at 11 05 54 AM

DanielJDufour:
@ericboucher , I think is ready for re-review now.

**Backwards Compatability**
You should still be able to use MapboxGL JS with this new code.  This is meant to help ease the migration process and to prevent breaking any instances running in production using the old set of environmental variables.  However, I haven't tested this as I don't have a Mapbox account.

**unexpected issue importing marked**
The issue with importing `marked` above went away when I did ` yarn install --frozen-lockfile`.  I think there might be some issue with dependency versions when one does yarn install.  fwiw, I checked the documentation and the code importing marked is correct according to the documentation.

**AnyLayer type**
I also ran into a weird typing issue when iterating through map layers and looking for a 'source' property.  It should be resolved as well.

Happy to answer any questions about the PR!

DanielJDufour:
Screen Shot 2022-02-02 at 4 31 16 PM
Screen Shot 2022-02-02 at 4 31 29 PM

ericboucher (approved):

wadhwamatic:
@ericboucher & @DanielJDufour - can you provide a couple of things related to this PR:
1. For WFP-owned instances of PRISM (i.e., not being handed over to government), we're interested in using Mapbox to benefit from newer features. What is the process for choosing between Mapbox and MapLibre?
2. Is the creation of basemaps using https://cloud.maptiler.com/maps/ completely free? For PRISM deployments we handover to governments, will creation of a Maptiler account be a requirement?

ericboucher:
1. I think we will have to proceed very carefully, if we start adding new features that are only available to Mapbox users but not maplibre users, it might get a bit out of hands. Do you have examples of such features?
2. Yes, creating a map is free and fairly easy

PR #370, opened 2022-01-27
Login using WFP account

### LINKED ISSUE #312: Implement OpenID connect for WFP Myanmar deployment
We have a temporary authentication in place for PRISM Myanmar which is implemented as a Lambda function called through CloudFront. The process is essentially this one: https://kynatro.com/blog/2018/01/03/a-step-by-step-guide-to-creating-a-password-protected-s3-bucket/ 

Our TEC colleagues have flagged this approach as insecure. We instead need to implement OpenID authentication. See info here:
https://newgo.wfp.org/how-do-i/integrate-openid-connect 

This is high priority as we need to address it quickly. For discussion in our next meeting @JorgeMartinezG .

JorgeMartinezG:
closes #312 

image


wadhwamatic:
This worked as expected for me, except it seems to bypass MFA? Here's what I experienced:
1. Logout of WFP via https://selfservice.go.wfp.org/
2. Start PRISM on localhost
3. Log in using WFP credentials
4. PRISM loads without MFA

I don't know if we're supposed to enforce MFA but did expect it to follow the same flow as any other WFP application which does usually require MFA. Can you check on this @JorgeMartinezG ?

One other minor comment - should we have the flag be more specific? Rather than 'authRequired' should it be 'WFPauthRequired'? I.e., is this implementation 100% specific to WFP, or would it also allow us to implement other authentication tools as well?

wadhwamatic:
@JorgeMartinezG - we'll need to add MFA. Please go ahead with that within this PR

wadhwamatic (approved):
Tested locally, works as expected. I modified the auth parameters in environment variables and deployed to https://prism.mmr.wfp.org/ to confirm it works as expected there. This is good to go

PR #371, opened 2022-01-28
Select latest admin level for exposed population analysis

### LINKED ISSUE #367: Exposure analysis uses the wrong order of admin levels
Exposure analysis treats the order of admin levels in the opposite direction of other features. For example in Myanmar, we have this admin layer configuration:
    "admin_level_names": ["ST", "DT", "TS"]

This lists the largest admin: admin1 (state), and then decreases to admin2 (district) and admin3 (township). However, the exposure analysis is grouping by admin1, rather than admin3. Exposure analysis should use the lowest level available and use the current order in the config (admin1, admin2, admin3).

JorgeMartinezG:
closes #367


ericboucher (approved):

wadhwamatic (approved):
Works as expected, thanks @JorgeMartinezG

Issue #351, opened 2022-01-14
Integrate EWS-1294 data into PRISM Cambodia


wadhwamatic:
EWS-1294 (http://ews1294.com) is a flood early warning system in Cambodia which monitors river levels using sensors across the country. Each sensors has a fixed location with multiple new readings every hour. 

There are two components to the API: locations, and data points. We need to be able to configure the EWS-1294 data as a 'point' layer which dynamically retrieves data through the API endpoint.

The descriptions below are from the EWS-1294 technical lead. 

**Locations API**
http://sms.ews1294.info/api/v1/location.geojson?type=river
In the objects, you can find the names (both English and Khmer), locations, trigger levels, status of each location we have (or plan to have) sensors installed (note that we have 2 status fields – `status` which is dynamically updated between active and inactive depending on whether the sensor is sending data or not, and ´status1´ which is our internally expected status, and can have the values of Test, Planned or Operational – For PRISM, I suggest you only show the ones marked as Operational). We expect the list of locations to reach about 50 before next rainy season.
 
**Datapoint API**
http://sms.ews1294.info/api/v1/sensors/sensor_event?location=22&start=2019-07-01&end=2019-07-19
Using the location id (from the location API), as well as a start and end date (in any format understood by PHP), you can get the water heights reported by the sensors in that period. Without start and end provided, the API will default to return only the last 6 hours. I would advise against loading more than one month at a time, as this might hit the server harder than needed 😊

wadhwamatic:
@erwagasore - as discussed today, let's approach this as two separate queries: 1) get data from all sensors on a single day, and 2) if a user clicks on a sensor location, query the past seven days of available data for that location (time interval TBC). 

For the first query, we'll need to summarize data. Let's start by computing the mean, median, min and max value for each sensor on a specific day. Once we have this task complete in a PR, we can start visiting the second query.

erwagasore:
@wadhwamatic I am experiencing issue with `ews-1294` API when called twice or three times in a short time it is crashing  with error e.g. 
```
error: Internal server error: 429 Client Error: Too Many Requests for url: http://sms.ews1294.info/api/v1/sensors/sensor_event?external_id=TEPv4.0-033&start=2022-01-28&end=2022-01-29
```
[...]

PR #381, opened 2022-02-14
Cambodia EWS1294 data with trend line graph

erwagasore:
Closes: #351
Deploy: ~https://cambodia-ews-data.surge.sh~

TODO:
- [x] Create `ews/data` endpoint in `api-flask` to process `locations` and `datapoint` APIs into PRISM ready data
- [x] Display `propinfo` on hover of `point_data`
- [x] Disable previous click event on `point_data` to only allow the dedicated click event for datapoints
- [x] Show  today's levels in a chart onclick

How-to Configure:
- [x] Run `api-flask` locally to expose `localhost/ews/data` endpoint
- [x] Configure `river level` details into `layer.json` and `prism.json` - Already done

Features:
Handle the logic of processing sensor locations and get censors data points
Cache the endpoint to avoid `internal server errors` specified [here](https://github.com/WFP-VAM/prism-frontend/issues/351#issuecomment-1024383188)

Screenshot 2022-02-14 at 12 40 02

### DIFF
--- a/api-flask/app/main.py
+++ b/api-flask/app/main.py
@@ -11,7 +11,7 @@ from app.errors import handle_error, make_json_error
 from app.timer import timed
 from app.zonal_stats import calculate_stats, get_wfs_response
 
-from ews import get_ews_location_response, get_ews_response, parse_ews_params
+from ews import get_ews_response, parse_ews_params
 
 from flask import Flask, Response, json, jsonify, request
 
@@ -152,14 +152,6 @@ def get_kobo_forms():
     form_responses = get_form_responses(begin_datetime, end_datetime)
 
     return Response(json.dumps(form_responses), mimetype='application/json')
-
-
-@app.route('/ews/location/data', methods=['GET'])
-def get_ews_location_data():
-    """Get datapoints for only one location."""
-    _, begin_datetime, end_datetime = parse_ews_params()
-    resp = get_ews_location_response(begin_datetime, end_datetime)
-    return Response(json.dumps(resp), mimetype='application/json')
 
 
 @app.route('/ews/data', methods=['GET'])
--- a/api-flask/app/main.py
+++ b/api-flask/app/main.py
@@ -11,6 +11,8 @@ from app.errors import handle_error, make_json_error
 from app.timer import timed
 from app.zonal_stats import calculate_stats, get_wfs_response
 
+from ews import get_ews_responses, parse_ews_params
+
 from flask import Flask, Response, json, jsonify, request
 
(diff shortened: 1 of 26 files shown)


wadhwamatic:
Great start on this @erwagasore. A few comments:
1. I don't see how to view historical river gauge data - only today shows up
2. As discussed today, let's add the chart title in the pop up window
3. I'm wondering if there's anything we can do to make it easier to get the hover details on the gauges. Right now, the admin hover action is what you see until you precisely hover over the gauge location

In one instance, the connection to the EWS-1294 API failed and produced a 502 error:
HTTPError: 502 Server Error: Bad Gateway for url: https://api.ews1294.com/v1/datapoints?external_id=TEPv4.0-013&start=2022-02-14&end=2022-02-15

We should handle a failed connection. When this happened, nothing was displayed on the front-end. No message and no data. If it fails to connect, we should leave the layer toggled off, and give a user-friendly message like 'Unable to connect to EWS-1294. Please try again'.  Then retry if the user toggles the layer on again.

api-flask/app/main.py
```diff
@@ -155,7 +155,6 @@ def get_kobo_forms():
 
 
 @app.route('/ews/data', methods=['GET'])
-@cache.cached(timeout=900, query_string=True)
```
wadhwamatic: @erwagasore - what is the implication of removing this?

wadhwamatic:
@erwagasore - I'm seeing a number of issues when I start digging into the details of the data that is shown when we click on an individual sensor:

1. There is not a new network request for the sensor data which seems odd. So I looked at the initial request when the river gauge data is first requested. It seems we're getting a large amount of data on that first request, and the structure of that response is less than optimal. I.e., the response from this query http://localhost/ews/data?beginDateTime=2021-11-11&endDateTime=2021-11-11& produces the attached. Can you explain this approach? Is this to deal with the EWS API failing if too many requests are sent? 
[data.json.txt](https://github.com/WFP-VAM/prism-frontend/files/8101196/data.json.txt)
2. The date formatting that was added doesn't work. See screenshot below which shows a date as 11/57/2021 (along with several other erroneous dates)
3. The water height in the tooltip doesn't match the last value in the line chart which doesn't make sense. In the same screenshot, see the water height value of -81 while the line chart shows the last value as -7074

Screen Shot 2022-02-18 at 17 39 10

erwagasore:
closed and moved all changes in branch [ews-modal](https://github.com/WFP-VAM/prism-frontend/tree/feature/ews-modal)

PR #244, opened 2021-08-26
🎉  Feature/layer config content

erwagasore:
- Enabled load `contents` from `layer` configuration file
- Added `info` button on the left side of `layer legend`
- Once clicked a dialog pop-up and show `markdown` or `html` contents within `.md` file

PR #382, opened 2022-02-15
Initialize multi-language support using i18n

ericboucher:
Implement internationalization using i18n.
Add a button for language selection

Screen Shot 2022-02-16 at 7 38 06 PM

Todo:
- [x] When a layer is activated, the content in the legend is not translated (name, legend text, presumably legend labels)
- [x] The layer names and their category headers are not translated in the 'run analysis' layer drop down menus
- [x] The 'run analysis' button at the bottom of the run analysis section is not translated
- [x] When I add the word for Khmer in local language as the label to use for the translation in cambodia/index.ts, the spacing on the UI is larger than 'en' and 'fr' -> I think we should use `km` for now to be consistent. If it is a strict requirement, we could have the button size adapt to the text length.
- [x] How do I remove the French language option without deleting the info in i18n.ts? -> commented out for now
- [x] Think through how to handle 'local name' in boundary files / analysis resullts tables
- [x] Can we also configure the 'admin_level_local_names' attribute in 'admin_boundaries' so that it is tied to a specific translation and become active?
- [x] Use 'date_locale' throughout the app to decide on how to format dates. Right now it is only use in the date selectors.
- [x] adapt font size when needed
Screen Shot 2022-02-16 at 7 31 49 PM


wadhwamatic:
@ericboucher - I added translation configurations for Cambodia. Easy to implement and works well. There's a few things that are missing though:
1. When a layer is activated, the content in the legend is not translated (name, legend text, presumably legend labels)  
2. The layer names and their category headers are not translated in the 'run analysis' layer drop down menus
3. The 'run analysis' button at the bottom of the run analysis section is not translated
4. Can we also configure the 'admin_level_local_names' attribute in 'admin_boundaries' so that it is tied to a specific translation and become active?
4. How do I remove the French language option without deleting the info in i18n.ts?
5. Does 'date_locale' do anything? I was expecting it to change to dd-mm-yyyy but that doesn't seem to be the case

src/components/MapView/Analyser/AnalysisTable/index.tsx
```diff
@@ -14,11 +14,20 @@ import {
...
+import { useTranslation } from 'react-i18next';
 import { TableRow as AnalysisTableRow } from '../../../../context/analysisResultStateSlice';
 import { showPopup } from '../../../../context/tooltipStateSlice';
 import { Column } from '../../../../utils/analysis-utils';
+import { formattedTranslation } from '../../../../i18n';
 
 function AnalysisTable({ classes, tableData, columns }: AnalysisTableProps) {
+  // only display local names if local language is selected, otherwise display english name
+  const { i18n } = useTranslation();
+  const filteredColumns = columns.filter(column =>
```
ericboucher, line 26: I think a better way to do this would be to just have a "name" column and change the value inside.

Note that we also need to translate the column names

ericboucher: And for now you can simply use `i18n.resolvedLanguage === "en"`

wadhwamatic:
I continued testing for Cambodia and added the 'go to' feature. Here are a few more inputs:
1. After running an analysis, we're not displaying all levels of local admin names like we do for the English version
2. I wasn't able to add translations for 'rows per page', and '___ of ____' at the bottom of the analysis results pane
3. The text 'go to' below the 'run analysis' button doesn't translate 
4. The admin names in the 'go to' should be in local language if we're using Khmer

To replicate, use Cambodia instance REACT_APP_COUNTRY=cambodia yarn start on this branch

src/components/MapView/Analyser/AnalysisTable/index.tsx
tdlinh2712: Good idea, done

tdlinh2712:
@wadhwamatic your above comments are resolved, please check
Regarding changing the translations for 'rows per page', and '___ of ____', they are material-UI default name, hence we can change localization for them by changing locale in MuiTheme (https://mui.com/guides/localization/). However material-ui only supports Khmer from v5.0.0, hence we will need to upgrade `material-ui` and `react-dom` versions to use them. Besides, I don't think Material-UI supports Burmese and other languages in PRISM yet, so I'm implementing a quick fix of manually changing the texts we use first.

What do you think? should we go with the quick fix for now or use Material-UI localization feature?

wadhwamatic:
Thanks @tdlinh2712 for the great work on this! I see the quick fix you mentioned was made and the other issues addressed. Looks good.  

There are three things missing that I see as of now:
1. The tooltip text is not translated. This includes the use of local admin names, and the translated name of the layer - for example: 'IDPoor' displays in English instead of the Khmer text.
2. The 'go to' feature uses the English admin names rather than Khmer
3. Just noticing that the calendar date picker and timeline navigation also do not translate. I tried adding short month names (Jan, Feb, etc) to the translation file, and the full name too but without success

ericboucher:
The calendar date picker uses the date_locale that needs to be loaded from `date-fns/locale/`. I'll try to add some documentation around that.

ericboucher:
@tdlinh2712 the "go to" feature should have the same behavior as the table admin column and simply use "localAdminNames"

src/components/MapView/MapTooltip/index.tsx
```diff
@@ -1,30 +1,35 @@
...
+        {isLocalLanguageChosen(i18n)
+          ? popup.locationLocalName
+          : popup.locationName}
+      </h4>
       {Object.entries(popup.data)
         .filter(([, value]) => value.coordinates === popup.coordinates)
         .map(([key, value]) => (
           <h4 key={key}>
-            {key}: {value.data}
+            {safeTranslate(t, key)}: {safeTranslate(t, value.data.toString())}
```
ericboucher, line 32: @wadhwamatic do you think translating the value of the tooltip data is useful? My feeling is that these are mostly numbers, right?

src/components/NavBar/LanguageSelector/index.tsx
```diff
@@ -1,11 +1,21 @@
...
 function LanguageSelector({ classes }: LanguageSelectorProps) {
-  const { i18n } = useTranslation();
+  const { t, i18n } = useTranslation();
 
+  const moment = extendMoment(Moment as any);
+
+  const handleChangeLanguage = (lng: string): void => {
+    i18n.changeLanguage(lng);
+    moment.locale(
+      isLocalLanguageChosen(i18n) ? safeTranslate(t, 'date_locale') : 'en',
```
ericboucher, line 16: You can probably just load `safeTranslate(t, 'date_locale')` since the default will be 'en'

src/components/MapView/Analyser/AnalysisTable/index.tsx
```diff
@@ -14,11 +14,16 @@ import {
...
+import { useTranslation } from 'react-i18next';
 import { TableRow as AnalysisTableRow } from '../../../../context/analysisResultStateSlice';
 import { showPopup } from '../../../../context/tooltipStateSlice';
 import { Column } from '../../../../utils/analysis-utils';
+import { isLocalLanguageChosen, safeTranslate } from '../../../../i18n';
 
 function AnalysisTable({ classes, tableData, columns }: AnalysisTableProps) {
+  // only display local names if local language is selected, otherwise display english name
+  const { t, i18n } = useTranslation();
+  const filteredColumns = columns.filter(({ id }) => id !== 'localName');
```
ericboucher, line 26: Another option would be to remove the `localName` column completely. And as the name column, load the english name or the local name depending on what i18n language is selected. This will also simplify the logic below.

src/components/MapView/DateSelector/index.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { forwardRef, Ref, useEffect, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 import Moment from 'moment';
+import 'moment/locale/km';
```
ericboucher, line 5: This was loaded here as an example. We probably want to load this at a higher level, likely in `i18n.ts` ?

src/components/NavBar/LanguageSelector/index.tsx
tdlinh2712: not sure why but for english language, `safeTranslate(t, 'date_locale')` return 'date_locale' instead of 'en', that's why i did this

ericboucher:
Quick note, let's make sure that the date in the URL parameters stays in the English format

src/components/MapView/MapTooltip/index.tsx
wadhwamatic: The values will always be numbers AFAIK

wadhwamatic:
Great work! We're getting close. Here's a few more missing pieces (hopefully the last):
1. The local title of a data layer doesn't show as translated in the tooltip. For example, the 10-day rainfall estimate shows correct title in the legend, but not on the tooltip. See screenshot below

Screen Shot 2022-03-15 at 14 56 02

2. The text that is added to the legend for the results of the 'run analysis' feature is not translated. This includes the pre-layer name text "Impact Analysis on " and the layer description. See screenshot

Screen Shot 2022-03-15 at 15 03 53

3. A feature that wasn't yet used in Cambodia is the additional layer content (https://github.com/WFP-VAM/prism-frontend/pull/244). I enabled this feature for the IDPoor data layer to think about how we can address this. Perhaps we should just have a second version of the markdown file which is in local language and add it as another layer attribute?

tdlinh2712:
@wadhwamatic for "the text that is added to the legend" as well as legends for the results of the 'run analysis' feature, it's because there's no translation for fields `legend_text` and `legend` in `translation.json`. Is there any way we can do it nicely instead of entering in `translation.json` (i.e add another `translated_layer.json` file and load it to i18n)?

src/i18n.ts
```diff
@@ -122,9 +126,9 @@ export const safeTranslate = (translator: any, key: string | undefined) => {
     return translator(key);
   }
   // eslint-disable-next-line no-console
-  console.info(
-    `Translation for "${key}" is not configured in your translation file.`,
-  );
+  // console.info(
```
ericboucher, line 129: I think this is useful to keep for now while we are "developin"

src/utils/analysis-utils.ts
```diff
@@ -631,10 +632,19 @@ export class BaselineLayerResult {
...
-    }`;
+  getTitle(t?: any): string {
+    return t
+      ? this.getTranslatedTitle(t)
+      : `${this.getBaselineLayer().title} exposed to ${
+          this.getHazardLayer().title
+        }`;
+  }
+
+  getTranslatedTitle(t: any): string {
```
Ry-DS, line 643: possible to avoid this any?

src/utils/analysis-utils.ts
```diff
@@ -631,10 +632,19 @@ export class BaselineLayerResult {
     return LayerDefinitions[this.baselineLayerId] as AdminLevelDataLayerProps;
   }
 
-  getTitle(): string {
-    return `${this.getBaselineLayer().title} exposed to ${
-      this.getHazardLayer().title
-    }`;
+  getTitle(t?: any): string {
```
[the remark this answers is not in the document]
Ry-DS: We could also do a better name than `t`

src/types/i18n.d.ts
```diff
@@ -0,0 +1,7 @@
+import { resources } from '../i18n';
+
+declare module 'react-i18next' {
```
Ry-DS, line 3: Isn't this redundant? https://www.npmjs.com/package/@types/react-i18next exists, and according to the link the package should have its own types

src/i18n.ts
```diff
@@ -0,0 +1,139 @@
...
+    ns: ['translation'],
+    defaultNS: 'translation',
+  });
+
+export const safeTranslate = (translator: any, key: string | undefined) => {
+  if (key === undefined) {
+    return undefined;
+  }
+  if (key in resources.en.translation) {
+    // @ts-ignore
```
Ry-DS, line 125: a comment to explain why we are ignoring this. Maybe we can fix the typing?

src/i18n.ts
```diff
@@ -0,0 +1,139 @@
...
+    interpolation: {
+      escapeValue: false, // react already safes from xss
+    },
+    fallbackLng: 'en',
+    preload: languages,
+    ns: ['translation'],
+    defaultNS: 'translation',
+  });
+
+export const safeTranslate = (translator: any, key: string | undefined) => {
```
Ry-DS, line 120: a type other than `any`, and use optional `key?:string`

src/i18n.ts
```diff
@@ -0,0 +1,139 @@
...
+  if (key === undefined) {
```
Ry-DS, line 121: falsey check?

src/i18n.ts
```diff
@@ -0,0 +1,139 @@
...
+    return translator(key);
+  }
+  // eslint-disable-next-line no-console
+  // console.info(
+  //   `Translation for "${key}" is not configured in your translation file.`,
+  // );
+  return key;
+};
+
+export function isLocalLanguageChosen(lang: typeof i18n): boolean {
```
Ry-DS, line 135: Not sure if 'local' is accurate here, as english wouldn't be considered local? Better to just call `isEnglishLanguageSelected`?

src/components/NavBar/LanguageSelector/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+function LanguageSelector({ classes }: LanguageSelectorProps) {
+  const { i18n } = useTranslation();
+
+  const handleChangeLanguage = (lng: string): void => {
+    i18n.changeLanguage(lng);
+  };
+  // If there is only one language, hide the selector
+  if (languages.length <= 1) {
+    return <></>;
```
Ry-DS, line 14:
```suggestion
    return null;
```
is this ok?

src/components/MapView/Analyser/AnalysisTable/index.tsx
```diff
@@ -112,6 +121,12 @@ function AnalysisTable({ classes, tableData, columns }: AnalysisTableProps) {
         page={page}
         onChangePage={handleChangePage}
         onChangeRowsPerPage={handleChangeRowsPerPage}
+        labelRowsPerPage={safeTranslate(t, 'Rows Per Page')}
+        labelDisplayedRows={({ from, to, count }) => {
+          return `${from}–${to} ${safeTranslate(t, 'of')} ${
```
Ry-DS, line 126: a comment would be nice here, looks complex. Can't tell why we are checking -1 just by looking

Ry-DS (commented):
Great stuff!

src/i18n.ts
Ry-DS: In fact, this looks like a great log to keep permanently so future features are reminded to add translations for their feature. Possibly use the notification context to publish a visible popup if a translation is not found.

src/i18n.ts
tdlinh2712: not sure about this, @ericboucher can you explain?
i also prefer the function to return "" instead of undefined

src/i18n.ts
tdlinh2712: the logic is if the `resolvedLanguage` is not 'en', and in the future should be if `resolvedLanguage` is not en and fr, so localLanguage should be correct

src/i18n.ts
```diff
@@ -0,0 +1,139 @@
...
+      Statistic: 'Statistic',
+      Threshold: 'Threshold',
+      Above: 'Above',
+      Below: 'Below',
+      Mean: 'Mean',
+      Median: 'Median',
+      Date: 'Date',
+    },
+  },
+  // fr: {
```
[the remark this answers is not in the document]
tdlinh2712: @ericboucher should i uncomment this or leave it there for now?

src/i18n.ts
tdlinh2712: @ericboucher i think key should be a required argument? is there any case where key is `undefined`?

src/i18n.ts
tdlinh2712: yes I can put this log back, however it's just make the console window full of console.log() statements which actually interfere when i was debugging other issues

src/components/MapView/DateSelector/__snapshots__/index.test.tsx.snap
```diff
@@ -27,7 +27,7 @@ exports[`renders as expected 1`] = `
             <mock-button
               variant="outlined"
             >
-              07/02/2018
+              Jul 2, 2018
```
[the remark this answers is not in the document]
tdlinh2712: yes, because we changed the date format to "PP" for localisation

tdlinh2712:
@ericboucher regarding changing the date format, I noticed that all formatted dates in the repo at the moment are not for display, but for `updateHistory()` function, which updates the URL, therefore I don't think it's necessary to change the date format to fit localisation. the dates that are displayed at the moment are in date pickers, which have already been translated properly.

src/i18n.ts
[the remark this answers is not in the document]
ericboucher: @wadhwamatic can you think of a country that will use French? We can move that in a translation file for that country directly

src/i18n.ts
ericboucher: Maybe call it, `isLocalLanguageEnglish` and revert the boolean?

src/components/MapView/DateSelector/__snapshots__/index.test.tsx.snap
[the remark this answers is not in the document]
ericboucher: Yes but in English we might want to keep it as we had before. It's probably ok though, just something to watchout for

src/i18n.ts
ericboucher: hmm it should not be full of logs, otherwise it means one of:
1. the translation is not available
2. we are translating the same content multiple times

src/i18n.ts
ericboucher: This is just to be safe with "0". The function is always returning a string

src/components/MapView/Analyser/AnalysisTable/index.tsx
ericboucher: this is temporary, I added a small comment

ericboucher:
TODO:
- translate errors
- decide what to do with "No Data"

ericboucher:
@wadhwamatic for the additional layer content, we would likely use the folder structure and `layer.contentPath` to toggle between different `.md` files.

If it's not needed for Cambodia, I think we can hold of on implementing it until we actual have translated files to toggle between. wdyt?

src/components/MapView/Analyser/AnalysisTable/index.tsx
ericboucher: @tdlinh2712 what do you think?

tdlinh2712: I tried to do it, however since `columns` are loaded from props, the value of name column are not reloaded when language is changed

src/i18n.ts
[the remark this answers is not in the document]
wadhwamatic: @ericboucher - we'll use French for the west / central africa deployment called 'rbd'

wadhwamatic:
> * [ ]  (optional) translate error messages (@wadhwamatic how important is it? One potential downside I see is around communication, where keeping error messages in English might make debugging easier).

@ericboucher - is this referring to the messages like 'no data found for mm-dd-yyyy, loading the closest date mm-dd-yyyy instead'? it would be good to have those translated too, but not urgent

wadhwamatic:
> If it's not needed for Cambodia, I think we can hold of on implementing it until we actual have translated files to toggle between. wdyt?

we'll need time to get that content all finalized and translated. i would guess that won't happen for another 2-3 weeks

src/i18n.ts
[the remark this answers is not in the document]
tdlinh2712: @ericboucher regardless, if french is chosen, we will use location's English name instead of its local name, is that correct? if that's the case, in `isEnglishLanguageSelected` we will still need to check if the chosen language is English/French, therefore I'm still wondering if that's the best function name?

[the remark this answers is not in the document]
ericboucher: No because the app's base language will still be English, so I think `isEnglishLanguageSelected` is a good name for now. We can adapt if needed

src/components/MapView/Analyser/AnalysisTable/index.tsx
ericboucher: ok, good catch, then I guess we can leave it like that :)

src/i18n.ts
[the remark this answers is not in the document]
tdlinh2712: done

src/components/MapView/Layers/ImpactLayer/index.tsx
```diff
@@ -24,18 +24,17 @@ import {
...
+import { i18nTranslator, safeTranslate } from '../../../../i18n';
 
 const linePaint: LinePaint = {
   'line-color': 'grey',
   'line-width': 1,
   'line-opacity': 0.3,
 };
 
-function getHazardData(evt: any, operation: string) {
+function getHazardData(t: i18nTranslator, evt: any, operation: string) {
```
ericboucher, line 35: Nice work. For all these functions, let's make `t` optional and move it as the last argument.

wadhwamatic:
To test the exposure analysis flow, I added translations to Myanmar (dummy data copied from Cambodia). That flow isn't addressed yet. See screenshots below
1. Missing translations in the legend area of 'Exposure analysis', 'Clear analysis' and 'Table view'. 
2. After clicking on table view, we're also missing translations in the data drawer of the layer name and 'Download as CSV'

Steps to replicate:
1. REACT_APP_COUNTRY=myanmar yarn start
2. Load tropical storm layer on 2016-05-21: http://localhost:3000/?hazardLayerId=adamts_buffers&date=2016-05-21
3. Click 'exposure analysis' button on the wind buffer layer in the legend

Screen Shot 2022-03-19 at 12 05 53
Screen Shot 2022-03-19 at 12 06 03
Screen Shot 2022-03-19 at 12 17 51

src/components/MapView/DateSelector/index.tsx
```diff
@@ -114,13 +119,16 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
     });
     setDateRange(range);
     const dateIndex = findIndex(range, date => {
-      return date.label === moment(stateStartDate).format('MMM DD YYYY');
+      return (
+        date.label ===
+        moment(stateStartDate).locale(locale).format('MMM DD YYYY')
```
Ry-DS, line 124: This format is duplicated on line 115, consider making it a constant string?

src/components/MapView/Layers/ImpactLayer/index.tsx
```diff
@@ -23,17 +24,17 @@ import {
...
```
Ry-DS, line 35: chance we can fix the any type while we're here?

src/components/MapView/Layers/ImpactLayer/index.tsx
```diff
@@ -85,8 +87,13 @@ const ImpactLayer = ({ classes, layer }: ComponentProps) => {
   };
 
   const hazardLayerDef = LayerDefinitions[layer.hazardLayer];
+  // TODO - Translate
```
Ry-DS, line 90: TODO reminder

src/i18n.ts
```diff
@@ -0,0 +1,142 @@
...
+  (a, v) => ({ ...a, [v]: { translation: translation[v] } }),
+  {},
+);
+
+const englishKeys = Object.keys(translation)
+  .flatMap(language => {
+    return Object.keys(translation[language]);
+  })
+  .reduce((a, v) => ({ ...a, [v]: v }), {});
+
```
Ry-DS, line 90: better names than `a` and `v`?

src/i18n.ts
```diff
@@ -0,0 +1,142 @@
...
+  formattedTranslation,
+);
+
+export const languages = Object.keys(resources);
+
+// Register other date locales to be used by our DatePicker
+registerLocale('en', en);
+registerLocale('fr', fr);
+registerLocale('km', km);
+const moment = extendMoment(Moment as any);
```
Ry-DS, line 105: I remember we've done this before out of neccessity. Is it possible to consolidate it to one util file so we can fix it in one place if we ever find a way to remove the `any`?

src/i18n.ts
```diff
@@ -0,0 +1,142 @@
...
+    },
+    fallbackLng: 'en',
+    preload: languages,
+    ns: ['translation'],
+    defaultNS: 'translation',
+  });
+
+export const safeTranslate = (
+  translator: i18nTranslator,
+  key: string | undefined,
```
Ry-DS, line 123:
```suggestion
  key: string | undefined,
```
```suggestion
  key?: string,
```

src/i18n.ts
```diff
@@ -0,0 +1,142 @@
...
+  if (key === undefined) {
+    return '';
+  }
+  if (key in resources.en.translation) {
+    return translator(key);
+  }
+  // eslint-disable-next-line no-console
+  console.info(
+    `Translation for "${key}" is not configured in your translation file.`,
+  );
```
Ry-DS, line 134: Consider making this `console.warning/error` or making a popup with the notification context

Ry-DS (commented):
Mostly code cleanup things but otherwise looking good!

src/components/NavBar/index.tsx
```diff
@@ -16,24 +16,28 @@ import {
...
-  },
-];
+import LanguageSelector from './LanguageSelector';
 
 function NavBar({ classes }: NavBarProps) {
+  const { t } = useTranslation();
+
+  const rightSideLinks = [
+    {
+      title: t('about'),
```
Ry-DS, line 30: make sure to wrap this in safeTranslate

Ry-DS:
A better architecture for this might be a new hook:
```typescript
export function useSafeTranslation(): {
  t: i18nTranslator;
  i18n: typeof i18n;
  ready: boolean;
} {
  const { t, ...rest } = useTranslation();
  return {
    t: (key: string) => {
      if (key === undefined) {
        return '';
      }
      if (key in resources.en.translation) {
        return t(key);
      }
      // eslint-disable-next-line no-console
      console.info(
        `Translation for "${key}" is not configured in your translation file.`,
      );
      return key;
    },
    ...rest,
  };
}
```

Which will make devs less likely to forget wrapping the `t` function in `safeTranslate`. Thoughts?

src/components/NavBar/index.tsx
```diff
@@ -79,7 +83,7 @@ function NavBar({ classes }: NavBarProps) {
               component={Link}
               to="/"
             >
-              Prism
+              {t('Prism')}
```
Ry-DS, line 86: Remember to wrap this too

src/i18n.ts
tdlinh2712: done

src/components/MapView/Layers/ImpactLayer/index.tsx
tdlinh2712: I can, but looking into it, i think it might result from not imposing typing in other props updating functions like `getFeatureInfoPropsData`, so I prefer to do this in a different refactoring work item

ericboucher:
@Ry-DS I think `t` should still be mandatory in `safeTranslate` so maybe the hook is overkill? wdyt?

src/components/MapView/DateSelector/TimelineItems/index.tsx
```diff
@@ -50,7 +51,9 @@ function TimelineItems({
             )}
             {availableDates
               .map(availableDate =>
-                moment(availableDate + USER_DATE_OFFSET).format('MMM DD YYYY'),
+                moment(availableDate + USER_DATE_OFFSET).format(
+                  DEFAULT_DATE_FORMAT,
```
ericboucher, line 55: This was `MONTH_FIRST_DATE_FORMAT`

src/components/MapView/DateSelector/index.tsx
```diff
@@ -109,16 +113,16 @@ function DateSelector({ availableDates = [], classes }: DateSelectorProps) {
...
+        month: date.format(MONTH_ONLY_DATE_FORMAT),
         isFirstDay: date.date() === date.startOf('month').date(),
       };
     });
     setDateRange(range);
     const dateIndex = findIndex(range, date => {
       return (
         date.label ===
-        moment(stateStartDate).locale(locale).format('MMM DD YYYY')
+        moment(stateStartDate).locale(locale).format(DEFAULT_DATE_FORMAT)
```
ericboucher, line 125: This was `MONTH_FIRST_DATE_FORMAT`

ericboucher:
@tdlinh2712 I translated the exposure analysis piece and resolved a few minor mismatch with dates. I will give it a final review but overall do you think we are gtg?

ericboucher:
@wadhwamatic we can not translate the storm legend icons however because they are fetched directly as an image from the geoserver. Can you cleanup/remove the myanmar translation once you're done testing? I'll move the FR translation but otherwise I think we're getting ready to merge this

Ry-DS:
> @Ry-DS I think `t` should still be mandatory in `safeTranslate` so maybe the hook is overkill? wdyt?

Yup the current implementation works too, up to personal preference since we can either forget one or the other when using it in new places.

tdlinh2712:
@ericboucher i think using the hook is better since we won't have to import 2 things from 2 different files. I can refactor it quickly tomorrow if you're ok with it, then i think it's good to go.

ericboucher:
@tdlinh2712 if you think you have a clean plan in mind, go for it :)

wadhwamatic:
> @wadhwamatic we can not translate the storm legend icons however because they are fetched directly as an image from the geoserver. Can you cleanup/remove the myanmar translation once you're done testing? I'll move the FR translation but otherwise I think we're getting ready to merge this

@ericboucher - translation of the exposure analysis UI looks good, but when I click on an admin area to see a result I get an error. To reproduce, run exposure analysis on tropical storm layer on 2016-05-21 (http://localhost:3000/?hazardLayerId=adamts_buffers&date=2016-05-21). After the results load, click on an admin area in the storm path. The tooltip appears with the results but within a second, there's an error:

Screen Shot 2022-03-21 at 12 37 58

Guessing the fix for decimals wasn't applied to exposure analysis yet?

wadhwamatic:
@ericboucher . Thanks for quick fix on the rounding issue. Looks like the last piece is to translate 'Download as CSV' in the exposure analysis results:
Screen Shot 2022-03-21 at 16 12 34

ericboucher:
@wadhwamatic it was just a matter of letter capitalization -> `Download as CSV`

wadhwamatic:
> @wadhwamatic it was just a matter of letter capitalization -> `Download as CSV`

ah, I tried a few combinations but not that one. I see it working now. I'll go ahead and remove the Myanmar tests so we can merge this. Thanks

wadhwamatic (approved):
This is great work, thanks @ericboucher, @tdlinh2712 and @Ry-DS

tdlinh2712:
@ericboucher 

> @tdlinh2712 if you think you have a clean plan in mind, go for it :)

Done. However in `DateSelector.tsx`, when using the `useSafeTranslation` hook to update the date range, it resulted in the timeline picker being re-rendered too many times, so I used the `useTranslation` hook, and I think it's better to have a work item to investigate further/refactor this `DateSelector` component, but for now functionality wise, i think this is good to go

ericboucher:
@wadhwamatic we kept the console logs for now, so the app should complain in the console about missing translations. You can just copy-paste the casing from there. We could also try to be fancy and match case insensitive keys, but that's not how i18n works by default so we would have to change a few things

Ry-DS:
> @ericboucher
> 
> > @tdlinh2712 if you think you have a clean plan in mind, go for it :)
> 
> Done. However in `DateSelector.tsx`, when using the `useSafeTranslation` hook to update the date range, it resulted in the timeline picker being re-rendered too many times, so I used the `useTranslation` hook, and I think it's better to have a work item to investigate further/refactor this `DateSelector` component, but for now functionality wise, i think this is good to go

Consider investigating if `useMemo` or `useCallback` will solve this. Shouldn't be too hard to solve now as this happens quite often in react haha. Let me know if you have any issues using the above two functions (what I would try first is wrapping the new `t` function in `useCallback`)

PR #391, opened 2022-03-01
Add a sort option in 'run analysis' results

### LINKED ISSUE #357: Add a sort option in 'run analysis' results
It would be useful if the table view of analysis results allowed for sorting by each column header. It's also not clear how the initial results are currently sorted. Please clarify.

tdlinh2712:
Closes #357 
Deploy: http://prism-357.surge.sh/
Currently the analysis table returns the results from the API, not sure how they are sorted.
Added the option to sort results asc/desc by each column header. 
image

### DIFF
--- a/src/components/MapView/Analyser/AnalysisTable/index.tsx
+++ b/src/components/MapView/Analyser/AnalysisTable/index.tsx
@@ -38,7 +38,7 @@ function AnalysisTable({ classes, tableData, columns }: AnalysisTableProps) {
     setPage(0);
   };
 
-  const handleChangeOrderBy = (newSortColumn: string) => {
+  const handleChangeOrderBy = (newSortColumn: Column['id']) => {
     const isAsc = sortColumn === newSortColumn && sortDirection === 'asc';
     setSortColumn(newSortColumn);
     setSortDirection(isAsc ? 'desc' : 'asc');
@@ -52,7 +52,10 @@ function AnalysisTable({ classes, tableData, columns }: AnalysisTableProps) {
               {columns.map(column => (
                 <TableCell key={column.id} className={classes.tableHead}>
                   <TableSortLabel
-                    classes={{ root: classes.sortLabel }}
+                    classes={{
+                      root: classes.tabelSortLabelRoot,
+                      icon: classes.tabelSortLabelIcon,
+                    }}
                     active={sortColumn === column.id}
                     direction={sortColumn === column.id ? sortDirection : 'asc'}
                     onClick={() => handleChangeOrderBy(column.id)}
@@ -128,10 +131,18 @@ const styles = (theme: Theme) =>
     innerAnalysisButton: {
       backgroundColor: '#3d474a',
     },
-    sortLabel: {
+    tabelSortLabelIcon: {
+      color: theme.palette.text.primary,
+      opacity: 1,
+      '&:hover': {
+        opacity: 0.5,
+      },
+    },
+    tabelSortLabelRoot: {
       color: theme.palette.text.primary,
       '&:hover': {
-        color: theme.palette.grey[100],
+        color: theme.palette.text.primary,
+        opacity: 0.5,
       },
     },
   });
--- a/src/components/MapView/Analyser/AnalysisTable/index.tsx
+++ b/src/components/MapView/Analyser/AnalysisTable/index.tsx
@@ -9,7 +9,6 @@ import {
   TablePagination,
   TableRow,
   TableSortLabel,
-  Theme,
   withStyles,
   WithStyles,
 } from '@material-ui/core';
@@ -52,10 +51,6 @@ function AnalysisTable({ classes, tableData, columns }: AnalysisTableProps) {
               {columns.map(column => (
                 <TableCell key={column.id} className={classes.tableHead}>
                   <TableSortLabel
-                    classes={{
-                      root: classes.tabelSortLabelRoot,
-                      icon: classes.tabelSortLabelIcon,
-                    }}
                     active={sortColumn === column.id}
                     direction={sortColumn === column.id ? sortDirection : 'asc'}
                     onClick={() => handleChangeOrderBy(column.id)}
@@ -118,7 +113,7 @@ function AnalysisTable({ classes, tableData, columns }: AnalysisTableProps) {
   );
 }
 
-const styles = (theme: Theme) =>
+const styles = () =>
   createStyles({
     tableContainer: {
       border: '2px solid',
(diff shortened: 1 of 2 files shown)


ericboucher:
Awesome!

ericboucher:
@Ry-DS could you help @tdlinh2712 deploy to surge so that we can test this out? :)

wadhwamatic:
This is great, thanks @tdlinh2712 ! One minor request - can we make the arrows visible without having to hover? It's easy to not notice that the header is sortable otherwise.

src/components/MapView/Analyser/AnalysisTable/index.tsx
```diff
@@ -41,13 +51,20 @@ function AnalysisTable({ classes, tableData, columns }: AnalysisTableProps) {
             <TableRow>
               {columns.map(column => (
                 <TableCell key={column.id} className={classes.tableHead}>
-                  {column.label}
+                  <TableSortLabel
+                    classes={{ root: classes.sortLabel }}
+                    active={sortColumn === column.id}
+                    direction={sortColumn === column.id ? sortDirection : 'asc'}
+                    onClick={() => handleChangeOrderBy(column.id)}
```
Ry-DS, line 58: `column.id` is strongly typed here:
https://github.com/WFP-VAM/prism-frontend/blob/de8105cc82950b7006e39ab3ab7b8eef112eab9e/src/utils/analysis-utils.ts#L52
but we're losing it by classifying it to string. 
Reckon we can use `Column['id']` instead of `string`?

src/components/MapView/Analyser/AnalysisTable/index.tsx
```diff
@@ -111,6 +128,12 @@ const styles = () =>
     innerAnalysisButton: {
       backgroundColor: '#3d474a',
     },
+    sortLabel: {
+      color: theme.palette.text.primary,
```
Ry-DS, line 132: nice use of the theme! We need to do this more, as you could notice in places like line 129

src/components/MapView/Analyser/AnalysisTable/index.tsx
```diff
@@ -19,6 +22,8 @@ import { Column } from '../../../../utils/analysis-utils';
 function AnalysisTable({ classes, tableData, columns }: AnalysisTableProps) {
   const [page, setPage] = useState(0);
   const [rowsPerPage, setRowsPerPage] = useState(10);
+  const [sortColumn, setSortColumn] = useState('');
```
Ry-DS, line 25: Consider typing this with `Column['id']` (see later comment)

Ry-DS (commented):
Looking good! Small suggestion to improve typing otherwise looks great.

src/components/MapView/Analyser/AnalysisTable/index.tsx
tdlinh2712: changed to `Column['id']`

src/components/MapView/Analyser/AnalysisTable/index.tsx
tdlinh2712: done

tdlinh2712:
@wadhwamatic @ericboucher @Ry-DS deploy link updated in description

ericboucher:
Great! From a design perspective, I would actually reverse the white/black color, no?

src/muiTheme.tsx
```diff
@@ -110,6 +110,28 @@ const theme: any = createMuiTheme({
         color: 'white',
       },
     },
+    MuiTableSortLabel: {
```
Ry-DS, line 113: Good call putting this in the main theme :+1:

src/components/MapView/Analyser/AnalysisTable/index.tsx
```diff
@@ -19,6 +21,8 @@ import { Column } from '../../../../utils/analysis-utils';
 function AnalysisTable({ classes, tableData, columns }: AnalysisTableProps) {
   const [page, setPage] = useState(0);
   const [rowsPerPage, setRowsPerPage] = useState(10);
+  const [sortColumn, setSortColumn] = useState<Column['id']>();
+  const [sortDirection, setSortDirection] = useState<'asc' | 'desc'>('asc');
```
Ry-DS, line 25: Nitpick: For something with 2 states I'd prefer a boolean (isAscending) so less can go wrong (string can be set to something random even if it's protected by TS). This is completely up to personal preference however.

Ry-DS (approved):
Works great!

wadhwamatic (approved):
Looks good, nice work!

src/muiTheme.tsx
```diff
@@ -110,6 +110,28 @@ const theme: any = createMuiTheme({
...
+      root: {
+        color: white,
+        '&:hover': {
+          color: white,
+          opacity: 0.5,
+        },
+        '&$active': {
+          '&& $icon': {
```
ericboucher, line 121: Is there a more material-ui way to do achieve this maybe? cc @Ry-DS

tdlinh2712: I don't think so because it's how material-ui set the styles in `@material-ui/core/TableSortLabel/TableSortLabel.js`.

src/components/MapView/Analyser/AnalysisTable/index.tsx
tdlinh2712: agreed. done

src/muiTheme.tsx
Ry-DS: I reckon there is actually. Looking at the applied classes: 
image
We should be able to target them directly with cleaner lables instead of this:
```ts
    MuiTableSortLabel: {
      root:{
        ...
      },
      active:{
        ...
      },
      icon:{
        ...
      }
    }
```

ericboucher: @tdlinh2712 can you give this a shot?

tdlinh2712: @ericboucher @Ry-DS doesn't work, I've tried using Ryan's suggestion, got:
image

, and
```
root: {
...
      },
      active: {
        '&& $icon': {
          opacity: 1,
          color: white,
        }
      },
```
, got:
image

Both were overwritten by the root class. I've tried it and other approaches over the past few days and I couldn't find one that worked apart from writing it inside the root element.

Ry-DS: Yup that's fair. I got it to work with `important` of course:
```
    MuiTableSortLabel: {
      root: {
        color: white,
        '&:hover': {
          color: white,
          opacity: 0.5,
        },
      },
      active: {
        '&>svg': { color: `${white} !important` },
      },
      icon: {
        color: white,
        opacity: 0.2,
        '&:hover': {
          opacity: 0.5,
        },
      },
    }
```
But honestly I reckon what we have here works fine (obvious to the reader that it activates on active)

ericboucher:
@tdlinh2712 I merged that a bit too quickly. Small issue: when we change the selected column for sorting, we should also reset pagination at the same time. Could you open a small PR for this? Thanks :)

PR #392, opened 2022-03-01
Add admin parent name to analysis table

### LINKED ISSUE #356: 'Run analysis' results should show each admin level's name
When we run an analysis on an admin 2, or admin 3 (etc.) level dataset, we should display the parent admin level's names in the table as well.

tdlinh2712:
This closes #356 
deploy: http://prism-issue-356.surge.sh/
modified the Name column in analysis table to include parent admin level's names. This also makes the name matches location name on tooltip
image

### DIFF
--- a/src/components/MapView/Layers/BoundaryLayer/index.tsx
+++ b/src/components/MapView/Layers/BoundaryLayer/index.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { useDispatch, useSelector } from 'react-redux';
-import { get } from 'lodash';
 import { GeoJSONLayer } from 'react-mapbox-gl';
 import * as MapboxGL from 'mapbox-gl';
 import { showPopup } from '../../../../context/tooltipStateSlice';
--- a/src/components/MapView/Layers/BoundaryLayer/index.tsx
+++ b/src/components/MapView/Layers/BoundaryLayer/index.tsx
@@ -8,7 +8,7 @@ import { LayerData } from '../../../../context/layers/layer-data';
 import { layerDataSelector } from '../../../../context/mapStateSlice/selectors';
 import { toggleSelectedBoundary } from '../../../../context/mapSelectionLayerStateSlice';
 import { isPrimaryBoundaryLayer } from '../../../../config/utils';
-import { getLocationName } from '../../../../utils/name-utils';
+import { getFullLocationName } from '../../../../utils/name-utils';
 
 function onToggleHover(cursor: string, targetMap: MapboxGL.Map) {
   // eslint-disable-next-line no-param-reassign, fp/no-mutation
@@ -30,7 +30,7 @@ function BoundaryLayer({ layer }: { layer: BoundaryLayerProps }) {
 
   const onClickFunc = (evt: any) => {
     const coordinates = evt.lngLat;
-    const locationName = getLocationName(layer, evt.features[0]);
+    const locationName = getFullLocationName(layer, evt.features[0]);
     dispatch(showPopup({ coordinates, locationName }));
     // send the selection to the map selection layer. No-op if selection mode isn't on.
     dispatch(
--- a/src/components/MapView/Layers/BoundaryLayer/index.tsx
+++ b/src/components/MapView/Layers/BoundaryLayer/index.tsx
@@ -9,6 +9,7 @@ import { LayerData } from '../../../../context/layers/layer-data';
 import { layerDataSelector } from '../../../../context/mapStateSlice/selectors';
 import { toggleSelectedBoundary } from '../../../../context/mapSelectionLayerStateSlice';
 import { isPrimaryBoundaryLayer } from '../../../../config/utils';
+import { getLocationName } from '../../../../utils/name-utils';
 
 function onToggleHover(cursor: string, targetMap: MapboxGL.Map) {
   // eslint-disable-next-line no-param-reassign, fp/no-mutation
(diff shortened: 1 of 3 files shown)


src/context/analysisResultStateSlice.ts
```diff
@@ -126,7 +126,9 @@ function generateTableFromApiData(
     );
 
     const name: string =
-      featureBoundary?.properties?.[adminLevelName] || 'No Name';
+      adminLayer.adminLevelNames
```
ericboucher, line 129: Let's create a util function so that the tooltip and table can reuse the same and stay in sync

wadhwamatic:
This looks good, but two questions come up:
1. How will this work in combination with PR 391 https://github.com/WFP-VAM/prism-frontend/pull/391? I assume we're sorting on the the lowest admin level (admin 3), but this will look rather confusing for a user. It may make more sense to include each name level as a separate column instead to make sorting more clear
2. It seems inconsistent to have local name only display the lowest level while English shows multiple levels. However, since we're working on localization separate (PR 382 https://github.com/WFP-VAM/prism-frontend/pull/382) it seems we can just drop local names in this view if the localized version will include this information anyway. @ericboucher - thoughts?

tdlinh2712:
@wadhwamatic 
to answer q1: with the current solution, the name will be sorted alphabetically based on the highest admin level name. If two locations share the same level 1 name, then it would be sorted based on level 2 name, and so on. 
for example:

Kayah, Loikaw, Hpruso
Shan (North), Lashio, Hseni
Shan (North), Lashio, Kunlong
Shan (South), Loilen, Laihka
Shan (South), Loilen, Mongkaing

I don't think this is bad since we can see all locations of the same parent admin-level together

tdlinh2712:
@wadhwamatic @ericboucher deploy link updated

wadhwamatic:
> @wadhwamatic to answer q1: with the current solution, the name will be sorted alphabetically based on the highest admin level name. If two locations share the same level 1 name, then it would be sorted based on level 2 name, and so on. for example:
> 
> I don't think this is bad since we can see all locations of the same parent admin-level together

@tdlinh2712 - agreed, this works fine for us. This is good from my side, but pending final code review from @ericboucher or @Ry-DS 

@ericboucher - note 'local name' in boundary files needs to be thought through as part of the localization work

src/utils/name-utils.ts
```diff
@@ -0,0 +1,13 @@
+import { get } from 'lodash';
+import { BoundaryLayerProps } from '../config/types';
+
+export function getLocationName(
+  layer: BoundaryLayerProps,
+  featureBoundary: any,
```
Ry-DS, line 6: Should avoid `any` use at all costs. Can't you use `Feature<Geometry, GeoJsonProperties>`? If there's a type we can't match it should be defined manually or with `unknown`

src/utils/name-utils.ts
```diff
@@ -0,0 +1,13 @@
+import { get } from 'lodash';
+import { BoundaryLayerProps } from '../config/types';
+
+export function getLocationName(
```
Ry-DS, line 4: This function should have a docstring
`/**
...
**/`

src/context/analysisResultStateSlice.ts
```diff
@@ -125,8 +126,7 @@ function generateTableFromApiData(
       feature => feature.properties?.[adminLevelName] === row[adminLevelName],
     );
 
-    const name: string =
-      featureBoundary?.properties?.[adminLevelName] || 'No Name';
+    const name: string = getLocationName(adminLayer, featureBoundary);
```
Ry-DS, line 129: specifying the type here seems uneccessary (from the previous code but we could remove it now)
```suggestion
    const name = getLocationName(adminLayer, featureBoundary);
```

src/utils/name-utils.ts
Ry-DS, line 4: I like `getFullName` better than `getLocationName` thoughts?

src/utils/name-utils.ts
tdlinh2712: thanks! done

src/utils/name-utils.ts
tdlinh2712: done. Can you check the docstring to make sure what i wrote makes sense?

tdlinh2712: changed to `getFullLocationName`

src/utils/name-utils.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Format full location name in ascending admin levels
+ * @param layer admin layer, both props and data
+ * @param featureBoundary geospatial features of the location, contains names of different admin layers
+ * @returns a string containing location name in ascending admin levels i.e level 1 name, level 2 name, etc
+ */
+
+export function getFullLocationName(
+  layer: BoundaryLayerProps,
+  featureBoundary:
```
Ry-DS, line 14: Use optional `featureBoundary?: <type>` instead of an explicit `undefined`

tdlinh2712: done

Ry-DS (approved):
Looks good!

src/components/MapView/Layers/BoundaryLayer/index.tsx
```diff
@@ -30,9 +30,7 @@ function BoundaryLayer({ layer }: { layer: BoundaryLayerProps }) {
 
   const onClickFunc = (evt: any) => {
     const coordinates = evt.lngLat;
-    const locationName = layer.adminLevelNames
-      .map(level => get(evt.features[0], ['properties', level], '') as string)
-      .join(', ');
+    const locationName = getFullLocationName(layer, evt.features[0]);
```
ericboucher, line 33: I think we only need to pass `layer.adminLevel`

PR #393, opened 2022-03-01
remove rounding for tooltip value

### LINKED ISSUE #368: Values displayed in analysis results in tooltip are rounded
The values we show in the tooltip after running an analysis are rounded to whole numbers. Can we have this somehow configurable based on the hazard layer? It also doesn't match the value displayed in the table. See screenshot below from Sri Lanka dev branch, SPI hazard layer.
Screen Shot 2022-01-26 at 19 28 20

tdlinh2712:
This closes #368 
Deploy: http://prism-368.surge.sh/
Remove rounding for tooltip values so that they match the values displayed in analysis table.

However, it can be ugly when the values have many decimal points, so should we round to 2dp instead?

### DIFF
--- a/src/components/MapView/Layers/AnalysisLayer/index.tsx
+++ b/src/components/MapView/Layers/AnalysisLayer/index.tsx
@@ -76,12 +76,10 @@ function AnalysisLayer() {
         dispatch(
           addPopupData({
             [analysisData.getStatTitle()]: {
-              data: Math.round(
-                get(
-                  evt.features[0],
-                  ['properties', analysisData.statistic],
-                  'No Data',
-                ),
+              data: get(
+                evt.features[0],
+                ['properties', analysisData.statistic],
+                'No Data',
               ).toLocaleString('en-US'),
               coordinates,
             },
--- a/src/components/MapView/Layers/AnalysisLayer/index.tsx
+++ b/src/components/MapView/Layers/AnalysisLayer/index.tsx
@@ -24,7 +24,7 @@ import {
 } from '../../../../config/utils';
 
 function getRoundedData(data: number): string {
-  return data ? data.toFixed(3) : 'No Data';
+  return data ? data.toFixed(2) : 'No Data';
 }
 
 function AnalysisLayer() {
--- a/src/components/MapView/Layers/AnalysisLayer/index.tsx
+++ b/src/components/MapView/Layers/AnalysisLayer/index.tsx
@@ -22,10 +22,7 @@ import {
   getBoundaryLayerSingleton,
   LayerDefinitions,
 } from '../../../../config/utils';
-
-function getRoundedData(data: number): string {
-  return data ? data.toFixed(2) : 'No Data';
-}
+import { getRoundedData } from '../../utils';
 
 function AnalysisLayer() {
   // TODO maybe in the future we can try add this to LayerType so we don't need exclusive code in Legends and MapView to make this display correctly
--- a/src/components/MapView/Layers/AnalysisLayer/index.tsx
+++ b/src/components/MapView/Layers/AnalysisLayer/index.tsx
@@ -23,6 +23,10 @@ import {
   LayerDefinitions,
 } from '../../../../config/utils';
 
+function getRoundedData(data: number): string {
+  return data ? data.toFixed(3) : 'No Data';
+}
+
 function AnalysisLayer() {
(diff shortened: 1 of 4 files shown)


ericboucher:
Yes, let's do 2 dp in the table and the tooltip I think. @wadhwamatic  any thoughts?

wadhwamatic:
Just had a look and it looks like 3 dp was implemented, which works for now. I did see that the other indicator (poverty headcount) is not rounded in the tooltip, while it is in the table view. See screenshot. Can we also round that to 3 dp? 

Screen Shot 2022-03-01 at 12 14 10

tdlinh2712:
@wadhwamatic @ericboucher changed tooltip values to be rounded to 3dp, deploy link updated

wadhwamatic:
> @wadhwamatic @ericboucher changed tooltip values to be rounded to 3dp, deploy link updated

sorry @tdlinh2712 - after seeing this in action I see that the best outcome is to round to the same values as the underlying data layer.  In the case of Sri Lanka, all of the 'baseline' layers are whole numbers. For Myanmar, we have a few layers with 1dp precision - for example 'improved_sanitation'. Ideally, we can either base rounding on the underlying data, or to add precision as a configuration option as mentioned in the issue. Can you let me know how feasible either of these approaches would be?

ericboucher:
@wadhwamatic @tdlinh2712 I think for now we should keep it simple and round to 2dp in all places and add comma separators for thousands

ericboucher:
@wadhwamatic can you confirm that you are ok with my latest suggestion? And @tdlinh2712 can you make this small tweak? :)

src/components/MapView/Layers/AnalysisLayer/index.tsx
```diff
@@ -23,6 +23,10 @@ import {
   LayerDefinitions,
 } from '../../../../config/utils';
 
+function getRoundedData(data: number): string {
```
ericboucher, line 26: Let's move this to a util function and make sure we use it in the table as well for homogeneity

ericboucher: maybe we could also expose the number of decimals in this util. defaulting to 2

wadhwamatic:
My preference would be to use the same precision as the baseline layer. I
was assuming it wouldn't be difficult to read it from the data and display
accordingly. What do you think?

wadhwamatic:
> My preference would be to use the same precision as the baseline layer. I was assuming it wouldn't be difficult to read it from the data and display accordingly. What do you think?

@ericboucher My preference would be to use the same precision as the baseline layer. I was assuming it wouldn't be difficult to read it from the data and display accordingly. What do you think?

tdlinh2712:
@ericboucher @wadhwamatic can you confirm which solution we are going with? I have no problem with both solutions.

tdlinh2712:
@wadhwamatic @ericboucher can you confirm which solution we are going to use? since I am good with both, 
@wadhwamatic regarding your preference: that means all values in both table and the tooltip will be rounded to the same decimal number as the baseline value, is that correct?

wadhwamatic:
> @wadhwamatic regarding your preference: that means all values in both table and the tooltip will be rounded to the same decimal number as the baseline value, is that correct?

@tdlinh2712 - no, what I mean is that the raster statistic is rounded (like you've already done) but that the baseline data uses the same precision as the underlying layer. the table and tooltip values should be consistent with each other.

ericboucher (approved):

wadhwamatic:
> @wadhwamatic I think that we should round it as well as you mentioned before. Unfortunately, some baseline layers have way too many decimals. So let's just round the tooltip to 3dp then and keep the table data 3dp?

@ericboucher - I thought from our call today the plan was to display the baseline layer values in the same format as the underlying data unless it is longer than 3dp. In that case, round to 3dp. I don't see that in this branch yet. Work in progress, or did I misunderstand?

ericboucher:
@wadhwamatic how about now? If it's not ok, can you give a step by step for reproduction?

wadhwamatic:
> @wadhwamatic how about now? If it's not ok, can you give a step by step for reproduction?

Nearly there. The only issue I see now is that the comma separator is missing in the tooltip.  Steps to reproduce:
1. REACT_APP_COUNTRY=srilanka
2. Run analysis > SPI 1-month + number of housholds
3. Compare number of households table values (comma separator present) with tooltip values (no comma separator). See screenshot

Screen Shot 2022-03-07 at 15 27 48

Issue #349, opened 2022-01-13
Integrate Cambodia's IDPoor data into PRISM


wadhwamatic:
The IDPoor system in Cambodia reports the number of households registered as poor by administrative area. For the upcoming deployment of PRISM in Cambodia, we need to ingrate this data as an admin_level_data layer type. The IDPoor data does not seem to be updated frequently and we're not sure about the reliability / performance of the API. 

Details about the IDPoor system can be found here: https://mop.idpoor.gov.kh/about/documents
Some light documentation on the API is here: https://mop.idpoor.gov.kh/api/external/swagger-ui.html
A query builder (which uses the API) is here: https://mop.idpoor.gov.kh/reporting/builder

A few challenges / caveats:
1. The IDPoor data collection occurs over multiple rounds over time. But from a data visualization / analysis perspective, we should retrieve the latest IDPoor data nationally, which is assembled over multiple rounds over time
2. The gazeteer codes used by the IDPoor system do not align with admin codes we've used. A boundary file that matches the IDPoor codes is available from the WFP Cambodia office
3. Multiple queries are also required to obtain all admin areas, as the queries for lower admin areas are unique to their parent. For example, to get admin2 data, a query is required for each admin1 area. In this example query, '3' is an admin1 code: https://mop.idpoor.gov.kh/api/analytics/public/units/3/children?region=ALL&isOnDemand=ALL&year=2019&round=ALL

wadhwamatic:
See also: https://docs.google.com/document/d/11nuAKVPKUTXnmLieLpgSKhJyZK2Xzdqos8iBKwk7K90/edit#heading=h.lmenolaz2xfh

ericboucher:
@wadhwamatic how often is the data updated?

wadhwamatic:
> @wadhwamatic how often is the data updated?

@ericboucher - I was trying to find the answer to this yesterday as well. I think at most, twice a year. Will confirm. Definitely ok to store locally and check for updates rather than fetching data every time

ericboucher:
If that's the case, then I think we should either store it locally and have a prebuild script that checks for updates. Or simply load at build time using `https://nextjs.org/docs/basic-features/data-fetching/get-static-props`. Do we know how reliable their API is?

wadhwamatic:
[...]

PR #394, opened 2022-03-01
Create prebuild script and load `IDPoor` cambodia data

erwagasore:
Closes: https://github.com/WFP-VAM/prism-frontend/issues/349

TODO:

- [x] Create a prebuild script to load latest `idpoor` layer data for cambodia
- [x] Make the prebuild modular to allow loading of other prebuild functions
- [ ] Create function to get and massage IDPoor data into format that will work with PRISM 

How-to use prebuild:
- Just run `yarn build` or `npm run build` and it will first run the prebuild then build processes respectively


erwagasore:
@ericboucher I did take the `prebuild` path instead of nextjs's `get-static-props` does the `script/prebuild.js` match what you wanted?

scripts/prebuild.js
```diff
@@ -0,0 +1,46 @@
+const path = require('path');
```
Ry-DS, line 1: I would avoid using js for anything. Consider making this in TS and using ts-node to run it?

public/data/cambodia/idpoor-back.json
```diff
@@ -0,0 +1,6312 @@
+{
```
Ry-DS, line 1: I'm assuming this file is now redundant as we fetch it from the prebuild script? Can we name it `idpoor-backup.json` and add some documentation in prebuild.ts to explain what this file is?

scripts/prebuild.ts
```diff
@@ -0,0 +1,63 @@
...
+
+function idPoorPrebuild(country: string): void {
+  const dataDir = path.join(rootDir, `public/data/${country}`);
+  const file = path.join(dataDir, 'idpoor.json');
+
+  switch (country) {
+    case 'cambodia':
+      console.warn('ID Poor Cambodia');
+
+      fs.access(file, fs.constants.F_OK, (fileMissingError: any) => {
```
Ry-DS, line 17: what can we do to avoid this `any` and the `any` on line 28?

scripts/prebuild.ts
```diff
@@ -0,0 +1,63 @@
...
+console.warn('Prebuilding...');
+
+const rootDir = path.dirname(__dirname);
+
+function idPoorPrebuild(country: string): void {
+  const dataDir = path.join(rootDir, `public/data/${country}`);
+  const file = path.join(dataDir, 'idpoor.json');
+
+  switch (country) {
+    case 'cambodia':
```
Ry-DS, line 14: There's a type in the frontend (`type Country`) which would be nice to use here. I have no clue if its possible however since it's in a completely different folder. All good with keeping this as a string if you don't see a way in using that.

scripts/prebuild.ts
```diff
@@ -0,0 +1,63 @@
...
+      console.warn('ID Poor Cambodia');
+
+      fs.access(file, fs.constants.F_OK, (fileMissingError: any) => {
+        if (!fileMissingError) {
+          // fetch data from idpoor api
+          axios
+            .get('http://localhost/idpoor/data')
```
Ry-DS, line 21: Out of curiosity, does this just work? Does anything need to start up before this works? If so, we should document it somewhere

public/data/cambodia/idpoor-back.json
erwagasore: I agree. I am working on a new approach which is going to better the process

erwagasore:
@Ry-DS I had a dicussion with @ericboucher on the better alternative to create an independent script which if combined with Github Actions will be able to update the `idpoor.json` and create a pull request. Also we decided to handle everything from the frontend thus `locahost` will be omitted in the next changes

wadhwamatic:
Closing this PR as we'll likely implement a different solution once the IDPoor system has been revamped

Issue #388, opened 2022-02-25
Error on 'run analysis' for RBD branch - Cannot read properties of undefined (reading 'coordinates')


wadhwamatic:
I've updated the RBD dev branch (https://github.com/WFP-VAM/prism-frontend/tree/rbd_dev_02_2022) with the latest on master (25-Feb) and included a new boundary file. When I run analysis, I get an error I've not seen before: "Cannot read properties of undefined (reading 'coordinates')"

Running the same analysis through cURL doesn't give any errors. 
curl 'https://prism-api.ovio.org/stats' \
  -H 'authority: prism-api.ovio.org' \
  -H 'cache-control: max-age=0' \
  -H 'sec-ch-ua: " Not A;Brand";v="99", "Chromium";v="98", "Google Chrome";v="98"' \
  -H 'accept: application/json' \
  -H 'content-type: application/json' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.109 Safari/537.36' \
  -H 'sec-ch-ua-platform: "macOS"' \
  -H 'origin: http://localhost:3000' \
  -H 'sec-fetch-site: cross-site' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-dest: empty' \
  -H 'referer: http://localhost:3000/' \
  -H 'accept-language: en-US,en;q=0.9,id;q=0.8' \
  --data-raw '{"geotiff_url":"https://odc.ovio.org/?service=WCS&request=GetCoverage&version=1.0.0&coverage=rfq_dekad&crs=EPSG%3A4326&bbox=-17.5%2C1.7%2C26.8%2C27.3&width=5096&height=2950&format=GeoTIFF&time=2022-02-21","zones_url":"https://prism-admin-boundaries.s3.us-east-2.amazonaws.com/wca_CHIPC_nov2021_admin1.json","group_by":"adm1_name"}' \
  --compressed

See screenshot below.
<img width="1437" alt="Screen Shot 2022-02-25 at 12 51 37" src="https://user-images.githubusercontent.com/3343536/155801248-25b1597e-6afe-4bdf-aae2-ea6a9de5caa9.png">

Steps to reproduce:
1. Use 'rbd_dev_02_2022' branch
[...]

PR #396, opened 2022-03-07
Use adminCode instead of names for analysis groupBy

wadhwamatic:
This PR addresses https://github.com/WFP-VAM/prism-frontend/issues/388. @ericboucher implemented a fix to group by admin code rather than admin name. 
This PR also includes RBD config updates which surfaced the group by issue.


wadhwamatic:
@ericboucher - I tested this in multiple deployments and found no issues except for what I noted here https://github.com/WFP-VAM/prism-frontend/issues/397 but I think that is something to do with the admin boundaries and/or stats API rather than this fix

tdlinh2712:
@ericboucher the issue is not because of the sort function, it's because there are multiple rows in the table have the same name (ie Ampil), and we were using location names as the row keys, hence the rows have non-unique keys. To fix this we can change the key to some autogenerated values/find another attribute that have unique values, or make the location names unique. Why are the names repeated in with 10-day rainfall + IDPoor?

wadhwamatic:
> @ericboucher the issue is not because of the sort function, it's because there are multiple rows in the table have the same name (ie Ampil), and we were using location names as the row keys, hence the rows have non-unique keys. To fix this we can change the key to some autogenerated values/find another attribute that have unique values, or make the location names unique. Why are the names repeated in with 10-day rainfall + IDPoor? image

@tdlinh2712 - yeah, the problem is that non-unique names is possible if we're not concatenating all admin levels. We didn't have all levels included in the Cambodia boundary file so couldn't specify the names for the hierarchy. I went ahead and generated a file that now includes names at all levels and updated the Cambodia configuration in this branch. That fixes the problem.  I don't think we should have this issue if it's clearly documented that names must be unique either on their own or through concatenation of levels

wadhwamatic:
@ericboucher - I tested the Mongolia issue (https://github.com/WFP-VAM/prism-frontend/issues/398) on this branch. Looks good now, thanks.

ericboucher:
So does this seem good to you @wadhwamatic ? Good to merge?

wadhwamatic:
> Looks all good now @wadhwamatic ? Good to merge?

@ericboucher - yes good to go. thanks

ericboucher (approved):

PR #399, opened 2022-03-08
Initial configuration for sierra leone

wadhwamatic:


src/config/index.ts
```diff
@@ -117,6 +123,12 @@ const configMap = {
     rawTables: rbdRawTables,
     defaultBoundariesFile: `${DEFAULT_BOUNDARIES_FOLDER}/wca_CHIPC_mar2021_projected_jun2021_simple.json`,
   },
+  sierraleone: {
```
ericboucher, line 126: Let's use the simpler way to import the config

ericboucher: See the translation PR

wadhwamatic: done. thanks

Issue #345, opened 2021-12-30
Maintain map position when loading a new layer


wadhwamatic:
Resetting the map position whenever we load a new layer is frustrating. A user may have zoomed / panned to a specific area of interest but then loses that positioning by activating a new layer. We need to change the behavior so that the map position is not reset whenever we load a new layer.

reed1:
Hi @wadhwamatic, there is a small patch here to prevent map re-centering. 4be2b7fa42b1084e10fe78f23a5ad7ece91d8767

You can test it on https://prism.wfp.or.id/ . If it's good enough I can make PR for that

wadhwamatic:
That's great, thanks @reed1. This just came up again today. Yes, please go ahead with a PR and I'll test it in other deployments as well.

PR #404, opened 2022-03-15
Maintain map position when loading a new layer

reed1:
For #345


ericboucher:
@reed1 I just merged a big change to use maplibre instead of mapbox and it seems to have broken your change. Sorry!

reed1:
@ericboucher @wadhwamatic It's updated now, still works with maplibre, it's just removing the center and zoom from map initial props. A better solution might be patching the library or adding position in state variable and watch the changes? but I think this will help for the time being

wadhwamatic (commented):
Tested and works as expected. Thanks!
Let's wait for review from @Ry-DS or @ericboucher before merging

ericboucher (approved):

PR #412, opened 2022-03-18
Update number with commas in case we have more than 3 decimals

ericboucher:
The current regex would not work well for more than 3 decimals.

Using `toLocaleString` instead as commas is not the default in all countries. In prevision of #382


wadhwamatic (approved):
Tested in a few instances and on Chrome and Firefox. Looks good, thanks!

tdlinh2712 (approved):
Working well, thank you!

PR #414, opened 2022-03-18
Add parsing for grouped KOBO data

ericboucher:
Still a work in progress but the API is able to return the following form data and the frontend doesn't raise the same error.

`
{'start': '2022-03-15T20:57:19.311-07:00', 'end': '2022-03-15T20:58:01.896-07:00', 'q_Enum': 'asddf', 'q_Funtion': 'asdf', 'q_Phone': 'asdf', 'Date_Dis': datetime.datetime(2022, 3, 1, 0, 0, tzinfo=datetime.timezone.utc), 'DisTyp': '4', 'Specify': 'asdf', 'Province': '20', 'District': '2007', 'Commune': '200706', 'NumVillAff': 24, 'NumFamAff': 25, 'NumPeoAff': 26, 'NumFeAff': 27, 'lat': 11.558558, 'lon': 104.927453, 'date': datetime.datetime(2022, 3, 1, 0, 0, tzinfo=datetime.timezone.utc), 'status': None}
`

Note - it requires the API to be run locally


wadhwamatic:
Thanks @ericboucher - this gets us closer. I added some more attributes in layers.json in this branch. I'm trying to use a value from the response to use as a 'measure' to plot a form submission on the map with symbology based on the value. The data has a field called 'NumPeoAff'. I can get that value back correctly in the 'feature_info_props', but not in the 'measure' attribute (see 'Number of people affected' and 'No Data' respectively in the screenshot).

Note, if I include the group name using the value 'group_tc1fy38/NumPeoAff' as the 'measure', the locations show up correctly on the map, but no value is given. If I use 'NumPeoAff', the locations do not appear on the map. I've got two data points in Phnom Penh on 1-March-2022 for testing. The screenshot below is based on 'group_tc1fy38/NumPeoAff' as the measure.

Screen Shot 2022-03-18 at 16 28 34

A couple of other things to mention:
1. The goal is still to eventually use the Commune ID to plot the location and not depend on lat/long. I just included lat/long for now so that we can figure out the grouping issue
2. The form I am testing with 'test-form-admin-codes' is a simplified version of the full form called 'ទម្រង់ ឧបទ្ទវហេតុ'. The full form was created by the WFP Cambodia office based on input from the national disaster management agency. We could request them to change the names of the variable names for form fields if needed (i.e. remove spaces, use English, etc.). Variable names are not seen by Kobo users

src/config/cambodia/layers.json
```diff
@@ -895,7 +895,7 @@
         "label": "Number of people affected"
       }
     },
-    "measure": "NumPeoAff",
+    "measure": "group_tc1fy38/NumPeoAff",
```
ericboucher, line 898: @wadhwamatic the results get flattened so you need to use the measure name without the group name. We can also change that and decide to return the full metric name including the group name. What do you prefer?

tdlinh2712:
@wadhwamatic can you let me know how to reproduce the issue in the screenshot (which layer to select) because I can't reproduce the issue?

src/config/cambodia/layers.json
wadhwamatic: @ericboucher - see my earlier comment. Currently (after flattening), if I exclude the group name, the survey submission doesn't appear on the map. At a quick glance, it looks like Kobo doesn't allow a variable name to repeat across groups, so I think using the name without the group name would be simpler. We just need to make sure the value is available consistently. As of now, the measure value isn't showing up as valid. See the 'No Data' response and black point vs. the expected red.

wadhwamatic:
> @wadhwamatic can you let me know how to reproduce the issue in the screenshot because I can't reproduce the issue?

@tdlinh2712 - use this branch: khm_kobo_03_2022. You'll have to run docker locally and run 'make api' in prism-frontend/api-flask. Then REACT_APP_COUNTRY=cambodia yarn start.

Choose the layer: exposure > field reports. http://localhost:3000/?hazardLayerId=kh_incident_report&date=2022-03-01. 

Click on a survey result (black dot north of Penh as an example). You should see two conflicting results in the tooltip. "ទម្រង់ ឧបទ្ទវហេតុ: No Data" & "Number of people affected: 2000". We should see '2000' as the value in both of these attributes.

src/config/cambodia/layers.json
ericboucher: It loaded ok for me after the latest changes on legendToStops

ericboucher:
@wadhwamatic with the flattened name and the fix on the legendStops, here is what I get:
Screen Shot 2022-03-21 at 2 52 35 PM

wadhwamatic:
> @wadhwamatic with the flattened name and the fix on the legendStops

@ericboucher - great, this is working now as expected. I added new boundary files which now should match the admin codes used in the form. I also reverted the Kobo form to 'test-form-admin-codes' which has codes as strings with a leading 0 when admin1 is < 10

src/config/cambodia/layers.json
```diff
@@ -890,97 +872,52 @@
     "legend_text": "",
     "base_url": "https://geonode.wfp.org/geoserver"
   },
-  "disaster_report": {
-    "title": "Disaster impact report",
+  "kh_incident_report": {
+    "title": "ទម្រង់ ឧបទ្ទវហេតុ",
```
ericboucher, line 876: let's use the english name until we have translation

src/config/cambodia/layers.json
```diff
@@ -890,97 +872,52 @@
...
     "type": "point_data",
-    "data": "https://prism-api.ovio.org/kobo/forms",
-    "date_url": "https://prism-api.ovio.org/kobo/forms",
+    "data": "http://localhost/kobo/forms",
```
ericboucher, line 878: Let's switch this back

Issue #411, opened 2022-03-18
Vertical scrollbars occasionally appear while the top navigation becomes inaccessible


hav142:
There is a small issue with the PRISM layout where scrollbars appear on the right and bottom sides. 

![image](https://user-images.githubusercontent.com/25983260/158934703-40a70c93-a6c1-4eda-8da5-f45e0d336512.png)

It should be fixed by moving some attributes in the app.css as I noticed that the HTML structure for the map container seems to have changed a bit from the previous version.

![image](https://user-images.githubusercontent.com/25983260/158946852-3674d489-bd0e-44b6-8208-cbba8e5ac8f0.png)

@wadhwamatic let me know if this fix is OK then I will submit a PR.

wadhwamatic:
@hafidz142 - please go ahead and implement your suggested change through a PR and we can then review it

PR #415, opened 2022-03-21
update css for app container

hav142:
For issue #411

### DIFF
--- a/src/components/App/app.css
+++ b/src/components/App/app.css
@@ -1,10 +1,10 @@
 html, body { height: 100% }
 #root {
-    display: flex;
-    flex-direction: column;
     height: 100%;
 }
 
 #app {
-    flex: 1;
+    display: flex;
+    flex-direction: column;
+    height: 100%;
 }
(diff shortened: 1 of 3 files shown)


hav142:
> @hav142 on what browser are you having this issue?

@ericboucher I tried on firefox and chrome

ericboucher:
I am not seeing this, can you deploy your version to surge.sh or something so that we can compare?

Do you also see the issue you're mentioning on https://prism-mongolia.org/ ? Can you make sure that you are not zoomed in?

Also, what's with the image changes? Let's focus on the css container issue for this PR

reed1:
I think the issue was because of `#app` container addition for login feature

The https://prism-mongolia.org/ doesn't have that container yet, but these have:
- https://prism-demo-moz.surge.sh/
- https://prism-demo-moz.surge.sh/

This is this branch deployment:
https://prism-css-container.surge.sh/

For image changes, `yarn test` should have REACT_APP_COUNTRY variable unset so it will use 'myanmar' as default country (config/index.ts) to produce test snapshots. I think @hav142 might have set it to 'mozambique'. Should we revert the commit?

hav142:
Sorry I just realized about the country config as the cause of the snapshot file change. thanks @reed1 for mentioning

ericboucher (approved):

ericboucher:
Ok sorry, I did not have the context. Looks good, thanks for the update.

PR #416, opened 2022-03-21
update boundary files for mozambique and add new layers

wadhwamatic:
Updated boundary files based on a newer version from WFP Mozambique office. Added new IPC data as admin_level_data, and included SPI wms layers


src/config/index.ts
```diff
@@ -99,7 +99,7 @@ const configMap = {
     appConfig: mozambiqueConfig,
     rawLayers: mozambiqueRawLayers,
     rawTables: mozambiqueRawTables,
-    defaultBoundariesFile: `${DEFAULT_BOUNDARIES_FOLDER}/moz_admin_boundaries.json`,
```
ericboucher, line 102: Let's use this opportunity to use to the simplified config setup for mozambique

wadhwamatic: @ericboucher - done

ericboucher (approved):

PR #418, opened 2022-03-22
update sri lanka boundaries. update zimbabwe configuration and additional wms layers

wadhwamatic:
added multiple levels for sri lanka boundaries
revisited zimbabwe layers, and added additional boundary file


src/config/index.ts
```diff
@@ -136,8 +136,7 @@ const configMap = {
     appConfig: zimbabweConfig,
     rawLayers: zimbabweRawLayers,
     rawTables: zimbabweRawTables,
-    // TODO - Add selected defaultBoundary to S3
-    defaultBoundariesFile: '',
```
ericboucher, line 140: Let's use this opportunity to simplify the config for zimbabwe and sri lanka

wadhwamatic: > Let's use this opportunity to simplify the config for zimbabwe and sri lanka

@ericboucher - done

ericboucher (approved):

Issue #352, opened 2022-01-14
Support Kobo forms that map to admin_level_data rather than point data


wadhwamatic:
In some instances (including Cambodia), Kobo form data will not include a lat/long but instead will capture data at the admin level (i.e. district or commune). We'll need to have a similar approach for pulling data from the Kobo API, but do render the result as an admin level layer rather than point layer. I'll create a form for testing soon. See 'PRISM-KHM-Disaster-Report-v2-no-lat-long' in our shared Kobo user in the meantime. It has province and district fields, but returns text rather than the admin code.

 
https://docs.google.com/document/d/1O9VvyfhY7FEprpVWN6YYGYleF92615QodWMs2o1qLxw/edit

wadhwamatic:
@ericboucher and @tdlinh2712 - I added a branch 'khm_kobo_03_2022' to test and create forms that will need to use admin codes instead of a lat/long. See issue https://github.com/WFP-VAM/prism-frontend/issues/407 for a problem related to groups in forms. 

In layers.json, change "test-form-admin-codes" to "test-form-admin-codes-no-groups". That will give you a working form with admin codes as part of the submitted data. The data fields of interest are called Province, District and Commune. For now, this form also has a lat/long as our current functionality doesn't work without it. Hoping this gives you enough to start implementing this change.

One thing of note - I think we should consider two ways to visualize data that is captured at admin level: 1) shade the full admin polygon or 2) place a point at the center of the polygon. It would be great if we could do either and allow it to be a config option.

ericboucher:
Does the commune match data we have in our admin boundary layer?

wadhwamatic:
> Does the commune match data we have in our admin boundary layer?

@ericboucher - Almost. The discrepancy I see is that the Kobo forms use leading 0s for provinces 0-9 and that carries over to district and commune codes. 

For example, in the Kobo form, province '7' is labeled '07', a district within it is '0702', and a commune within it is '070204'.
In the boundary file, that same commune is '70204'. 

I'm not sure if removing the leading 0 in the Kobo form will break something. We can try it out.

PR #421, opened 2022-03-24
Display point_data as admin_level_data

tdlinh2712:
This PR resolves https://github.com/WFP-VAM/prism-frontend/issues/352.

Currently, I only added a special fetching for when dataField equals to 'NumPeoAff'

TODO:

- [x] adapting URL generation in `admin_level_data.ts` to accept additional query params
- [x] Adapt admin_level_data.ts to correctly match admin boundaries from the KOBO api response


tdlinh2712:
@ericboucher I found that when changing the `kh_incident_report` layer to type `admin_level_data`, the date selector doesn't appear anymore, and the problem is that layers with type `admin_level_data` are considered baseline layer, and baseline layers are not considered date compatible layers. However, I don't want to change all baseline layers to date compatible layers if it only applies for  `kh_incident_report` (i.e IDPoor is a baseline layer but not a date compatible layer). How do you think we should approach this? Should we change the criteria for a date-compatible layer so that it does not depend on the layer type?
image

tdlinh2712:
TODO:
[] Refactor Layer components into reusable components so that it is easier to switch between admin layer and point layer

src/config/cambodia/layers.json
```diff
@@ -1642,6 +1642,11 @@
   "kh_incident_report": {
     "title": "Disaster impact report",
     "type": "point_data",
+    "admin_level_display": true,
+    "path": "https://prism-api.ovio.org/kobo/forms",
+    "adminCode": "Commune",
+    "adminLevel": 1,
+    "dataField": "NumPeoAff",
```
ericboucher, line 1649: We could probably simplify this and have something like this instead:

`admin_level_display: {adminCode: Commune, adminLevel: 1}`

Also, a lot of the config seems redundant betwee:
- path / data / date_url
- dataField / measure_field / feature_info_props / measure

src/config/cambodia/layers.json
```diff
@@ -1642,6 +1642,11 @@
   "kh_incident_report": {
     "title": "Disaster impact report",
     "type": "point_data",
+    "admin_level_display": true,
+    "path": "https://prism-api.ovio.org/kobo/forms",
+    "adminCode": "Commune",
```
ericboucher, line 1647: The config should be using snake_case whenever possible, the config loader should translate everything into camelCase automagically :)

src/config/cambodia/layers.json
```diff
@@ -1656,10 +1661,6 @@
       "Commune": {
         "type": "text",
         "label": "Commune"
-      },
-      "NumPeoAff": {
```
ericboucher, line 1660: I noticed that removing this breaks the "measure" field field of the popup data, not sure why

src/config/cambodia/layers.json
tdlinh2712: @ericboucher can you check if the current fix is sufficient enough?

src/components/MapView/Layers/PointDataLayer/index.tsx
```diff
@@ -28,54 +26,58 @@ function PointDataLayer({ layer }: { layer: PointDataLayerProps }) {
   const dispatch = useDispatch();
 
   const { data } = layerData || {};
+  const { features } = data || {};
   const { t } = useSafeTranslation();
   useEffect(() => {
-    if (!data) {
+    if (!features) {
```
ericboucher, line 32: why the switch to testing on features vs data?

ericboucher (approved):
Nice work! Thanks @tdlinh2712

src/config/types.ts
```diff
@@ -457,3 +463,8 @@ export enum DownloadFormat {
   CSV,
   JSON,
 }
+
+type AdminLevelDisplayType = {
+  adminCode: string;
+  dataField: string;
```
ericboucher, line 469: No more dataField I think, no?

src/components/MapView/Layers/PointDataLayer/index.tsx
tdlinh2712: because i changed the type of `PointLayerData` to `{features: PointData[] }`, and when `AdminLayerDisplayed` option is on, it will return type `AdminDataLayerData`. Both these types store the location points in `features`, hence we are checking for features instead of data

src/config/types.ts
tdlinh2712: fixed

PR #428, opened 2022-03-29
Add Swagger UI to the API

vikhyatsharma:
Auto generated Swagger documentation.

The doc is available at the root of the API. On localhost, visit `http://localhost`.


vikhyatsharma:
Screenshots of the Swagger UI

Capture3
Capture4
Capture5
Capture6
Capture8
Capture
Capture2

ericboucher:
@vikhyatsharma looks like we're missing a few docstrings. `api-lint` can be a bit annoying sometimes as it shows you a type of error and then exists early, so you will probably have to solve issues iteratively.

ericboucher:
@vikhyatsharma make sure to update the screenshots with your latest changes :)

vikhyatsharma:
image (3)
image (2)

api-flask/Dockerfile
```diff
@@ -7,3 +7,4 @@ COPY requirements.txt .
 RUN pip3 install --upgrade pip
 RUN pip3 install -r requirements.txt
 RUN pip3 install greenlet==0.4.17
+RUN pip3 install flask-restx
```
ericboucher, line 10: This should be added to `requirements.txt` instead

api-flask/app/main.py
```diff
@@ -1,290 +1,340 @@
...
-
-    # TODO - Properly encode before returning. Mongolian characters are returned as hex.
-    return jsonify(features)
-
-
-if __name__ == '__main__' and getenv('FLASK_ENV') == 'development':
-    PORT = int(getenv('PORT', 80))
-    # Only for debugging while developing
-    app.run(host='0.0.0.0', debug=True, port=PORT)
+"""Flask API for geospatial utils."""
```
ericboucher, line 1: It is very odd that the whole file is seen as new. Can you try to fix this git problem?

vikhyatsharma:
> Screenshots of the Swagger UI
> 
> Capture3
> 
> Capture4
> 
> Capture5
> 
> Capture6
> 
> Capture8
> 
> Capture
> 
> Capture2
image (5)
image (4)

api-flask/Dockerfile
vikhyatsharma: At first, I did try to add this in the requirement.txt but for some reason, it didn't work. I can try one more time if you would like :)

ericboucher: The you will need to rebuild the docker. Let me know if you can't get it to work and I can have a look

ericboucher (approved):

ericboucher:
@vikhyatsharma let's setup example values for `/stats` and not just the type. You can use the values from the `/demo` code.

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
ericboucher, line 1: @vikhyatsharma we are using yarn for this repo. Please delete the extra files that you commited. And add the raster cache to the `api-flask/.gitignore`

api-flask/app/statsdata.txt
```diff
@@ -0,0 +1,5 @@
+{
```
ericboucher, line 1: Instead of these test files, let's create this sample data in a sample_requests.py file. You can simplify the alertdata a lot by just having a very simple GeoJSON, it doesn't need to be as big as the one you have there. It just needs to be a valid shape.

src/config/myanmar/prism.json
```diff
@@ -90,4 +90,4 @@
       ]
     }
   }
-}
+}
```
ericboucher, line 93: Please re-add the newline here and delete `package-lock.json`

api-flask/app/main.py
```diff
@@ -153,135 +181,149 @@ def stats():
...
-        alert_db.write(alert)
-
-    except rasterio.errors.RasterioError as e:
-        logger.error(e)
-        raise e
-
-    return Response(response='Success', status=200)
 
+@api.route('/kobo/forms')
+class Get_kobo_forms(Resource):
```
ericboucher, line 186: Let's have class names in CapWords -> https://peps.python.org/pep-0008/#class-names

api-flask/app/sample_requests.py
```diff
@@ -0,0 +1,116 @@
...
+                            ]
+                        ]      
+                    ]
+                }
+            }
+        ]
+    },
+    "email": "vik@gmail.com",
+    "prism_url": "https://prism-mongolia.org"
+}
```
ericboucher, line 116: "/n"

ericboucher: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

api-flask/app/main.py
```diff
@@ -67,84 +91,88 @@ def _calculate_stats(zones,
...
+            stats=['min', 'max', 'mean', 'median', 'sum', 'std'],
+            prefix='stats_',
+            group_by=group_by,
+            geojson_out=geojson_out,
+            wfs_response=wfs_response,
+            intersect_comparison=intersect_comparison
+        )
 
-    return jsonify(features)
+        return features
```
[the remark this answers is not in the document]
vikhyatsharma: Flask RESTX takes care of it automatically. Whatever iterator you send, it will first convert it into a JSON object and then send it across so we don't have to explicitly call `jsonify` :)

api-flask/app/main.py
```diff
@@ -44,6 +49,25 @@
...
+
+stats_model = api.model('Stats', stats_dic)
+
+alerts_dic = {
+    'email': fields.String(alert_data['email']),
+    'prism_url': fields.String(alert_data['prism_url']),
+    'alert_name': fields.String(alert_data['alert_name']),
+    'alert_config': fields.Raw(alert_data['alert_config']),
+    'zones': fields.Raw(alert_data['zones']),
+}
```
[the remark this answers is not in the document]
vikhyatsharma: The dictionary requires `fields` which get imported from flask-restx, so if we move it to "sample_requests" we will have to import the library twice.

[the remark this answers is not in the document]
ericboucher: I think that's fine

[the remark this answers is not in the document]
ericboucher: And you only import "fields"

.gitignore
```diff
@@ -25,6 +25,9 @@ yarn-error.log*
 .env.test.local
 .env.production.local
 
+# Package-lock
+.package-lock.json
```
ericboucher, line 29: This should not exist, you need to delete the file completely

ericboucher (approved):

PR #431, opened 2022-03-30
Polygon analysis

DanielJDufour:
**Final Update:**

This PR adds polygon intersection analysis support to PRISM, so you can run analysis on vector-based hazard layers, specifically tropical storms and earthquakes (shakemap).  The following notable changes:
- icons next to hazard layers indicating whether the layer is based on a raster or polygon (thanks to @ericboucher )
- adding the [zonal](https://www.npmjs.com/package/zonal) dependency, which does zonal statistics between vector layers
- zonal uses the terminology zones and classes, in our case, zones is the admin boundaries and classes is the hazards layer
- by default, zonal will look for a "label" property in the class/hazard layer and use that as the value on which to run zonal statistics.  however, if the layer happens not to use label, you'll have to add a zonal configuration option to the layer json:
```js
   "zonal": {
        // use "wind_speed" instead of "label"
        "class_properties": ["wind_speed"]
    }
```
- zonal statistics run inside of a web worker
- zonal statistics speed is determined by the total number of vertices in the zones (admin boundaries) and classes (hazards)
- zonal statistics support analysis over a time range, in which case area that was hit by a category 4 hurricane twice won't be double counted
- created a Common folder with a new SimpleDropdown component in it, which we can reuse when creating more dropdowns in the future
- great translation supported (thanks to @tdlinh2712)
- lots more work listed here (by @ericboucher and @tdlinh2712): https://github.com/WFP-VAM/prism-frontend/pull/431/commits

~~Added polygon intersection analysis support to Prism.  Supersedes https://github.com/WFP-VAM/prism-frontend/pull/405.  Most everything is dynamic and will work for any country, except the admin-utils.tsx file which I need to update to be dynamic and not hard-coded.~~


DanielJDufour:
Saw a CI job fail. Quick read is that I just need to update a test snapshot.

ericboucher:
@DanielJDufour can you recommend a country and a step by step guide to run locally for testing? Thanks!

DanielJDufour:
Just pushed an update to admin-utils, so every country should be supported.

Here's the test case:
- Select a country (you can start with Myanmar)
- Click Run Analysis Dropdown (in top left)
- Click Choose hazard layer
- Click a vector hazard layer (like "Tropical Storms - Wind buffers")
- Select an Admin Level in the dropdown
- Select a start and end date
- Click Run Analysis Button
- Toggle map view and table view
- Click download under the table
- Click download in legend to download data as JSON (I think it's just a JSON array, not a GeoJSON, so we'll have to update that at some point)


Also, I think I caught an issue with the legend rendering. It might be rounding the breaks.  I'll try to fix that.
Screen Shot 2022-03-31 at 5 58 37 PM

src/utils/admin-utils.ts
```diff
@@ -0,0 +1,26 @@
+import { FeatureCollection } from 'geojson';
+
+import {
+  getBoundaryLayerSingleton,
+  getDisplayBoundaryLayers,
+} from '../config/utils';
+
+export function getAdminLayerURL(adminLevel = 1): string {
+  const lyrs = getDisplayBoundaryLayers();
```
ericboucher, line 9: layers?

src/utils/analysis-utils.ts
```diff
@@ -434,8 +435,31 @@ export async function loadFeaturesClientSide(
...
+          if (typeof value === 'number') {
+            return value.toLocaleString('en-US');
+          }
+          if (value === null) {
+            return 'null';
+          }
+          if (value === undefined) {
+            return 'null';
+          }
+          return value;
```
ericboucher, line 457: this could simply be

`return isNil(value) ? 'null' : value`

src/components/MapView/Analyser/index.tsx
```diff
@@ -107,17 +134,22 @@ function Analyser({ extent, classes }: AnalyserProps) {
...
-      setSelectedDate(dates[dates.length - 1]);
+      const lastDate = availableHazardDates[
+        availableHazardDates.length - 1
+      ].getTime();
+      setSelectedDate(lastDate);
+      setStartDate(lastDate);
+      setEndDate(lastDate);
     }
-  }, [availableDates, hazardLayerId]);
+  }, [availableDates, availableHazardDates?.toString(), hazardLayerId]);
```
ericboucher, line 152: It seems very odd to have `?toString()` in the dependency, what are you trying to achieve?

src/components/MapView/Analyser/index.tsx
```diff
@@ -215,7 +247,12 @@ function Analyser({ extent, classes }: AnalyserProps) {
 
   const clearAnalysis = () => {
     dispatch(clearAnalysisResult());
-    deactivateUniqueBoundary();
+
+    // vector hazards already use admin boundaries
+    // so no need for a unique boundary
+    if (hazardDataType !== GeometryType.Polygon) {
```
ericboucher, line 253: This seems to do the opposite of the what the docstring says?

ericboucher: Maybe group with what's happening on line `272`?

public/images/icon_polygon.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 302 301" fill="#378e00">
```
ericboucher, line 1: Make these one line svgs

src/components/MapView/Analyser/index.tsx
```diff
@@ -95,6 +104,24 @@ function Analyser({ extent, classes }: AnalyserProps) {
...
+    hazardLayerId === undefined
+      ? null
+      : (LayerDefinitions[hazardLayerId] as WMSLayerProps).geometry ||
+        RasterType.Raster;
+  const hazardServerLayerName = hazardLayerId
+    ? (LayerDefinitions[hazardLayerId] as WMSLayerProps).serverLayerName
+    : undefined;
+  const availableHazardDates = hazardServerLayerName
+    ? availableDates[hazardServerLayerName]?.map(d => new Date(d)) || []
+    : undefined;
```
ericboucher, line 123: Let's try to reuse / augment `getPossibleDatesForLayer`

src/context/analysisResultStateSlice.ts
```diff
@@ -386,6 +403,65 @@ export const requestAndStoreAnalysis = createAsyncThunk<
...
+    lyr: hazardLayer,
+    startDate,
+    endDate,
+  });
+
+  const result = calculate({
+    zones,
+    zone_properties: [getAdminNameProperty(adminLevel)],
+    classes,
+    class_properties: hazardLayer?.zonal?.class_properties, // eslint-disable-line camelcase
```
ericboucher, line 425: What happens if there is no class_properties?

src/utils/admin-utils.ts
```diff
@@ -0,0 +1,26 @@
...
+  );
+  return adminLayer ? adminLayer.path : lyrs[0].path;
+}
+
+export function getAdminNameProperty(adminLevel: number = 1): string {
+  const { adminLevelNames } = getBoundaryLayerSingleton();
+  return adminLevelNames[adminLevel - 1];
+}
+
+export function fetchAdminLayerGeoJSON(
```
ericboucher, line 21: I think we can probably simplify this logic a lot. Why do we need to re-fetch the layers?

Most likely the layers json are already accessible.

We also need to consider the case where we only have one file and the adminLevelName is what decides how we want to group the results.

Thinking about it a bit more, we could probably always run the analyis on the getBoundaryLayerSingleton() and then use the adminBoundary level after the fact to group results.

src/utils/analysis-utils.ts
```diff
@@ -434,8 +435,31 @@ export async function loadFeaturesClientSide(
...
 
 export function getAnalysisTableColumns(
-  analysisResult: BaselineLayerResult,
+  analysisResult: BaselineLayerResult | PolygonAnalysisResult,
 ): Column[] {
+  if ('tableColumns' in analysisResult) {
+    return analysisResult.tableColumns.map(column => {
+      return {
+        id: column as any,
+        // remove prefix from column labels
```
ericboucher, line 444: This should be handled by functions specific to zonal

src/utils/wfs-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+): Promise<FeatureCollection> {
+  const { lyr, startDate, endDate } = options;
+
+  if (lyr.type !== 'wms') {
+    throw Error('unexpected layer type');
+  }
+
+  const wfsServerURL = `${lyr.baseUrl}/wfs`;
+
+  const params = new URLSearchParams();
```
ericboucher, line 22: Let's mimick how we do things in `server-utils` a bit more. This function could probably live there as well.

src/components/MapView/Analyser/index.tsx
```diff
@@ -13,6 +13,7 @@ import {
   FormControlLabel,
```
ericboucher, line 13: This file is clearly getting out hand, let's add a TODO to split it into different files per type of analysis

src/components/MapView/Analyser/index.tsx
ericboucher: And/or 278 which already has a "deactivateUniqueBoundary();"

src/utils/analysis-utils.ts
```diff
@@ -648,3 +676,59 @@ export class BaselineLayerResult {
...
+  ) {
+    this.featureCollection = featureCollection;
+    this.tableData = tableData;
+    this.tableColumns = tableColumns;
+    this.statistic = statistic;
+    this.threshold = threshold;
+    this.adminLevel = adminLevel;
+
+    // color breaks from https://colorbrewer2.org/#type=sequential&scheme=Reds&n=5
+    this.legend = [
```
ericboucher, line 710: Why not use % instead? -> 20, 40, ...

ericboucher: (this would solve the legend issue)

src/components/MapView/Analyser/index.tsx
DanielJDufour: I used it because availableHazardDates is being created on each render, deriving from availableDates and hazardLayerId (which are both stored in state).  When I passed the array object of availableHazardDates into a useEffect dependency array, it automatically triggers a new call to the useEffect callback because React thinks the dependencies changed.  Although the values in availableHazardDates are in effect the same, React thinks they are different because useEffect is using a "shallow comparison".  I evaluated a few different options.

1) store availableHazardDates in a useState.  This isn't a bad solution, but each call to useState triggers a new render cycle, so I figured it was best to save one render cycle by deriving it from the state of availableDates and hazardLayerid.
2) implement [useDeepCompareEffect](https://github.com/kentcdodds/use-deep-compare-effect).  I've used this library successfully before and the maintainer is reliable, but I felt it wasn't necessary to add a dependency just for a small purpose.  It might be worth considering in the future though.
3) I could have also used JSON.stringify instead of toString.  JSON.stringify is traditionally thought of as kinda slow, so I figured toString is faster.  toString sometimes stringify objects poorly, but I checked it and it seems to work for dates well.  It just adds a comma between the dates.

Happy to discuss alternatives, too.

src/context/analysisResultStateSlice.ts
DanielJDufour: if no class_properties are provided, zonal will give an id to each hazard feature in the order that they appear in the hazardLayer and use that as the class property.

src/utils/analysis-utils.ts
DanielJDufour: if I understand correctly, I should create a new function called something like `getPolygonAnalysisTableColumns`?

src/utils/analysis-utils.ts
DanielJDufour: awesome. will do

public/images/icon_polygon.svg
DanielJDufour: done

src/utils/analysis-utils.ts
ericboucher: No, I think this could simply be moved inside `requestAndStorePolygonAnalysis` for now

src/utils/analysis-utils.ts
DanielJDufour: confirming it works
Screen Shot 2022-03-31 at 10 43 18 PM

src/components/MapView/Analyser/index.tsx
DanielJDufour: done

DanielJDufour: good idea.

src/utils/analysis-utils.ts
DanielJDufour: ah. okay. will do.

DanielJDufour: done

src/utils/admin-utils.ts
DanielJDufour: done

src/utils/admin-utils.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function getAdminLayerURL(adminLevel = 1): string {
+  const layers = getDisplayBoundaryLayers();
+  const adminLayer = layers.find(
+    layer => layer.adminLevelNames.length === adminLevel,
+  );
+  return adminLayer ? adminLayer.path : layers[0].path;
+}
+
+export function getAdminNameProperty(adminLevel: number = 1): string {
```
Ry-DS, line 16: This as well looks like it can use the new AdminLevel type

src/utils/analysis-utils.ts
```diff
@@ -434,8 +436,25 @@ export async function loadFeaturesClientSide(
...
 export function getAnalysisTableColumns(
-  analysisResult: BaselineLayerResult,
+  analysisResult: BaselineLayerResult | PolygonAnalysisResult,
 ): Column[] {
+  if ('tableColumns' in analysisResult) {
+    return analysisResult.tableColumns.map(column => {
+      return {
+        id: column as any,
+        // remove prefix from column labels
+        // example: replace "stat:area" with "area"
```
Ry-DS, line 446: thanks for describing the regex! :fire:

src/utils/analysis-utils.ts
```diff
@@ -434,8 +436,25 @@ export async function loadFeaturesClientSide(
 }
 
 export function getAnalysisTableColumns(
-  analysisResult: BaselineLayerResult,
+  analysisResult: BaselineLayerResult | PolygonAnalysisResult,
 ): Column[] {
+  if ('tableColumns' in analysisResult) {
+    return analysisResult.tableColumns.map(column => {
+      return {
+        id: column as any,
```
Ry-DS, line 444: avoid any. Can't we fix up the underlying type to type columns properly?

src/utils/analysis-utils.ts
```diff
@@ -434,8 +436,25 @@ export async function loadFeaturesClientSide(
...
+        // remove prefix from column labels
+        // example: replace "stat:area" with "area"
+        label: column.replace(/^[a-z]+:/i, ''),
+        format: (value: number | string) => {
+          if (typeof value === 'number') {
+            return value.toLocaleString('en-US');
+          }
+          return isNil(value) ? 'null' : value;
```
Ry-DS, line 452: putting null in string here also feels dangerous. If this is something we're showing to the user, I'd prefer something like `No value` since to a computer parsing `'null'` could have unintended side-effects esp in JS.

src/utils/analysis-utils.ts
```diff
@@ -462,7 +481,9 @@ export function getAnalysisTableColumns(
   ];
 }
 
-export function downloadCSVFromTableData(analysisResult: BaselineLayerResult) {
+export function downloadCSVFromTableData(
+  analysisResult: BaselineLayerResult | PolygonAnalysisResult,
```
Ry-DS, line 485: This union is done quite often, consider making a type? This will allow us to search usages of this union better, and also better document why the union exists now.

src/utils/analysis-utils.ts
```diff
@@ -648,3 +671,59 @@ export class BaselineLayerResult {
...
+    this.threshold = threshold;
+    this.adminLevel = adminLevel;
+
+    // color breaks from https://colorbrewer2.org/#type=sequential&scheme=Reds&n=5
+    this.legend = [
+      { label: '20%', value: 0.2, color: '#fee5d9' },
+      { label: '40%', value: 0.4, color: '#fcae91' },
+      { label: '60%', value: 0.6, color: '#fb6a4a' },
+      { label: '80%', value: 0.8, color: '#de2d26' },
+      { label: '100%', value: 1, color: '#a50f15' },
```
Ry-DS, line 710: consider commenting what all these colors are, or use `hsl()` since that's easier for humans to read. Or even try using MUI colors (MUI library and theme comes with a lot of pre-defined colors)

src/utils/analysis-utils.ts
```diff
@@ -648,3 +671,59 @@ export class BaselineLayerResult {
       : `${this.getHazardLayer().title} (${this.statistic})`;
   }
 }
+
+export class PolygonAnalysisResult {
```
Ry-DS, line 675: Quite interesting using OOP here (not really done elsewhere as this project is mostly functional) but all for it if it gets the job done

src/utils/wfs-utils.ts
```diff
@@ -0,0 +1,37 @@
+import { FeatureCollection } from 'geojson';
+
+import { WMSLayerProps } from '../config/types';
+
+interface propTypes {
+  lyr: WMSLayerProps;
+  startDate: any;
+  endDate: any;
```
Ry-DS, line 8: avoid any

Ry-DS (commented):
Nice! Quite a lot here, mostly around TS issues and following a more immutable standard. Let me know if you have any questions

src/components/MapView/Analyser/index.tsx
```diff
@@ -95,6 +103,24 @@ function Analyser({ extent, classes }: AnalyserProps) {
...
   const [thresholdError, setThresholdError] = useState<string | null>(null);
 
+  // for polygon intersection analysis
+  const [adminLevel, setAdminLevel] = useState<AdminLevelType>(1);
+  const [startDate, setStartDate] = useState<number | null>(null);
+  const [endDate, setEndDate] = useState<number | null>(null);
+
+  // get variables derived from state
+  const hazardDataType: HazardDataType | null =
+    hazardLayerId === undefined
```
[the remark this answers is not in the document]
DanielJDufour: great catch!  I fixed it.  I also noticed I was repeating `(LayerDefinitions[hazardLayerId] as WMSLayerProps)`, so I just reused selectedHazardLayer.

src/components/MapView/Layers/AnalysisLayer/index.tsx
```diff
@@ -62,10 +63,14 @@ function AnalysisLayer() {
...
-      ? (analysisData.statistic as string)
-      : 'data';
+  let property;
+  if (analysisData instanceof ExposedPopulationResult) {
+    property = analysisData.statistic as string; // eslint-disable-line fp/no-mutation
+  } else if (analysisData instanceof PolygonAnalysisResult) {
+    property = 'zonal:stat:percentage'; // eslint-disable-line fp/no-mutation
+  } else {
+    property = 'data'; // eslint-disable-line fp/no-mutation
+  }
```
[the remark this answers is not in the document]
DanielJDufour: done

src/components/MapView/Layers/LayerDropdown.tsx
```diff
@@ -29,7 +29,8 @@ function LayerDropdown({
       ...category,
       layers: category.layers.filter(layer =>
         layer.type === 'wms'
-          ? layer.type === type && !layer.geometry
+          ? layer.type === type &&
+            [undefined, 'polygon'].includes(layer.geometry)
```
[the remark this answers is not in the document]
DanielJDufour: good question.  It confused me at first as well.  WMS tiles technically still don't.  I don't know the history, but geometry was being used to indicate the underlying geometry which is accessible via WFS.  I suspect it has something to do with preventing some duplication.  I originally tried the approach of creating a WFS alternative for each WMS-related class /   function, but found it to be unmanageable.  It required a lot less code change to just tack on to the WMSLayerProp and other types/classes.  In the future, I'm definitely open to discussing a clearer way of organizing the layer data in a way that pulls the WFS and WMS info into their own separate objects while reducing duplication.

src/context/analysisResultStateSlice.ts
```diff
@@ -386,6 +403,65 @@ export const requestAndStoreAnalysis = createAsyncThunk<
...
+    row['stat:area'] = Math.round(
+      convertArea(row['stat:area'], 'meters', 'kilometers'),
+    );
+
+    // preformat null values as "null"
+    // so they display in the table
+    Object.keys(row).forEach((key: string) => {
+      const v: any = row[key];
+      if (v === null) {
+        row[key] = 'null'; // eslint-disable-line no-param-reassign
```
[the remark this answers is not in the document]
ericboucher: I think the table functions already display something like "no data" actually

src/utils/wfs-utils.ts
DanielJDufour: done. changed to number

src/utils/wfs-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export function fetchWMSLayerAsGeoJSON(
+  options: propTypes,
+): Promise<FeatureCollection> {
+  const { lyr, startDate, endDate } = options;
+
+  if (lyr.type !== 'wms') {
+    throw Error('unexpected layer type');
```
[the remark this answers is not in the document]
DanielJDufour: I think eslint prefers string templating, so I'm doing this in another commit, but the essence of your suggestion is there.  Good idea.

package.json
```diff
@@ -48,8 +48,10 @@
     "@turf/bbox": "^6.5.0",
     "@turf/bbox-polygon": "^6.5.0",
     "@turf/boolean-point-in-polygon": "^6.5.0",
-    "@types/colormap": "^2.3.1",
+    "@turf/clone": "^6.5.0",
+    "@turf/meta": "^6.5.0",
     "@turf/union": "^6.5.0",
+    "@types/colormap": "^2.3.1",
```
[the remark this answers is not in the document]
DanielJDufour: sounds good. done

src/utils/analysis-utils.ts
DanielJDufour: I'm hesitant to use hsl because as far as I can tell, hsl isn't currently being used in the code base and I'm trying to keep with the established practice of using hex.  I'd love to learn more about the hsl function.  Honestly, rgb is the easiest for me to read, but that might be my own personal preference.  I think I can add some comments describing the hsl values, so at least people can read them better.  I'm open to standardizing and potentially refactoring how we do colors, but I think that when we do so, it should happen everywhere and that should be a separate PR.

src/context/analysisResultStateSlice.ts
```diff
@@ -386,6 +403,65 @@ export const requestAndStoreAnalysis = createAsyncThunk<
...
+    zones,
+    zone_properties: [getAdminNameProperty(adminLevel)],
+    classes,
+    class_properties: hazardLayer?.zonal?.class_properties, // eslint-disable-line camelcase
+    preserve_features: true,
+  });
+
+  const tableColumns = result.table.columns;
+
+  const tableRows = result.table.rows.map((row: any, i: number) => {
```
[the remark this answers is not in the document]
DanielJDufour: I created ZonalPolygonRow and replaced any with it.

src/components/MapView/Analyser/index.tsx
```diff
@@ -107,17 +133,22 @@ function Analyser({ extent, classes }: AnalyserProps) {
...
-      setSelectedDate(dates[dates.length - 1]);
+      const lastDate = availableHazardDates[
+        availableHazardDates.length - 1
+      ].getTime();
+      setSelectedDate(lastDate);
+      setStartDate(lastDate);
+      setEndDate(lastDate);
     }
-  }, [availableDates, hazardLayerId]);
+  }, [availableDates, availableHazardDates?.toString(), hazardLayerId]); // eslint-disable-line @typescript-eslint/no-base-to-string, react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
DanielJDufour: I looked at this again and a simple solution was just staring back at me.  I elevated the calculation of the last date outside the use effect.  It's strictly dependent on the values of availableDates and hazardLayerId, like selectedHazardLayer, hazardDataType, and availableHazardDates, so we just need to store availableDates and hazardLayerId in state.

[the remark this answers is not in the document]
DanielJDufour: resolving, but feel free to open another issue if you'd like.  thank you for encouraging me to take a closer look at this.

src/components/MapView/Analyser/index.tsx
DanielJDufour: see comment above, basically calculated lastDate outside useEffect, so could pass shallow number to useEffect instead of array of dates.

src/utils/analysis-utils.ts
DanielJDufour: Hope this answers the question.  Please resolve if so.  Thank you.

src/utils/analysis-utils.ts
Ry-DS: Yup thats fair. See if you can use MUI colors at least?
e.g:
```ts
import { red } from '@material-ui/core/colors';
//...
color: red['50'],
// ...
```

src/utils/analysis-utils.ts
Ry-DS: Yup! No need to change any of this I think it's fine, just curious. Thanks for the rational tho, I'll resolve this.

src/utils/wfs-utils.ts
DanielJDufour: done with https://github.com/WFP-VAM/prism-frontend/pull/431/commits/844b6675ee6d13bff9080ede9232caf45a73f2c4

DanielJDufour: resolving, but please reopen if there's an issue with it

src/utils/admin-utils.ts
DanielJDufour: fixed with https://github.com/WFP-VAM/prism-frontend/pull/431/commits/c6900a19270979a57f4943f39bc4fec3ad6c87d9

Now, basically pulling the admin boundary data for the zones from state.

src/components/MapView/Layers/LayerDropdown.tsx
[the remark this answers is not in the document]
DanielJDufour: resolving, but let me know if you'd like to talk about it more

src/utils/analysis-utils.ts
DanielJDufour: This was addressed with https://github.com/WFP-VAM/prism-frontend/pull/431/commits/8868d382d2bbf18c0a090b1fe49505d7420c8ae3.  Please resolve if you agree.  Thank you.

src/components/MapView/Layers/AnalysisLayer/index.tsx
```diff
@@ -62,10 +63,16 @@ function AnalysisLayer() {
...
+  const property = (() => {
+    switch (true) {
+      case analysisData instanceof ExposedPopulationResult:
+        return analysisData.statistic as string;
+      case analysisData instanceof PolygonAnalysisResult:
+        return 'zonal:stat:percentage';
+      default:
+        return 'data';
+    }
+  })();
```
Ry-DS, line 75: I noticed the following:
image

Is this true?

DanielJDufour: hey, that's a great catch!  I just fixed it with https://github.com/WFP-VAM/prism-frontend/pull/431/commits/915d7f157b039d841c28a1846a6d4d8c3b00d7fb.  The typescript compiler and VS Code didn't flag that for me.  Are you using another IDE or have a plugin/extension installed?  Thanks!

src/utils/analysis-utils.ts
DanielJDufour: addressed with https://github.com/WFP-VAM/prism-frontend/pull/431/commits/937cf00fd8ad7d52a573708887ffb67c71fb701e.  

Here are all the base colors on mui:
Screen Shot 2022-04-02 at 11 46 21 PM

I went through all the red ones:
Screen Shot 2022-04-02 at 11 46 51 PM

Although there are matches for some of the red ones, the color scale doesn't go as dark as the one by colorbrewer2.org

I googled each color currently being used, trying to find any common term or description for it.  I then added a name (or short description) to the comment for each color.

Resolve if this fixes the issue.  Otherwise, let's keep talking.

Thank you.

src/components/MapView/Layers/AnalysisLayer/index.tsx
Ry-DS: I'm surprised this wasn't caught by TS; isn't there anywhere you try assign `analysisData` to a `PolygonAnalysisResult`? TS should've caught it then. Maybe this is highlighting a deeper issue, since in TS's eyes analysisData could never be a `PolygonAnalysisResult` :eyes:

I'm using WebStorm, which has a bunch of random extra checks which is quite nice in this situation!

src/utils/analysis-utils.ts
Ry-DS: Great! Thanks, sorry if it was a little tiresome getting this resolved, all good now

src/components/MapView/SimpleDropdown/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+type OptionValue = number | string;
+type OptionLabel = string;
+type DropdownOption = [OptionValue, OptionLabel];
+
+interface SimpleDropdownProps {
+  options: DropdownOption[];
+  value: OptionValue;
+  // want to do (val: OptionValue)
+  // but this causes type conflicts when a user knows that val will be an enum
+  onChange: (val: any) => void;
```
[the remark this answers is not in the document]
DanielJDufour: I refactored OptionValue to be a constrained generic instead of a static type and so we can just remove any here and use the React.ChangeEvent type assigned by default.   But I did replace the onChange param type from any to OptionValue.  Here's the commit:

https://github.com/WFP-VAM/prism-frontend/pull/431/commits/9c857a165c8f186868f94e89a1e3df54fc091279

I'm going to leave this one open for your review, because constrained generics are complicated.  Let me know if you have any feedback or questions.

src/components/MapView/SimpleDropdown/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  options,
+  value = '',
+  onChange,
+  ...rest
+}: SimpleDropdownProps) {
+  return (
+    <FormControl {...rest}>
+      <Select
+        value={value}
+        onChange={(e: any) => {
```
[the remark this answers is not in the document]
DanielJDufour: I refactored OptionValue to be a constrained generic instead of a static type and so we can just remove any here and use the React.ChangeEvent type assigned by default.

src/components/MapView/SimpleDropdown/index.tsx
```diff
@@ -0,0 +1,38 @@
+import { FormControl, MenuItem, Select } from '@material-ui/core';
```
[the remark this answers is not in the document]
DanielJDufour: I definitely agree.  I intentionally built it in order to be reused across the application :-) I looked and didn't see a great folder home for it, so I just put it an existing folder that made the most sense to me.  Would perhaps creating a `src/components/Common` folder make sense to you?  It would be a folder for Common components like this that could be shared across the application.  `src/components/Shared` is also an option.

Also, definitely open to your suggestions for other folder names or paths.

src/config/types.ts
```diff
@@ -161,6 +161,29 @@ export enum GeometryType {
...
 
+export enum RasterType {
+  Raster = 'raster',
+}
+
+export type HazardDataType = GeometryType | RasterType;
+
+// not including standard deviation and sum quite yet
+// because we won't be able to re-use the WMS legend
+export enum DisplayZonalStatsEnum {
```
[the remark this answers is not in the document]
DanielJDufour: fair point.  I'll remove the `"Enum"`.  I agree it's not used elsewhere, so I shouldn't arbitrarily break with practice.  In case you are curious, I did it because it allows me to remember if something is a type or enum without having to move my cursor over it in VS Code.

src/context/analysisResultStateSlice.ts
```diff
@@ -181,6 +190,14 @@ export type AnalysisDispatchParams = {
   isExposure: boolean;
 };
 
+export type PolygonAnalysisDispatchParams = {
+  hazardLayer: WMSLayerProps;
+  adminLevel: AdminLevelType;
+  extent: Extent;
+  startDate: ReturnType<Date['getTime']>; // just a hint to developers that we give a date number here, not just any number
+  endDate: ReturnType<Date['getTime']>; // just a hint to developers that we give a date number here, not just any number
```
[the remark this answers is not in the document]
DanielJDufour: done with https://github.com/WFP-VAM/prism-frontend/pull/431/commits/0a189012ca8d44c247e9419ced539e875c87e7f2

src/context/analysisResultStateSlice.ts
```diff
@@ -386,6 +403,65 @@ export const requestAndStoreAnalysis = createAsyncThunk<
...
+  const tableRows = result.table.rows.map((row: any, i: number) => {
+    // eslint-disable-next-line no-param-reassign
+    row['stat:area'] = Math.round(
+      convertArea(row['stat:area'], 'meters', 'kilometers'),
+    );
+
+    // preformat null values as "null"
+    // so they display in the table
+    Object.keys(row).forEach((key: string) => {
+      const v: any = row[key];
```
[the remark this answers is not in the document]
DanielJDufour: fixed with https://github.com/WFP-VAM/prism-frontend/pull/431/commits/4e5fffb56397de06a7bbb90f2cc968c57d2087c0

src/context/analysisResultStateSlice.ts
```diff
@@ -386,6 +403,65 @@ export const requestAndStoreAnalysis = createAsyncThunk<
...
+      if (v === null) {
+        row[key] = 'null'; // eslint-disable-line no-param-reassign
+      }
+    });
+
+    // create key for AnalysisTable
+    // eslint-disable-next-line no-param-reassign
+    row.name = i;
+
+    return row;
```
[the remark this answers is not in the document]
DanielJDufour: this is fixed.

fwiw, performance and memory usage is exactly why I chose a mutable approach, but I supposed I should have added a comment about that.  I did test locally and it doesn't seem that the performance is much different between the two approaches in practice, probably because the data sizes are small enough.  Happy to take this up again if we find performance to be an issue.

src/context/analysisResultStateSlice.ts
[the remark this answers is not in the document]
DanielJDufour: I think fixed with https://github.com/WFP-VAM/prism-frontend/pull/431/commits/228a6068d78f50fda862b7fe3399c064d6c7ec3a

I used getRoundedData for any null values in the data.  In the metadata, I reformatted null classes, to say "no class".
Screen Shot 2022-04-03 at 10 40 03 AM

Let me know what you think.

src/utils/admin-utils.ts
```diff
@@ -0,0 +1,26 @@
+import { FeatureCollection } from 'geojson';
+
+import {
+  getBoundaryLayerSingleton,
+  getDisplayBoundaryLayers,
+} from '../config/utils';
+
+export function getAdminLayerURL(adminLevel = 1): string {
```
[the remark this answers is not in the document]
DanielJDufour: good suggestion.  fixed with https://github.com/WFP-VAM/prism-frontend/pull/431/commits/9ef1104d3392e80ea63ffc36ca87b1bbe25304a5

src/utils/admin-utils.ts
DanielJDufour: fixed with https://github.com/WFP-VAM/prism-frontend/pull/431/commits/9ef1104d3392e80ea63ffc36ca87b1bbe25304a5

src/utils/analysis-utils.ts
DanielJDufour: you are welcome

src/utils/analysis-utils.ts
DanielJDufour: fixed
Screen Shot 2022-04-03 at 11 32 26 AM

src/utils/analysis-utils.ts
DanielJDufour: fixed.  metadata properties that start with "class:" now say "no class".  Data values now use getRoundedData, so show "no data" translated.

src/components/MapView/Layers/AnalysisLayer/index.tsx
DanielJDufour: oh cool.  I should check out WebStorm sometime.

I think I didn't catch it earlier because I reused / updated the functions used by BaselineLayerResult rather than creating functions specific for `analysisResult`, so we never call a function where `analysisResult` needs to be a `PolygonAnalysisResult`.

Lastly, I thought it was interesting that types aren't tracked through the Redux Store.  You can pass something in as PolygonAnalysisResult, but pull it out as BaselineLayerResult.  It's not really a bug, but I'll try to keep that in mind in the future :-)

I'll resolve this now as the bug has been fixed.  Feel free to open a new one if you find any more issues.

ericboucher:
Getting there, a few things:
- The mouse seems to disappear behind the table of results
- Let's remove areas with no overlap (directly through zonal if possible)

DanielJDufour:
Removed features with no overlap via https://github.com/oviohub/zonal/releases/tag/v0.4.0 and https://github.com/WFP-VAM/prism-frontend/pull/431/commits/7c0af96a86f5664ea729b52d204f5172559f68ab

DanielJDufour:
Finished removing "no class" rows from table via https://github.com/oviohub/zonal/releases/tag/v0.5.0 and https://github.com/WFP-VAM/prism-frontend/pull/431/commits/acef3d6bd39ce4f4e1a56fbee696448e7c40fe28

src/config/myanmar/translation.json
```diff
@@ -65,6 +65,18 @@
   "Daytime Land Surface Temperature - 10-day (MODIS)": "សីតុណ្ហភាពផ្ទៃដីពេលថ្ងៃ - 10 ថ្ងៃ (MODIS)",
   "Land Surface Temperature - 10-day Anomaly (MODIS)": "សីតុណ្ហភាពផ្ទៃដី - ភាពមិនប្រក្រតីរយៈពេល 10 ថ្ងៃ (MODIS)",
   "Land Surface Temperature - 10-day Amplitude (MODIS)": "សីតុណ្ហភាពផ្ទៃដី - អំព្លីទីត 10 ថ្ងៃ (MODIS)",
+  "Potential flooding (Landsat-8)": "ទឹកជំនន់ដែលអាចកើតមាន (Landsat-8)",
```
ericboucher, line 68: @tdlinh2712 just a quick note that this is in Khmer, which is not the language used in Myanmar. It's ok for now for tetsing puposes.

@wadhwamatic are there any plans to get the translation for Myanmar?

wadhwamatic: @ericboucher - no immediate plans for myanmar translation. khmer is ok for testing. we'll remove it before merging

DanielJDufour:
refactored and reused generateTableFromApiData, so cursor shouldn't be hidden behind table rows anymore and clicking on a table row will open a popup.  I still need to refactor popup to display class information for each zone.

wadhwamatic:
Great to see this progress @DanielJDufour. I tested analysis of tropical storms in Myanmar and found a few issues. My go to test date is May 21, 2016 since its a storm that made landfall and with more than one wind speed category.

1. The end date is not inclusive. I was expecting that by running an analysis with the same start and end date, I'd get a result but that's not the case
2. If I choose admin 1, the resulting map still shows admin 3 boundaries. In the previous version of run analysis, we were able to use the boundary file based on the baseline layer a user input. Can the same logic be applied here?
3. Admin 3 tropical storm analysis fails. See screenshot below
4. It would be preferable to have separate columns for each category, and a total across categories by admin area. See an example from the exposure analysis
5. The labels for admin levels are not quite right. Can we make this a configuration detail so that each deployment uses the correct names for admin levels?

> Issue 1: End date is not inclusive
Screen Shot 2022-04-05 at 15 38 08

> Issue 2: Boundary layer shown is for admin 3 but the analysis is based on admin 1
Screen Shot 2022-04-05 at 15 39 41

> Issue 3: Admin 3 analysis fails
Screen Shot 2022-04-05 at 15 40 42

> Issue 4: Each category should have a separate column as we've done with exposure analysis
Screen Shot 2022-04-05 at 15 57 38

DanielJDufour:
fyi, I'm looking at the end date not being inclusive issue right now.

DanielJDufour:
issue with end date not being inclusive has been fixed with https://github.com/WFP-VAM/prism-frontend/pull/431/commits/1e26f23a424b6434bdc37358134f3acb357754ba

DanielJDufour:
@wadhwamatic ,  do you know where I can get the country-specific admin level names?  I currently know how to access the admin_level_names and admin_level_local_names from the JSON file, but they aren't very human readable.  Myanmar uses
```
{
    "admin_level_names": ["ST", "DT", "TS"],
    "admin_level_local_names": ["mmr_polbnd", "DT_MMR4", "TS_MMR4"],
}
```
Tajikistan uses:
```
    "admin_level_names": ["admin1Name_en", "admin2Name_en"],
    "admin_level_local_names": ["admin1Name_ru", "admin2Name_ru"],
```

Is this something we would want to add to the JSON file configuration if it's not there already?

src/components/MapView/Analyser/index.tsx
DanielJDufour: I believe Eric addressed this with a previous commit

src/components/MapView/SimpleDropdown/index.tsx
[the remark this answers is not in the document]
DanielJDufour: I moved SimpleDropdown to a new Common folder via https://github.com/WFP-VAM/prism-frontend/pull/431/commits/c6ab13d58924846e69d8204edd7cdd9c6c410c30

src/context/analysisResultStateSlice.ts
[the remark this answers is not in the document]
DanielJDufour: this has been superseded by events as we no long show "no class" rows

src/components/MapView/Analyser/index.tsx
```diff
@@ -277,37 +342,67 @@ function Analyser({ extent, classes }: AnalyserProps) {
...
+        hazardLayer: selectedHazardLayer,
+        baselineLayer: selectedBaselineLayer,
+        date: selectedDate,
+        statistic,
+        extent,
+        threshold: {
+          above: parseFloat(aboveThreshold) || undefined,
+          below: parseFloat(belowThreshold) || undefined,
+        },
+        isExposure: false,
```
ericboucher, line 401: Nothing seems to use this flag? Should we remove it from the type?

src/components/MapView/Analyser/index.tsx
```diff
@@ -448,7 +600,13 @@ function Analyser({ extent, classes }: AnalyserProps) {
                   {isTableViewOpen && (
                     <AnalysisTable
                       tableData={analysisResult.tableData}
-                      columns={getAnalysisTableColumns(analysisResult)}
+                      columns={
+                        'tableColumns' in analysisResult
+                          ? // have to add "as any" here because typescript compiler mistakenly thinks
+                            // analysisResult is a union of BaselineLayerResult and PolygonAnalysisResult
+                            (analysisResult as any).tableColumns
```
ericboucher, line 607: And what is it instead? Could you cast to the actual type?

src/components/MapView/Analyser/index.tsx
```diff
@@ -547,6 +706,12 @@ const styles = (theme: Theme) =>
     calendarPopper: {
       zIndex: 3,
     },
+    dateRangePicker: {
```
ericboucher, line 709: Does this already exist for other date pickers?

src/config/tajikistan/layers.json
```diff
@@ -319,6 +319,11 @@
       "id": "wp_pop_cicunadj",
       "key": "label"
     },
+    "zonal": {
+      "class_properties": [
+        "label"
```
ericboucher, line 324: If label is the default, do we even need this in the config?

ericboucher, line 324: If this is the default for all, should we just add it as the default and not need it in the config?

src/components/MapView/Analyser/index.tsx
ericboucher: And from the different places where it is set?

src/config/tajikistan/layers.json
DanielJDufour: good idea. done with https://github.com/WFP-VAM/prism-frontend/pull/431/commits/750abe11ebe142022333b40dae39f891bef602e5

src/components/MapView/Analyser/index.tsx
DanielJDufour: Thanks for catching this.  Fixed with https://github.com/WFP-VAM/prism-frontend/pull/431/commits/40303b2f38eb134d92ce99d1ceb0add53fe279f7

src/components/MapView/Analyser/index.tsx
DanielJDufour: agreed. done with a140d23

src/context/analysisResultStateSlice.ts
```diff
@@ -483,6 +559,37 @@ export const analysisResultSlice = createSlice({
...
+
+    builder.addCase(
+      requestAndStorePolygonAnalysis.fulfilled,
+      (
+        { result, ...rest },
+        { payload }: PayloadAction<PolygonAnalysisResult>,
+      ): AnalysisResultState => ({
+        ...rest,
+        isLoading: false,
+        result: payload as any,
```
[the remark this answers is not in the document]
DanielJDufour: fixed

src/config/myanmar/translation.json
DanielJDufour: can this be resolved now?  Is the plan to just comment out the import of this translation file?

tdlinh2712: @DanielJDufour i think it should be since @ericboucher removed the import of the translation file?

ericboucher (approved):

PR #435, opened 2022-04-05
Remove kobo password from docker-compose

ericboucher:
Remove KOBO password from dc file and load from environment variable instead


wadhwamatic (approved):
works as expected. thanks

PR #436, opened 2022-04-05
Add min and max stats to raster analysis

ericboucher:
Screen Shot 2022-04-05 at 7 26 11 PM


wadhwamatic (approved):
works as expected. thanks!

PR #437, opened 2022-04-06
Fixing groupBy selection for analysis API

ericboucher:
Fixing an issue which prevented layers with a specific boundary to get "groupedBy" properly.

Screen Shot 2022-04-06 at 12 28 32 PM

To reproduce, run the app for Mongolia and run an analysis with the Poverty headcount baseline.

This is also fixing tooltips and layers ordering. Boundary layers should now stay at the top of our stack


wadhwamatic (approved):
tested mongolia and mozambique (multi poverty index) and this works as expected. thanks for this fix eric!

Issue #426, opened 2022-03-29
Look into grouping layers of the same data product but different time intervals / data rounds


wadhwamatic:
We're interested to see how we could reduce the number of layers users have presented to them by making the date interval a second option after choosing a data product. 

For example in Mozambique (https://prism-demo-moz.surge.sh/) instead of having a separate layer for each rainfall amount at 10-day, 1-month, 3-month, etc. we could instead have a single rainfall amount layer which defaults to 10-day, and then gives the user the ability to switch to another time interval. This would help us have less layers appear here:

<img width="533" alt="Screen Shot 2022-03-28 at 21 36 14" src="https://user-images.githubusercontent.com/3343536/160534431-3163ee65-f2c6-488c-9938-39521cab8a40.png">

The grouping of data layers in the UI as was implemented in https://github.com/WFP-VAM/prism-frontend/pull/163 could be reused to solve for this. This should be a UI only change. We don't want to depend on a new attribute from GeoServer like PR 163 does. We simply want to logically group layers. 

See this relevant PR on layer grouping as well: https://github.com/WFP-VAM/prism-frontend/pull/152

wadhwamatic:
Addressed via https://github.com/WFP-VAM/prism-frontend/pull/449

PR #449, opened 2022-04-12
Layer menu grouping to simplify list of menu

hav142:
Based on issue #426 

The idea is to make a group of similar layers into one menu so it's simpler.
For example, using Mozambique's menu like below:

Screenshot 2022-04-12 103804

grouped into:

Screenshot 2022-04-25 125701

then user can select time interval options on each of grouped menus:

image

if we set attribute `activate_all = true` in the prism.json config, the drop-down will not appear and it will activate all layers in group at once 

demo: http://prism-menugroup.surge.sh/


wadhwamatic:
Good start @hav142! I added another set of layers to see how the configuration works and have more layers to test with. There's an issue with text and the opacity slider getting cut off if the 'label' text is too long for the legend area (see screenshot). It would also be good if can allow for configuration of the text 'Layer Option' for the dropdown menu.

@ericboucher had some concerns with potential conflicts with https://github.com/WFP-VAM/prism-frontend/pull/431. Eric - can you let Hafidz what the concern is?

Screen Shot 2022-04-12 at 12 56 22

hav142:
@reed1 @wadhwamatic I fixed above issues on the new commits.
and now we can set the layer option title on the prism.json: 

```
"title": "IPC",
"optionTitle": "IPC Layer Option",
"layers": ... 
```

src/components/MapView/Legends/index.tsx
```diff
@@ -211,6 +230,22 @@ function LegendItem({
     initialOpacity || 0,
   );
 
+  const handleChangeFormInput = (event: any) => {
+    const layerId: LayerKey = event.target.value;
+    const layer = LayerDefinitions[layerId];
+    const ADMIN_LEVEL_DATA_LAYER_KEY = 'admin_level_data';
```
ericboucher, line 236: Just use `admin_level_data` directly below

ericboucher:
Overall, this looks like a good idea to simplify the menu and layer selection. However, I would prefer to have this logic happen in the menu selector directly. Adding this logic to the legend items adds complexity and will be harder to maintain. Indeed, we end up with multiple complex functions spread out in the code that need to understand our layer, layer grouping, and layer selector logic.

Noting that the "group" word is already used to describe a group of layers that are all activated at the same time. We need to find better words to differentiate the concepts more clearly.

wadhwamatic:
Thanks for the input @ericboucher. Good suggestion to move this feature to the menu rather than legend area. I think this will be a more intuitive user experience too. 

For the naming convention, maybe we use 'menu-group' or something along those lines to differentiate this from the current group feature.

hav142:
I was thinking about how the UI will be if we move the menu grouping into the navbar.

option 1: using accordion

image

option 2: using form select

image

Which one do you prefer, or any other suggestions?

ericboucher:
I think the dropdown should work well, but it should be only on the second part if that's easy.

Where you select the layer "Rainfall" and then the selection dropdown has the options "dekad 1-day", "aggregate 1-month", and so on

hav142:
update for the navbar:

- activate a menu

image

- then drop-down option can be selected

image

wadhwamatic:
Thanks @hav142. @ericboucher and I both had a look. We'd like to see the dropdown option implemented differently.  We should be able to select a different time interval prior to toggling the first option. 

So for example, if I want to view rainfall aggregate at 1-month interval, I should be able to click the dropdown next to rainfall aggregate to show all the options available and choose 1-month without first having to toggle on the 10-day aggregate. Hope this is clear.

Layer 1

ericboucher:
@hav142 as a quick note, can you try and keep the PR body (initial comment) up to date with the latest screenshots?

hav142:
Thanks for the feedback @wadhwamatic @ericboucher 
I updated both the behavior and the screenshot

wadhwamatic:
Thanks @hav142, this looks good. I have one additional suggestion - what about having the selection of a sub-group item in the dropdown menu automatically toggle on the layer? I expected when I selected one of those options, that the layer would activate. Instead, I had to select it, and then toggle on the layer.

wadhwamatic:
Thanks for the additional change @hav142. Works well for me. I think it would be useful to change the attribute 'title' to something more clear such as 'layer_group_title'. This will make for easier configuration and troubleshooting. What do you think @ericboucher ?

Please also update your surge deployment with your latest change.

src/config/mozambique/prism.json
```diff
@@ -26,26 +26,41 @@
...
-        "rain_anomaly_1year",
-        "spi_1m",
-        "spi_3m",
-        "spi_6m",
-        "spi_9m",
-        "spi_1y"
+        {
+          "title": "Rainfall Anomaly",
+          "layers": [
+            {"label": "10-day", "id": "rain_anomaly_dekad"},
```
ericboucher, line 46: do we need an `id` in here? or could we use the snake_case of `title + label`?

src/context/mapStateSlice/index.ts
```diff
@@ -75,12 +75,12 @@ export const mapStateSlice = createSlice({
       { payload }: PayloadAction<LayerType>,
     ) => ({
       ...rest,
-      layers: layers.filter(({ id, group }) =>
+      layers: layers.filter(({ id, group }) => {
```
ericboucher, line 78: why this change?

src/components/NavBar/MenuSwitch/SwitchItem/index.tsx
```diff
@@ -0,0 +1,187 @@
...
+    <Box key={layerId} display="flex" mb={1}>
+      <Switch
+        size="small"
+        color="default"
+        checked={selected}
+        onChange={e => toggleLayerValue(activeLayer, e.target.checked)}
+        inputProps={{
+          'aria-label': validatedTitle,
+        }}
+      />{' '}
```
ericboucher, line 158: You could use classes with a leftMargin here instead of a dangling space

src/config/mozambique/prism.json
ericboucher: Nevermind, it points to the layers.json.

Small thing, but can switch the order. id, then label?

src/config/types.ts
```diff
@@ -253,6 +273,9 @@ export class CommonLayerProps {
 
   @optional
   featureInfoProps?: { [key: string]: FeatureInfoProps };
+
+  @optional
+  menuGroup?: MenuGroup;
```
ericboucher, line 278: Add a docstring to explain how this is used

src/context/mapStateSlice/index.ts
hav142: there was a bug for the previous approach and need to update the `removeLayer` function by adding some checker and return statement. seems i forgot to revert the return statement back. fixed it in the latest commit

src/components/MapView/Legends/index.tsx
```diff
@@ -333,6 +335,13 @@ const styles = () =>
       padding: 8,
       width: 180,
     },
+    select: {
+      color: '#333',
```
ericboucher, line 339: Try to use the shared muiTheme when possible

src/config/types.ts
ericboucher: I meant, actually explain ;)

-> and give an example of formatted menuGroup

hav142: is this ok?

image

wadhwamatic:
@hav142 - I tested this feature with translations to make sure we've got each element in the configuration covered when switching languages.  The 'menu_group_title' attribute is correctly translating, but the 'label' for each layer is not. See screenshots below.

Screen Shot 2022-05-05 at 13 29 50

Screen Shot 2022-05-05 at 13 39 25

wadhwamatic:
@hav142 - thanks for the translation fix. This looks good from my perspective. Please resolve the conflict in src/components/NavBar/MenuSwitch/index.tsx. I also see that the latest features / config on master for Mozambique are not in this branch. Please also update that and resolve any new conflicts. Then, over to @ericboucher for his final review.

ericboucher:
@hav142 this is looking good.

Let's actually merge the two grouped layers types into your new implementation

And add use a flag like `activate_all` for the former grouped layers option.

@wadhwamatic what do you think?

wadhwamatic:
> And add use a flag like `activate_all` for the former grouped layers option.

@ericboucher - yes, that sounds like a good solution. thanks!

wadhwamatic:
@hav142  - any updates following @ericboucher 's suggestion to combine this feature with the existing layer grouping feature?

hav142:
@wadhwamatic i removed the `group` attribute from layers.json and merged it into prism.json like screenshot below:

image

@ericboucher should i change the `menuGroup` attribute into `group` to make it simpler since it's been merged?

wadhwamatic:
@hafidz142 - great work to combine these features! A couple of issues:
1. Something has broken when we try to load the latest tropical storm. When I activate the tropical storm layer, the application loads a storm with the date 2011-July-22 (which doesn't seem to exist). Instead, it should have loaded the latest storm, which is on 2022-April-27. 
2. CircleCI test for the API failed with your last commit. Please have a look. https://app.circleci.com/pipelines/github/WFP-VAM/prism-frontend/2493/workflows/ee8b7664-27e8-4ff7-96b7-4b591f53694a/jobs/4676

ericboucher:
@wadhwamatic tests passed after restarting

wadhwamatic:
Great work @hafidz142 . This looks good from my perspective. Tested for Mozambique and Myanmar. 

@ericboucher - gtg?

src/config/types.ts
```diff
@@ -18,12 +19,33 @@ export type LayerType =
...
+ * Check if a string/object is an explicitly defined layer in layers.json
+ * @param layerKey the string/object to check
  */
-export const isLayerKey = (layerKey: string): layerKey is LayerKey =>
-  layerKey in rawLayers;
+export const isLayerKey = (layerKey: string | MenuGroup) => {
+  if (typeof layerKey === 'string') {
+    return layerKey in rawLayers;
+  }
+  if (typeof layerKey === 'object') {
```
ericboucher, line 43: you probably don't need to check the type here since it will be a MenuGroup if not a string

src/utils/useDefaultDate.ts
```diff
@@ -16,9 +19,15 @@ import { DEFAULT_DATE_FORMAT } from './name-utils';
  */
 export function useDefaultDate(
   availableDatesLookupKey: keyof AvailableDates,
-  layerGroup?: GroupDefinition,
+  layerId?: LayerKey,
 ): number | undefined {
   const dispatch = useDispatch();
+  const selectedLayers = useSelector(layersSelector);
+  // check layer without group or main layer in group
+  const isMainLayer = !selectedLayers
```
ericboucher, line 27: some of the filtering is a bit duplcative throughout the code.

maybe expose `isMainLayer` as a function  next to `isLayerKey` for example so that you can re-use it

ericboucher:
Looks great overall, just a few small cleanups :)

ericboucher (approved):

src/config/types.ts
hav142: so, remove `if` and `return false`, make it like this?

```
if (typeof layerKey === 'string') {
   return layerKey in rawLayers;
}
-  if (typeof layerKey === 'object') {
-    const layers = map(layerKey.layers, 'id');
-    return every(layers, layer => layer in rawLayers);
-  }
-  return false;

+  const layers = map(layerKey.layers, 'id');
+  return every(layers, layer => layer in rawLayers);
 };
```

ericboucher: yes exactly, and maybe just add a quick comment :)

ericboucher: I am tempted to simplify this logic by creating the groups in `layers.json` instead of `prism.json`. Maybe something to consider for the future (not in this PR), cc @wadhwamatic

wadhwamatic: from a configuration perspective, I think I prefer to keep it in prism.json. It fits logically there IMO

ericboucher: dunno, I feel like calling `sdi_group` instead in prism.json would be quite convenient too. But we can revisit this later

ericboucher (approved):

ericboucher:
@hav142 can you deploy to surge so that @wadhwamatic can give his final green light? Thanks :)

hav142:
updated by @reed1 in the previous link https://prism-menugroup.surge.sh/

wadhwamatic (approved):
Great work @hav142. Thanks for making those additional changes. I tested locally with Moz and Myanmar. Both working as expected.

PR #450, opened 2022-04-12
v1 of Mozambique translation configuration

wadhwamatic:
This PR adds translation configuration for Mozambique
It also adds a new admin level data layer for population data

Screen Shot 2022-04-27 at 18 16 50


src/config/mozambique/translation.json
```diff
@@ -0,0 +1,96 @@
+{
+    "date_locale": "eu",
```
ericboucher, line 2: I don't have a really good idea on how to do it yet, but I think we should try and organize this file in sections.

// Application

// Layers and Legends

.../...

To make it a bit easier to navigate. What do you think?

wadhwamatic: good point. copying your suggestion from our conversation just now, comments could be added using something like:
“_application”: “section_comment”

will add this in the next iteration

PR #452, opened 2022-04-14
Fix Swagger bug with long line in sample parameters

ericboucher:
There was an issue with long sample request parameters. This PR ensures that they are considered as long strings and no 
"\n" gets inserted.

Before:
Screen Shot 2022-04-14 at 5 15 19 PM


ericboucher:
cc @vikhyatsharma

PR #453, opened 2022-04-15
Display temporal information from remote datasets

### LINKED ISSUE #121: Facilitate time-series data retrieval and visualization
PRISM users need to see trends for an Earth Observation product over time. For example, they would need to see how a precipitation measurement or NDVI has evolved over a season, or year-over-year. A clear example that PRISM needs to replicate can be found on WFP's DataViz platform, where the Seasonal Explorer allows a user to view Admin level data on an indicator over a specified period of time:
https://dataviz.vam.wfp.org/seasonal_explorer/rainfall_vegetation/visualizations

Key issues to look into:
1. With Open Data Cube as the primary data source for PRISM climate hazard data, the underlying data is stored as rasters
2. Can ODC efficiently serve an endpoint by way of the PRISM analysis API which aggregates the raster data over an Admin area, with multiple observations in a given time frame (ex: unique data points for every dekad in a 1-year period)
3. Would pre-processing of raster data to a tabular format provide a more scalable solution?

More issues on the UX in PRISM will follow..

JorgeMartinezG:
image

**How to test:**
- Select a layer which includes temporal information (10-day NDVI or 10-day rainfall estimate (mm))
- Click on any given admin boundary.

**note.** There is a provisional fix that replaces NaN values to null before doing response JSON parsing. Closes #121


wadhwamatic:
Great start @JorgeMartinezG! I see some 'stickiness' with the hover on admin areas to display the tooltip but if I remember correctly, this was resolved when we create a build.

I didn't expect to see a configuration of a new layer (layer name: "dataset") but rather an attribute for the existing layer. We could consider having an attribute in the individual layer, such as 'rfq_dekad' which includes a chart data source. But this is just a thought. Let's discuss next week. I'll also brief @ericboucher on this to get his thoughts. But also a reminder that this is likely a temporary solution until the Data Cube API is built out.

ericboucher:
Can we have a surge link for this?

A few thoughts:
- I agree with Amit that this feels like a config that should live within the layer instead
- I think that the dataViewer files should be as agnostic as possible, maybe receive data in a similar fashion than the Popup data. This way we should be able to re-use it for other use-cases, including river level data

JorgeMartinezG:
@ericboucher configuration is split in this commit: a9bac64213a6cd9682887cae1ce87df9e5fc8795

A surge deployment is found here: https://prism-chart-data.surge.sh/?hazardLayerId=ndvi_dekad_5&date=2022-04-01

cc @wadhwamatic

wadhwamatic:
Thanks @JorgeMartinezG. The new config makes a lot more sense to me. This looks good from my side. Over to @ericboucher for his input.

Reminder to resync with master

src/config/types.ts
```diff
@@ -265,6 +265,23 @@ type LayerStyleProps = {
   line: LinePaint;
 };
 
+type DatasetLevel = {
```
ericboucher, line 268: Can you add some docstring explaining the role of each item?

src/components/DataViewer/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+        url: serverParams.url,
+        serverLayerName: serverParams.layerName,
+        selectedDate,
+      };
+
+      dispatch(loadDataset(datasetParams));
+    }
+  }, [id, boundaryProps, dispatch, serverParams, selectedDate]);
+
+  if (
```
ericboucher, line 55: It seems to odd to have to check all of these. If dataset is not null, surely the other datasetSelector items are not null?

Maybe you could simplify the logic. Otherwise let's use a list and "some" or "all" here to make it more readable

src/components/DataViewer/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+        className={classes.adminButton}
+        onClick={() => dispatch(updateAdminId(adminId))}
+        size="small"
+        color="primary"
+        variant={id === adminId ? 'contained' : 'text'}
+      >
+        {level.name}
+      </Button>
+    ),
+  );
```
ericboucher, line 86: In the designs, it's not clear at first glance what the names are. Can we try and differentiate by level a bit more?

src/components/MapView/Layers/BoundaryLayer/index.tsx
```diff
@@ -28,32 +38,77 @@ function BoundaryLayer({ layer }: { layer: BoundaryLayerProps }) {
...
+  const { fillOnMouseMove, fillOnMouseLeave, fillOnClick } = isPrimaryLayer
     ? {
-        fillOnMouseEnter: (evt: any) => onToggleHover('pointer', evt.target),
+        fillOnMouseMove: onHoverHandler,
         fillOnMouseLeave: (evt: any) => onToggleHover('', evt.target),
         fillOnClick: onClickFunc,
       }
     : {
-        fillOnMouseEnter: undefined,
+        fillOnMouseMove: undefined,
```
ericboucher, line 111: I am not a fan of changing this behavior to onHover. In addition, this breaks how we display the results of analysis layers in popups

ericboucher:
This is starting to look good. However:
- Not a fan of the onHover behavior and find it quite distracting
- This is breaking the normal popup behavior, eg. for analysis results


Instead, maybe we could keep the onClick behavior for the popup and add a button in the popup to load the chart? What do you think?

ericboucher:
We can try that yes but it might be a bit confusing

JorgeMartinezG:
Applied changes @wadhwamatic @ericboucher 

url : https://prism-chart-data.surge.sh/

JorgeMartinezG:
> @JorgeMartinezG the popup should still be triggered when needed, eg. here:
> 
> https://prism-chart-data.surge.sh/?hazardLayerId=ndvi_dekad_5&date=2022-04-01&baselineLayerId=census

Applied change in commit 259c542fb6f8b007728297abfa68c0030248b6e9

wadhwamatic (commented):
This is great, nice work Jorge. I like this solution - we get a single click for the chart display and don't rely on hover. Over to @ericboucher for his input.

src/components/MapView/Layers/BoundaryLayer/index.tsx
```diff
@@ -28,21 +39,67 @@ function BoundaryLayer({ layer }: { layer: BoundaryLayerProps }) {
...
+
+    const adminBoundaryParams: AdminBoundaryParams = {
+      title,
+      boundaryProps,
+      serverParams: { layerName: serverLayerName, url },
+      chartType,
+    };
+
+    dispatch(setBoundaryParams(adminBoundaryParams));
+    dispatch(updateAdminId(lowestLevelId));
```
ericboucher, line 102: If all this logic is for the DataViewer, let's either:
1. package it into a function, maybe moving it to a file more relevant to the DataViewer and import it here
2. Group it and add some comments explaining what we are doing and why.

ericboucher (approved):

PR #457, opened 2022-04-18
Make downloaded CSV file consistent with analysis table

### LINKED ISSUE #402: The CSV file generated when downloading results doesn't have headers for each admin level
After running an analysis, the table view now shows all admin levels, separated by commas. But when we generate a CSV file to download, the headers don't exist for each admin level so we have mismatched columns. See screenshots below. Suggest to add column headers 'admin1', 'admin2', etc.

Steps to replicate:
1. On master branch, REACT_APP_COUNTRY=cambodia yarn start
2. Run analysis: 10-day rainfall + IDPoor, no thresholds, any date
3. At the bottom of the table results, click 'Download' and open the CSV file

Screen Shot 2022-03-11 at 08 32 04

Screen Shot 2022-03-11 at 08 32 52

tdlinh2712:
This PR closes #402.
Deploy link: http://prism-402.surge.sh/

This PR fixes the csv conversion so that the full location name stay together in one column.

To test this PR:
1. Run analysis: 10-day rainfall + IDPoor, no thresholds, any date
2. At the bottom of the table results, click 'Download' and open the CSV file

You can see that now when the language is Khmerr, the csv content is in Khmerr too:
image

Same thing when English is chosen, column "local name" is removed:
image


src/components/MapView/Analyser/index.tsx
```diff
@@ -115,6 +115,7 @@ function Analyser({ extent, classes }: AnalyserProps) {
   const [adminLevel, setAdminLevel] = useState<AdminLevelType>(1);
   const [startDate, setStartDate] = useState<number | null>(null);
   const [endDate, setEndDate] = useState<number | null>(null);
+  // const { t, i18n } = useSafeTranslation();
```
ericboucher, line 118: 👣

src/components/MapView/Analyser/index.tsx
```diff
@@ -404,6 +404,15 @@ function Analyser({ extent, classes }: AnalyserProps) {
     }
   };
 
+  const getTranslatedColumn = (
```
Ry-DS, line 407: Shouldn't this be plural? getTranslatedColumns

src/components/MapView/Analyser/index.tsx
```diff
@@ -404,6 +404,15 @@ function Analyser({ extent, classes }: AnalyserProps) {
...
+    analysis: BaselineLayerResult | PolygonAnalysisResult,
+  ) => {
+    return getAnalysisTableColumns(
+      analysis,
+      !isEnglishLanguageSelected(i18n),
+    ).map(col => ({ ...col, label: t(col.label) }));
+  };
```
Ry-DS, line 414: With all these modifications I would instead make `getAnalysisTableColums` a custom hook (useAnalysisTableColumns) and hide away all of this extra logic there.

tdlinh2712: done

src/utils/analysis-utils.ts
```diff
@@ -670,6 +647,59 @@ export class BaselineLayerResult {
...
+}
+
+export function useAnalysisTableColumns(
+  analysisResult?: AnalysisResult,
+): {
+  translatedColumns: Column[];
+  analysisTableColumns: Column[];
+} {
+  const { t, i18n } = useSafeTranslation();
+  return useMemo(() => {
```
Ry-DS, line 688: What was the motivation for using `useMemo`? Makes sense here, but it likely hinders performance than improves it for simple operations like this (mostly used if fed into `useEffect` or other hooks with dependency arrays, or in special cases where the computation is genuinely expensive and slowdowns are noticed)

Ry-DS (approved):

src/utils/analysis-utils.ts
Ry-DS: yup if you identified its needed to prevent substantial re-rendering then all good to keep

wadhwamatic:
@tdlinh2712 - this looks good, thanks! 

I found one issue, but I'm not sure if it relates to this PR or just wasn't caught before. When I download a CSV of a population exposure analysis, I see an additional column with a header which is not displayed in the table in PRISM. See screenshots below of 1) view in PRISM and 2) local CSV file. I'm wondering if this stems for your earlier work @DanielJDufour. I believe the missing column is the total.

Steps to reproduce:
2. Activate tropical storm layer using date '2022-01-24'
3. Click 'exposure analysis' in legend area
4. Download results as CSV

Screen Shot 2022-05-17 at 09 16 54

Screen Shot 2022-05-17 at 09 22 09

ericboucher:
While we are at it, let's use a better name for the CSV output. I think it should use the query parameters / layer names / selected dates. Instead of the current name which simply uses today's date. wdyt @wadhwamatic ?

wadhwamatic:
> While we are at it, let's use a better name for the CSV output. I think it should use the query parameters / layer names / selected dates. Instead of the current name which simply uses today's date. wdyt @wadhwamatic ?

Yes, agreed @ericboucher

tdlinh2712:
@ericboucher @wadhwamatic can you check the updated deploy link to see if the downloaded file's name is as expected?

ericboucher:
@tdlinh2712 looks good, let's format the date as YYYY_MM_DD to be consistent and gave a fully snake_cased name :)

tdlinh2712:
> @tdlinh2712 looks good, let's format the date as YYYY_MM_DD to be consistent and gave a fully snake_cased name :)

deploy link updated. can you check?

ericboucher:
Looks great, thanks!

PR #461, opened 2022-04-26
frontend EWS data

JorgeMartinezG:
This PR enables PRISM to read from EWS sms API. The information is displayed as a point data layer.
- On a given date, PRISM reads all values for each sensor, using the following request:
```
http://sms.ews1294.info/api/v1/sensors/sensor_event?start=2022-04-23T00:00:00Z&end=2022-04-23T23:59:59Z
```

Then, the mean, max and min statistics are computed for each location, which are included within the geojson locations. 

image

### DIFF
--- a/src/config/types.ts
+++ b/src/config/types.ts
@@ -390,6 +390,10 @@ export class ImpactLayerProps extends CommonLayerProps {
   api?: StatsApi;
 }
 
+export enum PointDataProcessing {
+  EWS = 'ews',
+}
+
 export class PointDataLayerProps extends CommonLayerProps {
   type: 'point_data';
   data: string;
@@ -420,6 +424,9 @@ export class PointDataLayerProps extends CommonLayerProps {
 
   @optional
   boundary?: LayerKey;
+
+  @optional
+  processing?: PointDataProcessing;
 }
 
 export type RequiredKeys<T> = {
@@ -524,4 +531,15 @@ export interface RequestFeatureInfo extends FeatureInfoType {
 
 type AdminLevelDisplayType = {
   adminCode: string;
+};
+
+export type PointData = {
+  lat: number;
+  lon: number;
+  date: number; // in unix time.
+  [key: string]: any;
+};
+
+export type PointLayerData = {
+  features: PointData[];
 };
--- a/src/utils/ews-utils.ts
+++ b/src/utils/ews-utils.ts
@@ -1,0 +1,111 @@
+import { FeatureCollection, Point } from 'geojson';
+import moment from 'moment';
+import { PointData } from '../config/types';
+
+const BASE_URL = 'http://sms.ews1294.info/api/v1';
+
+type statsEWS = {
+  max: number | null;
+  mean: number | null;
+  min: number | null;
+};
+
+/* eslint-disable camelcase */
+type EWSSensorData = {
+  location_id: number;
+  value: [string, number];
+};
+/* eslint-enable camelcase */
+
+export const createEWSDatesArray = (): number[] => {
+  const datesArray = [];
+  const endDate = moment(
+    moment().utcOffset('+0700').format('YYYY-MM-DD'),
+  ).valueOf(); // Asia/Phnom_Penh
+
+  const tempDate = moment('2021-01-01');
+  while (tempDate.valueOf() <= endDate) {
+    // eslint-disable-next-line fp/no-mutating-methods
+    datesArray.push(tempDate.valueOf());
+
+    tempDate.add(1, 'days');
+  }
+
+  return datesArray;
+};
+
+const fetchEWSLocations = async (): Promise<FeatureCollection> => {
+  const url = `${BASE_URL}/location.geojson?type=River`;
+
+  const resp = await fetch(url);
+  const featureCollection: FeatureCollection = await resp.json();
+
+  return featureCollection;
+};
+
+const fetchEWSDataPoints = async (date: number): Promise<EWSSensorData[]> => {
+  const momentDate = moment(date).utc();
+
+  const startDate = momentDate.startOf('day').format();
+  const endDate = momentDate.clone().endOf('day').format();
+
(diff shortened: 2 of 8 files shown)


wadhwamatic:
@JorgeMartinezG - great to see this moving along. A few issues:
1. It seems that we're not yet managing time zones? When I activate the EWS layer today (26-April), it tried to fetch data for 27-April. See screenshot below
2. The symbology for the point data on the map should be based on the river level for a given day (we're using max for now), and the trigger levels  _for each individual sensor_. The trigger levels are found in the locations endpoint. For example, "external_id": "TEPv4.0-001" has the following following: "trigger_levels": { "severe_warning": 5700, "warning": 5200, "watch_level": 4680}. 
3. We should use the same color coding as what is currently on https://www.prism-kh.info/ (Green, Yellow, Red). I believe Eugene had the symbology working correctly 

Screen Shot 2022-04-26 at 14 48 30
Screen Shot 2022-04-26 at 14 48 43

JorgeMartinezG:
@wadhwamatic applied changes. BTW, Eugene uses mean statistic

wadhwamatic:
@JorgeMartinezG - just a reminder - we also need to have the line chart appear when we click on a sensor

wadhwamatic:
This is looking good @JorgeMartinezG . As discussed today, let's utilize the max value to determine map symbology. This will allow us to more accurately show when a river exceeded a higher threshold at any given time during a day

wadhwamatic:
Thanks for the max level fix @JorgeMartinezG . I tested toggling between English and Khmer and consistently face an error (screen just goes blank). Steps to reproduce:
1. Activate EWS layer
2. Click on any data point / river sensor
3. Switch to Khmer

The same things happens if you switch to Khmer first, and then follow the steps above

src/components/DataDrawer/Chart/index.tsx
```diff
@@ -116,6 +121,7 @@ function formatChartData(data: TableData, config: ChartConfig) {
         backgroundColor: colors[i],
         borderColor: colors[i],
         borderWidth: 2,
+        pointRadius: data.EWSConfig ? 0 : undefined,
```
ericboucher, line 124: Add a small comment explaining what's happening?

ericboucher: if you keep it undefined doesnt it get overridden in datasetsWithTresholds already?

src/components/MapView/Layers/PointDataLayer/index.tsx
```diff
@@ -55,6 +65,13 @@ function PointDataLayer({ layer }: { layer: PointDataLayerProps }) {
     dispatch(
       addPopupData(getFeatureInfoPropsData(layer.featureInfoProps || {}, evt)),
     );
+
+    if (layer.processing === PointDataProcessing.EWS && selectedDate) {
+      dispatch(clearDataset());
+
+      const ewsDatasetParams = createEWSDatasetParams(feature.properties);
```
ericboucher, line 72: Should we do `feature?.properties` to be safe?

src/config/types.ts
```diff
@@ -440,6 +444,9 @@ export class PointDataLayerProps extends CommonLayerProps {
 
   @optional
   boundary?: LayerKey;
+
+  @optional
+  processing?: PointDataProcessing;
```
ericboucher, line 449: is it `processing` or `loading`?

ericboucher: I'm a bit confused about this `processing` enum and what it's used for? Can you explain?

ericboucher:
Looks good overall, just a few comments and we should be gtg.

JorgeMartinezG:
Applied feedback in commit f555d3e68acddf81330762eb164beca36a4e60e4

wadhwamatic (approved):
I tested again after the latest code refactoring. Works as expected. Over to @ericboucher for final clearance

wadhwamatic:
I found an error when I activated a second layer (flood monitoring) and then clicked back in time. See screenshot below. I would suggest we disable the ability to show two layers simultaneously which both have a time attribute for now. 

Steps to reproduce the error:
1. Activate EWS layer
2. Navigate to 2022-05-10
3. Activate flood monitoring data
4. Click back in the timeline to go to 2022-05-09

Screen Shot 2022-05-23 at 10 28 11

src/utils/admin-utils.ts
```diff
@@ -67,10 +60,10 @@ export const getChartAdminBoundaryParams = (
   );
 
   const adminBoundaryParams: AdminBoundaryParams = {
-    title,
     boundaryProps,
-    serverParams: { layerName: serverLayerName, url },
-    chartType,
+    serverLayerName,
+    url,
+    id: levels[levels.length - 1].id,
```
ericboucher, line 66: Let's create a util function for this (`levels[levels.length - 1].id`) since we are using it throughout the app and it could be error-prone if we were to change the order somewhere

src/utils/ews-utils.ts
```diff
@@ -0,0 +1,206 @@
+import GeoJSON, { FeatureCollection, Point } from 'geojson';
+import moment from 'moment';
+import { PointData, PointLayerData } from '../config/types';
+
+const BASE_URL = 'http://sms.ews1294.info/api/v1';
```
ericboucher, line 5: Let's pass this as a config parameter in layers.json, no?

src/components/DataViewer/index.tsx
```diff
@@ -18,63 +18,87 @@ import {
...
   useEffect(() => {
-    if (params && selectedDate && id) {
+    if (!params || !selectedDate) {
+      return;
+    }
+
+    if (isAdminBoundary(params)) {
+      const { id, boundaryProps, url, serverLayerName } = params;
       dispatch(
         loadDataset({
```
ericboucher, line 56: I think some of the logic should be moved into `loadDataset` and `loadDataset` should be the one function that behaves differently based on the `loader`. wdyt?

Issue #463, opened 2022-04-27
Add map tile loading indicator


reed1:
We need to show a little loading indicator on the map indicating that the layer tile images are still loading. This is especially useful for raster map with many blank areas (Sentinel-1 Flood for example) with slow internet connection. Our colleagues in Indonesia requested this feature, because when they look at rainfall extreme alerts layer (IDN Prism), they don't know if the empty map means the layer is still loading, or the layer image is loaded but it's just blank image (meaning no alerts). Let me know if there's any other way to tackle the issue

PR #464, opened 2022-04-27
Add layer loading indicator

reed1:
For issue #463 

In short, this adds a loading icon on top right side indicating map tile images are loading.
Demo URL: http://prism-r2.surge.sh/?hazardLayerId=hydra_s1&date=2022-04-23

Try panning / zooming, the loading indicator will show.

Apologize for the quirky code, but this is the approach that has the minimum performance impact I've found so far. I'll try to breakdown the issues:

1.  Originally I wanted to put `isTileLoading` to `mapStateSlice`, but as I tried, the dispatch for
```
setTileLoading: (state, { payload }: PayloadAction<boolean>) => ({
  ...state,
  isTileLoading: payload
}),
```
costs quite significant in profiler, it causes hiccup on UI. So a new state slice is created just for that map tile loading

2. `map.on('dataloading' | 'idle'...` is used manually instead of using MapboxMap props. This is because `react-mapbox-gl` is not exposing `onIdle` events. Maybe it's worth to fork the `react-mapbox-gl` library, replacing all occurrence of 'mapbox-gl' to 'maplibre-gl' and fix the discrepancies on Interfaces. Also with this, newer maplibre version (2.x) can be supported. I can work on this too if needed.

3. Manual state management on `let tileLoading = false`. The `dataloading` events will be emitted quite often from the map and apparently setting loading state from `true` value to `true` will still have noticeable impact, the scroll isn't as smooth. So the workaround is to monitor the loading manually with local variable, then only dispatch change to state when the local variable changes.


wadhwamatic:
@reed1  - why aren't we using the existing loading spinner in the center of the page?

@ericboucher  - your thoughts?

wadhwamatic:
> 2\. `map.on('dataloading' | 'idle'...` is used manually instead of using MapboxMap props. This is because `react-mapbox-gl` is not exposing `onIdle` events. Maybe it's worth to fork the `react-mapbox-gl` library, replacing all occurrence of 'mapbox-gl' to 'maplibre-gl' and fix the discrepancies on Interfaces. Also with this, newer maplibre version (2.x) can be supported. I can work on this too if needed.

@DanielJDufour - could you share your input on this?

wadhwamatic:
See issue https://github.com/WFP-VAM/prism-frontend/issues/463

reed1:
> @reed1 - why aren't we using the existing loading spinner in the center of the page?

@wadhwamatic I think it's for loading larger datasets, for loading tile images, the user should still be able to interact with map (panning & zooming), also to minimize distraction as the loading would trigger quite often

wadhwamatic:
@reed1 - I see your point about it being distracting. Another option to explore is to use the legend area and make it specific to the layer which is loading. I.e., it could be an overlay for that layer's section in the legend.

ericboucher:
@reed1 please update the surge links and add a quick video/gif demo when you get a chance.

I think the approach you took works well for now. We can always improve in the future but I think it is efficient and gets the job down, nice work. Ideally, we would want the info to be at the layer level but I understand that it can be tricky and consume a lot of resources, creating lag...

cc @wadhwamatic

src/components/MapView/TileLoadingIcon/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+function TileLoadingIcon() {
+  const loading = useSelector(loadingSelector);
+  if (loading) {
+    return (
+      <div style={{ paddingTop: '0.3rem', paddingRight: '0.3rem' }}>
+        {loading && <CircularProgress size="2rem" />}
+      </div>
+    );
+  }
+  return <></>;
```
ericboucher, line 15: Can you return null instead of an empty element?

reed1:
@ericboucher @wadhwamatic I updated the approach to track loading state per layer. The loading indicator will now show on the legends

image

you can try it on : http://prism-464.surge.sh/?hazardLayerIds=hydra_s1&date=2022-06-29
thanks to automated deploy url feature..

wadhwamatic (commented):
Nice work @reed1. I tested in a few scenarios and this worked as expected. Agreed with Eric that we may improve on this in the future (I'm thinking in design terms) but this works well for now. 

Over to @ericboucher for final approval before merging

src/components/MapView/Legends/LoadingBar.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const loading = loadingLayerIds.includes(layerId);
+  return (
+    <div className={classes.container}>
+      <div className={classes.lineBase} />
+      {loading && <div className={classes.line} />}
+    </div>
+  );
+}
+
+const styles = () =>
```
ericboucher, line 18: Can we simplify this style a bit now that it's stable? No big deal if not but I feel like some of this css is redundant / unused

ericboucher (approved):
Great work!
Just a minor comment on simplifying the CSS

src/components/MapView/index.tsx
```diff
@@ -427,6 +427,14 @@ function MapView({ classes }: MapViewProps) {
           dispatch(setLoadingLayerIds([...layerIds]));
         }
       }
+    };
+    map.on('dataloading', listener);
```
ericboucher, line 431: Can you explain this new changes?

wadhwamatic: @reed1 - please add some text explaining that last change, then we we can merge. thanks!

reed1:
@ericboucher @wadhwamatic

Deploy URL : http://prism-464.surge.sh/?hazardLayerIds=hydra_s1&date=2022-07-07
Here are the things that's changed.

1. I changed it back to "sourcedataloading" and "sourcedata". I thought some events was not fired, so I used "dataloading" and "data" events with sourceId filtering. But when I checked on maplibre-gl directly, it does fire "sourcedata" and "data" events at the same time. 

2. Minor issue: if we zoom in and immediately zoom out to previous zoom level, the sequence of fired events are:

move - zoom - render - dataloading (isSourceLoaded false) - sourcedataloading (isSourceLoaded false) - moveend - zoomend - render - idle

however there are no events which indicate the layer is now loaded (isSourceLoaded true). So the loading bar will continue forever. The newer version of maplibre-gl has separate sourcedataabort events fired, but also "sourcedata" events will be fired with (isSourceLoaded true) if requests are aborted. But upgrading to newer maplibre-gl is not simple process. I added idle event handler on the bottom of tracking code to workaround this issue. 

3. I removed the `<Grid container direction="column"...` on top Legends component, it seems to have issue with bar width, otherwise I need to hardcode the bar width to 180px, if the container grid is not there, the bar will fit on Legend component correctly

ericboucher:
@reed1 sounds good, can you:
- add more comments in the code
- open an issue to upgrade maplibre and explain what can be changed in your code once it's done

Thanks!

reed1:
@ericboucher sorry which part of it has to be commented ? I think I put several of them already.

As for maplibre, I think the v1.15.2 of maplibre might be the latest that is a drop-in replacement for mapbox-gl. The `sourcedataabort` was introduced in v2.1.1 https://github.com/maplibre/maplibre-gl-js/commit/1e8f75cc66a948eab89274ac139417cd5fa73001 in this PR which has many breaking changes with v1. I don't think it's still a drop-in replacement. I tried to replace build files `(dist/maplibre-gl.js)`, on v1 that is greater than 1.15.2, prism still compiles but has runtime error, on v2 it doesn't compile anymore. If we go with version 2, we might have to refactor the react-mapbox-gl library to sinchronize the interfaces.

ericboucher:
@reed1 ok this PR is go to go, can you open an issue for maplibe? Maybe @DanielJDufour will be able to help if it becomes a blocker

reed1:
@ericboucher I changed a bit on Loading bar, now it will use simple Divider on non-WMS layer. The loading for point data and admin level data is handled differently (using full screen circle progress). Or should we move all loading indicator to legends?

image

I my opinion, there's nothing wrong with the state of maplibre now, the issue is that we want to have a feature that's available on maplibre v2 (newer version) to be backported to v1 that is still a drop-in replacement for mapbox-gl so that it's still compatible with react-mapbox-gl library. Another possibility might be changing the react binding library from react-mapbox-gl (seems not maintained) to react-map-gl (still maintained and supports maplibre gl too: https://visgl.github.io/react-map-gl/docs/get-started/get-started#using-with-a-mapbox-gl-fork ). 

My initial finding for react-map-gl is that they have initialViewState which is less hackish than using `map` reference and call `map.jumpTo` directly to avoid map re-centering on layers change. Other properties seems similar, but they handle GeoJSONLayer a bit different with GeoJSONSource. I don't know if anyone  have looked into that, but seems a viable alternative

ericboucher:
@reed1 thanks for the context, can you actually open a task on GitHub and copy-paste it in there? https://github.com/WFP-VAM/prism-frontend/issues/new Thanks

ericboucher:
@reed1 regarding the design I'd prefer that when it's not loading the bars look the same.
For bar vs full-screen loading, I will let Amit decide @wadhwamatic  but I think am leaning towards using the bar you just developped for most cases.

wadhwamatic:
> @reed1 regarding the design I'd prefer that when it's not loading the bars look the same. For bar vs full-screen loading, I will let Amit decide @wadhwamatic but I think am leaning towards using the bar you just developped for most cases.

Agreed, let's use uniform loading bars

wadhwamatic:
@reed1 & @ericboucher - what's the status of this PR?

ericboucher:
@wadhwamatic waiting for small design changes + using the same bar for other layer loading events

reed1:
@ericboucher @wadhwamatic Now loading indicator is unified in legends:

Deploy URL: http://prism-464.surge.sh/

I'm still using mapStateSlice to track layerData loading, it doesn't incur when people are dragging the map so it's ok to put it there. I'm using the same loadingLayerIds variable naming for both mapStateSlice and mapTileLoadingStateSlice, but refer them as vector and tile on LoadingBar component, hopefully it's not too confusing. Let me know if I need to change anything.

ericboucher:
@reed1 can you resolve the remaining conflict with the master branch?

reed1:
@ericboucher sorry I didn't aware, resolved now

PR #470, opened 2022-05-06
Include translation for chart title

### LINKED ISSUE #469: Layer title in time-series chart isn't translated
Sorry, just noticed this after we merged the feature. The layer title which is displayed in the chart isn't using the translated text. To replicate, use Mozambique on master, change to Portuguese, and choose 10-day rainfall layer. Then click on an admin area. See screenshot below.

Screen Shot 2022-05-05 at 17 33 05

JorgeMartinezG:
closes #469 

image


wadhwamatic (approved):
Looks good, and works as expected. Thanks for the quick fix

PR #471, opened 2022-05-06
Map labels issue

PysarenkoDS-BWT:


PysarenkoDS-BWT:
Layers under Labels

src/components/MapView/Layers/ImpactLayer/index.tsx
```diff
@@ -95,7 +95,7 @@ const ImpactLayer = ({ classes, layer }: ComponentProps) => {
 
   return (
     <GeoJSONLayer
-      before={`layer-${boundaryId}-line`}
+      before={before || `layer-${boundaryId}-line`}
```
ericboucher, line 98: Technically this should not be needed, since `layer-${boundaryId}-line` will be below the labels, so if this layer is below it, it will also be below the layers

ericboucher: And doing this might actually shuffle the existing order for other layers

ericboucher (commented):
Great start!

I think that we can only add the before=firstSymbol for BoundaryLayers. So maybe we could centralize some of the logic in the `BoundaryLayer` function? What do you think?

src/components/MapView/Layers/ImpactLayer/index.tsx
PysarenkoDS-BWT: Yep, I fix it

src/components/MapView/index.tsx
```diff
@@ -409,6 +413,8 @@ function MapView({ classes }: MapViewProps) {
   // Saves a reference to base MapboxGL Map object in case child layers need access beyond the React wrappers.
   // Jump map to center here instead of map initial state to prevent map re-centering on layer changes
   const saveAndJumpMap = (map: Map) => {
+    const { layers } = map.getStyle();
+    setFirstSymbolId(layers?.find(layer => layer.type === 'symbol')?.id);
```
ericboucher, line 417: Let's add a small comment to explain what we are doing here

wadhwamatic:
Looks like we haven't handled the 'run analysis' output layer, and 'exposure analysis' layer. See screenshots below.
> Run analysis:
Screen Shot 2022-05-09 at 15 42 59

> Exposure analysis:
Screen Shot 2022-05-09 at 15 45 04

PysarenkoDS-BWT:
Add comment how you ask
Fix AnalysisLayer

вт, 10 мая 2022 г. в 01:47, Amit Wadhwa ***@***.***>:

ericboucher (approved):

Issue #473, opened 2022-05-19
PRISM build folder not working if served inside a folder


reed1:
Currently the `build` folder has to be the root folder served on web. If we put `build` folder on some other path, it won't work because some files referenced from absolute path.

PR #474, opened 2022-05-19
Allow PRISM build to be served from a subfolder

reed1:
Related issue #473 

This PR changes some URL reference so PRISM build folder does not has to be the root on a simple web server.


wadhwamatic:
Thanks @reed1 . 

@ericboucher - can you have a look? Ridwan found an issue when trying to deploy PRISM in a sub-folder. I was surprised that would break anything. I see this change requires every config to also change how paths to data files are referenced, which seems incorrect to me. Thanks for your input

ericboucher:
@reed1 can you deploy to surge twice? once normally and once in a subfolder? Thanks!

reed1:
@ericboucher 

I deployed the current master branch here within subfolder/build:
http://prism-sub-2.surge.sh/subfolder/build/

It won't work because it's not on the root folder.

also I deployed this branch within subfolder/build:
http://prism-sub-1.surge.sh/subfolder/build/

It will work

ericboucher (approved):

ericboucher:
@reed1 can you deploy to a country that has tables to complete testing? Looks good to me otherwise

reed1:
@ericboucher Sorry I didn't check the tables before, it's not working because it uses PUBLIC_URL prefix. Can I change the fetch URL to just plain text from config? It's also necessary if the URL is external (starts with http).

You can see the changes [here](https://github.com/WFP-VAM/prism-frontend/commit/6b8dc8d91a2f389909f70c9e7b80873e167e6744#diff-ade95bca901ca8dfe493e122ecb43e078b4e63f9642fd4730b6727689400c51d). I also did the same with table download and layer contentPath.

Deployed here: http://prism-sub-3.surge.sh/subfolder/build/

PR #483, opened 2022-05-29
Issue 150/automated surge

tdlinh2712:


.circleci/config.yml
```diff
@@ -1,6 +1,9 @@
 version: 2.1
+on:
+  pull_request:
+
 jobs:
-  build_frontend:
+  build_frontend_and_deploy:
```
ericboucher, line 6: while we are at it, let's splt this into:

- `frontend_tests` with test and lint
- `forntend_build_and_deploy`

do the same split for the api

thanks :)

package.json
```diff
@@ -17,6 +17,7 @@
     "start": "cross-env EXTEND_ESLINT=true react-scripts start",
     "analyze": "yarn build && npx source-map-explorer 'build/static/js/*.js'",
     "build": "cross-env EXTEND_ESLINT=true react-scripts build",
+    "surge": "surge --project ./build --domain \"prism-frontend.surge.sh\"",
```
ericboucher, line 20: rename this surge_deploy maybe?

tdlinh2712:
@ericboucher I managed to make the automated surge deploy link in the format prism-{pr sha1}.surge.sh since environment variable PR_NAME only available for forked PR. Any idea to set the country name automatically based on branch name?

Reference for available env variables in circleci: https://circleci.com/docs/2.0/env-vars/

ericboucher:
@tdlinh2712 the automated deploy needs to write a comment in this PR, similar to what we are doing for Ovio.

It might be easier to do this using GitHub actions. Then you will be able to grab PR labels more easily as well

PR #490, opened 2022-06-09
Multiple layers url

### LINKED ISSUE #403: Conflicts when we have multiple 'hazardLayerId' layers
For the Cambodia deployment, and any future deployments where want to display point and raster data simultaneously, we have conflicts that break layer display and data management. For example, the Cambodia deployment has point data from field assessments (Kobo) and integration of river sensor data from the [EWS-1294 ](http://ews1294.com/)API. PRISM treats these as 'hazardLayers' which seems to then cause issues when we activate a raster hazard layer such as 10-day rainfall estimate. One component is the presence of non-overlapping dates, which we're starting to address here https://github.com/WFP-VAM/prism-frontend/pull/385.

Besides the date issue, we don't seem to have the right process in place to display two types of layers (in the sense of WMS and point data) simultaneously as they are both defined as 'hazardLayers'. Should we consider removing the concept of 'hazardLayers' and 'baselineLayers' as they are seen in the URL parsing feature and simply have a list of layers of any type able to be viewed? We could then find a more specific approach to limiting the number of different layer types shown simultaneously. 

To replicate the issue in the case of Cambodia:
1. Branch: https://github.com/WFP-VAM/prism-frontend/tree/feature/ews-modal
2. Run the backend with docker
3. Toggle both River level sensors status and Potential flooding (Sentinel-1)
4. Navigate back in time using the timeline

See also the details written by @erwagasore here - https://ovio-public.slack.com/archives/C01PFEJ0NJD/p1647012594859929

JorgeMartinezG:
This PR modifies PRISM to include multiple HazardLayers of different types (point, WMS) within the url.

**How to test:** Using Cambodia deployment
- Select a PointDataLayer, i.e. EWS 1294 river level data.
- Select a WMSLayer, 3-month rainfall aggregate (mm).

The url parameter **hazardlayerids** will have both layer ids. In case user selects another layer with the same type of a previously selected, PRISM will do an id replace.

image

closes #403

### DIFF
--- a/src/components/MapView/index.tsx
+++ b/src/components/MapView/index.tsx
@@ -283,7 +283,7 @@ function MapView({ classes }: MapViewProps) {
 
     const dateInt = moment(urlDate).valueOf();
 
-    if (urlDate === null || dateInt === selectedDate) {
+    if (!urlDate || dateInt === selectedDate) {
       return;
     }
 
--- a/src/components/MapView/index.tsx
+++ b/src/components/MapView/index.tsx
@@ -211,14 +211,14 @@ function MapView({ classes }: MapViewProps) {
     const HAZARD_LAYER_PARAM = 'hazardLayerId';
     const BASELINE_LAYER_PARAM = 'baselineLayerId';
 
-    const hazardLayerId = urlParams.get(HAZARD_LAYER_PARAM);
+    const hazardLayerIds = urlParams.get(HAZARD_LAYER_PARAM);
     const baselineLayerId = urlParams.get(BASELINE_LAYER_PARAM);
 
     /*
       In case we don't have hazard or baseline layers we will use the default
       layer provided in the appConfig defined within `prism.json` file.
      */
-    if (!hazardLayerId && !baselineLayerId) {
+    if (!hazardLayerIds && !baselineLayerId) {
       const defaultLayer = get(appConfig, 'defaultLayer');
 
       if (defaultLayer) {
@@ -242,7 +242,7 @@ function MapView({ classes }: MapViewProps) {
     }
 
     if (
-      (!hazardLayerId && !baselineLayerId) ||
+      (!hazardLayerIds && !baselineLayerId) ||
       Object.keys(serverAvailableDates).length === 0
     ) {
       return;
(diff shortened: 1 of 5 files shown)


src/components/MapView/index.tsx
```diff
@@ -211,14 +211,14 @@ function MapView({ classes }: MapViewProps) {
     const HAZARD_LAYER_PARAM = 'hazardLayerId';
     const BASELINE_LAYER_PARAM = 'baselineLayerId';
 
-    const hazardLayerId = urlParams.get(HAZARD_LAYER_PARAM);
+    const hazardLayerIds = urlParams.get(HAZARD_LAYER_PARAM);
```
ericboucher, line 214: Should we rename HAZARD_LAYER_PARAM to hazardLayerIds?

wadhwamatic:
@JorgeMartinezG - this works as expected, thanks! Was Eric's comment addressed?

@ericboucher or @tdlinh2712 - can you provide a code review?

src/components/MapView/index.tsx
```diff
@@ -242,59 +242,68 @@ function MapView({ classes }: MapViewProps) {
...
-            message: 'Invalid date found. Using most recent date',
-            type: 'warning',
-          }),
-        );
-      } else {
-        dispatch(updateDateRange({ startDate: dateInt }));
-      }
+    const dateInt = moment(urlDate).valueOf();
+
+    if (urlDate === null || dateInt === selectedDate) {
```
tdlinh2712, line 292: `if (!urlDate || dateInt === selectedDate)`

src/components/NavBar/MenuSwitch/SwitchItem/index.tsx
```diff
@@ -29,12 +29,40 @@ import { removeLayer } from '../../../../context/mapStateSlice';
...
+
+const removeLayerFromUrl = (
+  urlLayers: string | null,
+  layerId: string,
+): string => {
+  const selectedLayersUrl = urlLayers !== null ? urlLayers.split(',') : [];
+
+  return selectedLayersUrl.filter(l => l !== layerId).join(',');
+};
+
```
tdlinh2712, line 59: I think we should generalize and put this in the `useUrlHistory` hook so that these functions can be reused later

ericboucher:
@JorgeMartinezG please provide a clearer description of the PR and what it does in the PR body (initial comment)

src/components/MapView/index.tsx
JorgeMartinezG: fixed in commit 244f3695edd712317dd482073e3fdad250bb0d2f

wadhwamatic (approved):
Nice work @JorgeMartinezG . Tested and working as expected

ericboucher (approved):

PR #492, opened 2022-06-24
remove scale label parameter in chart config if x-axis is undefined

JorgeMartinezG:
This PR fixes a bug where running an analysis followed by clicking on an layer with a chart associated with it causes an error.


ericboucher (approved):

ericboucher:
@JorgeMartinezG not needed here but in general try to add before/after screenshots as you work on it :)

PR #497, opened 2022-06-30
Revert "Multiple layers url"

wadhwamatic:
Reverts WFP-VAM/prism-frontend#490


wadhwamatic:
It looks like the last commit was merged but not the full branch history

Issue #178, opened 2021-05-04
Admin boundaries are not always visible when other layers are loaded


wadhwamatic:
With some raster layers, and when the opacity of other layers are set high, the admin boundaries are not clearly visible. Admin boundaries should always be the top layer to prevent this from happening. 

Difficult to see boundaries with a darker raster data layer loaded:
<img width="1426" alt="Screen Shot 2021-05-04 at 15 36 28" src="https://user-images.githubusercontent.com/3343536/117078392-a1b32c00-acee-11eb-9133-f9934f7bb18d.png">

Admin boundaries are not visible if user raises opacity:
<img width="1424" alt="Screen Shot 2021-05-04 at 15 39 11" src="https://user-images.githubusercontent.com/3343536/117078539-eb037b80-acee-11eb-9f7d-1c38cf9ce1f1.png">

wadhwamatic:
This functionality broke when we introduced multiple boundary layers. Now, only the default boundary persists above other layers. Secondary boundary layers are not visible

PR #502, opened 2022-07-01
Fix multiple admin boundaries

hav142:
Fix visibility for multiple admin boundaries (issue #178)

- before (boundary is not clear when layer transparency is high)
image

- after
image


wadhwamatic:
Thanks @hafidz142. This works well for a single layer, but when I increased the opacity for an exposure analysis, or 'run analysis' results layer, the admin boundaries become hidden. See the screenshots below for examples from Myanmar, but the issue exists in any deployment.

Select tropical storm layer, and date May, 21, 2016. Then click exposure analysis button, and increase result layer's opacity:
Screen Shot 2022-07-01 at 23 59 33

Run analysis on any hazard and vulnerability layer, then increase opacity of results layer:
Screen Shot 2022-07-02 at 00 01 01

wadhwamatic:
@hav142 - the latest commit fixed the issues I found earlier. I also verified that this works as expected when using the selection feature in the alert module

@ericboucher - can you review or assign for technical review?

src/components/MapView/index.tsx
```diff
@@ -457,7 +457,7 @@ function MapView({ classes }: MapViewProps) {
           });
         })}
         {/* These are custom layers which provide functionality and are not really controllable via JSON */}
-        <AnalysisLayer before={firstSymbolId} />
```
ericboucher, line 460: why not pass in the boundary as before here?

ericboucher (commented):
I like what you did, but you need to make it more explicit and add a few comments to explain that we are:

- setting boundary layers to be before firstSymbolId
- setting all other layers to be below the first boundary layer

ericboucher (approved):

hav142:
> I like what you did, but you need to make it more explicit and add a few comments to explain that we are:
> 
>     * setting boundary layers to be before firstSymbolId
> 
>     * setting all other layers to be below the first boundary layer

I noticed that not all layer types have a `before` parameter, such as `WMSLayer` which uses the `before` setting in the component itself. should we uniformly use this `before` parameter for the other layer types?

wadhwamatic:
> > I like what you did, but you need to make it more explicit and add a few comments to explain that we are:
> > ```
> > * setting boundary layers to be before firstSymbolId
> > 
> > * setting all other layers to be below the first boundary layer
> > ```
> 
> I noticed that not all layer types have a `before` parameter, such as `WMSLayer` which uses the `before` setting in the component itself. should we uniformly use this `before` parameter for the other layer types?

@ericboucher - can you give some final feedback to @hav142 on this?

src/components/MapView/index.tsx
```diff
@@ -457,7 +457,7 @@ function MapView({ classes }: MapViewProps) {
...
+        <AnalysisLayer before={boundaryLayer} />
```
ericboucher, line 460: I think `before` should still be a string, and you would pass `layer-${boundaryLayer.id}-line`

ericboucher: @hav142 yes I think it would be nice to push towards uniformization. For example we could have:
- all layers accept a before parameter
- control ordering from`MapView/index.tsx` so that it is clearer and easier to update in the future.

What do you think?

hav142: I agree that we should control the ordering from one place and use `before` parameter to easier maintenance. I have added it to the other layer types, but for `AdminLevelDataLayer` it seems have unique boundary and it own setting. should we just leave it like that or we can replace it as well?

ericboucher (dismissed):

src/components/MapView/index.tsx
```diff
@@ -416,10 +422,11 @@ function MapView({ classes }: MapViewProps) {
   // Saves a reference to base MapboxGL Map object in case child layers need access beyond the React wrappers.
   // Jump map to center here instead of map initial state to prevent map re-centering on layer changes
   const saveAndJumpMap = (map: Map) => {
-    // Find the first symbol on the map to make sure we add layers below them.
     const { layers } = map.getStyle();
-    // Find the first symbol on the map to make sure we add layers below them.
+    // Find the first symbol on the map to make sure we add boundary layers below them.
     setFirstSymbolId(layers?.find(layer => layer.type === 'symbol')?.id);
+    // and add other layers below boundary layer
```
ericboucher, line 428: `below boundary layers` with an S please

src/components/MapView/index.tsx
ericboucher: Have you tested that it works properly? If it does, then we can leave it like that for now @hav142

hav142: I have tested other layer types and they work, but I will do more detailed tests in other countries too.

ericboucher: Make sure to try `AdminLevelDataLayer` and once you confirm it works I think we are good to go

wadhwamatic (dismissed):
I tested across multiple layer types and see this working as expected. Nice work @hav142

wadhwamatic:
Ah, sorry. I just found a minor issue with an edge case. If you toggle between admin level data layers with different associated boundary files, the admin 1 boundary doesn't seem to come back to the top. See screenshots below. First one is before toggling to a lower level admin level layer, the second is after.

Steps to reproduce: 
2. Activate layer vulnerability/multidimensional poverty index
3. Next activate layer exposure/total population
4. Bump up opacity to 100% and you'll see the admin1 level boundary is no longer on top

Screen Shot 2022-07-11 at 11 07 21

Screen Shot 2022-07-11 at 11 07 47

hav142:
there is another issue when we activate analysis layer. when a layer has it own unique boundary, we can't use `before` parameter since the boundary layer will not be available on the map and it causes an error. So, I added a condition that if the layer has its own boundary, it will use the setting instead of the `before` parameter.

wadhwamatic:
Nearly there from a functionality perspective. The last issue I came across is that after running an analysis with a unique boundary, and then clearing the analysis, we should see the default boundaries again.

To reproduce:

2. Run this analysis: http://localhost:3000/?analysisHazardLayerId=rainfall_dekad&analysisBaselineLayerId=mpi&analysisDate=2022-07-21&analysisStatistic=mean
3. Clear analysis results using 'clear analysis' button. The map shows admin1 boundaries instead of reverting to default admin2 boundaries

Screen Shot 2022-07-20 at 18 45 58

wadhwamatic:
@hav142 - The last issue mentioned regarding reverting back to the default boundary file after running an analysis works. But, if I load an admin data layer which is not the default, and then load a WMS layer, nothing happens. Steps to reproduce:
2. Load vulnerability / multidimensional poverty index layer
3. Load any WMS layer (ex. 10-day rainfall anomaly). The WMS layer doesn't load

This is specific to admin data layers which do not use the default boundary. See screenshots below of the app and errors in console.

Screen Shot 2022-07-24 at 21 04 15

Screen Shot 2022-07-24 at 21 06 23

Screen Shot 2022-07-24 at 21 06 34

ericboucher:
I think this is a good example of why we should implement cypress tests to toggle layers on/off.
@wadhwamatic could write a layer test protocol that we could implement? I.e. a sequence of layers we want to turn on/off. And we can then add it as part of the CI

wadhwamatic:
> I think this is a good example of why we should implement cypress tests to toggle layers on/off. @wadhwamatic could write a layer test protocol that we could implement? I.e. a sequence of layers we want to turn on/off. And we can then add it as part of the CI

@ericboucher - sounds good. I don't have time to do this now but please create an issue and we can get to this later

wadhwamatic:
@hav142 - this is looking good now. thanks for all the changes. @ericboucher - could you give a final code review?

src/components/MapView/index.tsx
```diff
@@ -311,13 +314,23 @@ function MapView({ classes }: MapViewProps) {
...
   useEffect(() => {
     dispatch(loadAvailableDates());
+
     const displayBoundaryLayers = getDisplayBoundaryLayers();
+    /*
+      reverse the order off adding layers so that the first boundary layer will be placed at the very bottom,
+      to prevent other boundary layers being covered by any layers
+    */
+    // eslint-disable-next-line fp/no-mutating-methods
+    const reversedBoundaryLayer = displayBoundaryLayers
```
ericboucher, line 324: you can probably do it a bit simplified like this:

const reversedBoundaryLayer = displayBoundaryLayers
      .reverse()
      .map(layer => displayBoundaryLayers.filter(boundary => boundary.id === layer.id)[0]);

ericboucher (approved):
This looks good, just a small simplification suggestion and we're good to go

src/components/MapView/index.tsx
hav142: Yes it should work with just reverse(). I thought it didn't work before because it didn't change when I was debugging. Just realized that I have to clone the variable first before debugging because it destruct the initial variable 🙈

PR #503, opened 2022-07-01
Simplify layer group attribute

hav142:
This PR aims to simplify {country}/prism.json for layer group configuration, i.e:
```
"layers": [
      {
        "id": "rainfall_dekad",
        "label": "10-day",
        "main": true
      },
      {
        "id": "rainfall_agg_1month",
        "label": "1-month",
        "main": false
      },
      {
        "id": "rainfall_agg_3month",
        "label": "3-month",
        "main": false
      },
      ...
```
by removing the `main` attribute for all layers with `false` values, and setting `true` only for the main/default layer of the group.
We also changed the attribute from `main: true` to `default: true` to make it easier to identify


wadhwamatic:
Nice work @hav142. I appreciate the simplification of the config process! Let's get code review from @ericboucher or @JorgeMartinezG

ericboucher:
@hav142 I like the simplification, but I am not sure I agree with the name change from `main` to `default`. @wadhwamatic  what do you think?

wadhwamatic:
> @hav142 I like the simplification, but I am not sure I agree with the name change from `main` to `default`. @wadhwamatic what do you think?

I don't have a strong opinion on this. One hand, it makes sense to call it default as it will become the first option in the list. But main is used in the similar layer grouping feature. For consistency, let's revert back to main.

PR #505, opened 2022-07-05
Fix for "Allow PRISM build to be served from a subfolder"

reed1:
Sorry I forgot to check the "Run Analysis" in deployed URL for PR #474 (it has been merged now). The feature will not be working if it's served from a subfolder. This should fix it.

Deployed here: http://prism-sub-4.surge.sh/subfolder/build


ericboucher (approved):

ericboucher:
Ok, were you able to test that it works properly @reed1 ?

PR #489, opened 2022-06-06
Add access token for required layers

JorgeMartinezG:
image

PR #506, opened 2022-07-05
Layer authentication through JWT

JorgeMartinezG:
This PR enables PRISM to authenticate using json web tokens (JWT). On a layer with tokenRequired parameter, the user **must** provide a valid json webtoken. Then, the backend decodes it using the secret key defined within JWT_SECRET_KEY environment variable to get the region bounding box. Finally, the backend implements a geospatial filter to find the points within the polygon.

**How to test**: Using Cambodia deployment.
- Rebuild the docker image `make deploy`.
- Select the **disaster impact report** layer.
- include the access token for Phnom Penh:
```
eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJBZG0xX05hbWUiOiJQaG5vbSBQZW5oIiwiQWRtMV9OQ0REIjoxMiwiQURNMV9DT0RFIjoiS0gxMiIsImJib3giOiIxMDQuNzIwMDU5NDMxLDExLjQxOTQ5NDYyOSwxMDUuMDQ3NjkxNzk1LDExLjczNTk3NDU3OCJ9.PtzY8Ix_RRhiL9gK__LDX2vJTf4690O6pByx9TKea5U
```
image

**TODO:**
- Make a script to generate the json web tokens using the Makefile


wadhwamatic:
Looks good so far @JorgeMartinezG. Once it's merged, we also need to incorporate changes from https://github.com/WFP-VAM/prism-frontend/pull/485

Please also make the same changes noted in the first version of this feature (https://github.com/WFP-VAM/prism-frontend/pull/489#issuecomment-1175585167 - content pasted below)

Can we move the invalid credentials message to the authentication modal? Currently, it's not really noticeable since the authentication modal is displayed again, and then once you login correctly, you'll see the 'invalid credentials' message from before. Screenshots below.

Screen Shot 2022-07-05 at 15 59 05

Screen Shot 2022-07-05 at 16 00 05

JorgeMartinezG:
@wadhwamatic another option is to force the notification to appear on top of the modal, like in the figure below:

image

wadhwamatic:
@JorgeMartinezG - the revised flow for failed authentication works well. But, when I authenticate correctly and if no matching data is found, I'm brought again to the authentication modal. Is this intentional?

Screen Shot 2022-07-06 at 23 14 21

JorgeMartinezG:
@wadhwamatic yes, it is intentional

wadhwamatic:
> @wadhwamatic yes, it is intentional

We might consider deactivating the layer and not showing the modal again instead. If a user wants to try a different region, they could then reactivate the layer and get prompted again to authenticate. I think the current flow assumes you want to try to a different region but in practice, its more likely that I'm only viewing my province and if no data exists, that's the end of the process

wadhwamatic:
@JorgeMartinezG - I tried testing again but repeatedly get an 'Invalid credentials' message. I used the token you included in the instructions and have the same 'JWT_SECRET_KEY' as before. The backend returns "Invalid access token".

Please make sure to update the testing instructions so that anyone can test this locally. Thanks

wadhwamatic (commented):
Tested again. All previous comments have been addressed. There's an issue with the date selector showing data available for all forms, regardless of province. But we can address that separately

Issue #496, opened 2022-06-29
Add black formatting


ericboucher:
Add black formatting and isort

PR #512, opened 2022-07-08
Add black formatting and isort to the api code

PysarenkoDS-BWT:
Add black formatting and isort

Add black formatting #496


.github/workflows/test.yml
```diff
@@ -0,0 +1,43 @@
+name: PRISM Test
```
ericboucher, line 1: Since there is another file running PRISM CI already exists, what we can do is split tests between the frontend and the backend, so maybe rename this file: `api_test.yml` and the other one `frontend_test.yml` and split tests a bit more cleanly

ericboucher (commented):
Great start. Let's make sure that all the tests pass now :)
And add a comment in the api README to explain that we are using black and isort for formatting

ericboucher (approved):

.github/workflows/build.yml
```diff
@@ -104,8 +104,36 @@ jobs:
           message: |
             Build succeeded and deployed at http://${{ env.surge_url }} 
             (hash ${{ github.event.pull_request.head.sha }} deployed at ${{ steps.date.outputs.date }})
+  api_lint:
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
ericboucher, line 110: @PysarenkoDS-BWT we don't need these environment variables

.github/workflows/build.yml
```diff
@@ -104,8 +104,36 @@ jobs:
...
+      ORG_LEVEL_TOKEN: ${{secrets.ORG_LEVEL_TOKEN}}
+    strategy:
```
ericboucher, line 112: Since we are simply running formatting, we do not need to run on multiple python versions, let's just use 3.10

.github/workflows/build.yml
ericboucher, line 110: we don't need these environment variable @PysarenkoDS-BWT

PR #514, opened 2022-07-08
Update layers and boundaries COUNTRY=kyrgyzstan

wadhwamatic:
This PR makes updates to the initial configuration for Kyrgyzstan. New admin boundaries have been provided by the government counterpart, and all available layers from WFP's ODC instance have been configured

Surge deployment here:
https://prism-kgz-demo.surge.sh/

Screen Shot 2022-07-14 at 12 46 46


ericboucher:
Screen Shot 2022-07-14 at 1 29 21 PM

wadhwamatic:
> @wadhwamatic I pushed a fix but the reason it was not working is that the file `/kyrgyzstan/District_KRYG.json` is corrupted and the last data input only has null values.

Great, thanks a lot @ericboucher. FYI - I didn't see any errors prior to adding an admin-level layer. I'm wondering if we should add some validation to the boundary file and write issues to the console to catch this earlier.

PR #462, opened 2022-04-27
Add maxBounds, minZoom, and maxZoom to prism.json map config

reed1:
For issue #179

PR #525, opened 2022-07-15
Feature/show boundary info box

reed1:
This will show an optional boundary info box on bottom left corner:

image

This feature is optional (default false) and can be enabled with REACT_APP_SHOW_BOUNDARY_INFO set to "true". This feature is helpful to get the right maxBounds value for PR #462. We can pan and fit the country on the screen then copy the boundary and zoom info into the config file.


src/components/MapView/index.tsx
```diff
@@ -410,6 +412,24 @@ function MapView({ classes }: MapViewProps) {
...
+      const newZoom = map.getZoom();
+      dispatch(setLocation({ bounds, zoom: newZoom }));
+    };
+    map.on('dragend', onDragend);
+    map.on('zoomend', onZoomend);
+    // Show initial value
+    onZoomend();
+  };
+
+  const showBoundaryInfo = process.env.REACT_APP_SHOW_BOUNDARY_INFO === 'true';
```
ericboucher, line 432: Maybe use JSON.parse(REACT_APP_SHOW_BOUNDARY_INFO)?

https://stackoverflow.com/a/54906327/16585247

Please make sure it is False when the env is not set

wadhwamatic:
@reed1 - can we wrap this up? See Eric's comments from last week

src/components/MapView/index.tsx
ericboucher: Resolved in [4051438](https://github.com/WFP-VAM/prism-frontend/pull/525/commits/40514389418d2fbdf6e42c4120638fab65292c7d)

ericboucher (approved):

Issue #443, opened 2022-04-08
Remove dependency to S3 files


ericboucher:
Currently, running raster analysis locally requires to have a boundary file on S3. We should remove this dependency. To do so, we have two options.

1. When running locally, we send the geojson boundaries instead of url

2. Instead of using S3, we add a “defaultUrl” to each country, pointing to their main deployment URL. This way, we don’t need to maintain files on S3 and the URL will be something like [prism-mongolia.org/data/boundary_file.json](http://prism-mongolia.org/data/boundary_file.json)

Since the API already accepts `zones` as a GeoJSON, it is probably the easiest option.

PR #526, opened 2022-07-18
Remove dependency to S3 files

PysarenkoDS-BWT:
Issue #443 

Currently, running raster analysis locally requires to have a boundary file on S3. We remove this dependency.

Also we can add defaultURL property to each country (for example I add it to Cuba)
if this property exists, then we will receive data from the specified path

For example without it path is `http://localhost:3000/data/myanmar/admin_boundaries.json`
But if we add `defaultUrl: 'prism-cuba.org'` we take next path `prism-cuba.org/data/myanmar/admin_boundaries.json`


src/context/mapStateSlice/index.ts
```diff
@@ -48,17 +49,29 @@ export const mapStateSlice = createSlice({
...
   reducers: {
     addLayer: ({ layers, ...rest }, { payload }: PayloadAction<LayerType>) => {
-      const layersToAdd = payload?.group?.activateAll
+      const updatedPayload = {
+        ...payload,
+        ...(payload.type === 'boundary' &&
+          countryBaseURL && {
+            path: `${countryBaseURL}/${payload.path}`,
+          }),
+      };
```
ericboucher, line 58: can you explain why we're making this change?

PysarenkoDS-BWT: We need this to add baseURL to path field if countryBaseURL available

src/context/analysisResultStateSlice.ts
```diff
@@ -101,10 +99,6 @@ function getAdminBoundariesURL() {
   if (adminBoundariesPath.startsWith('http')) {
     return adminBoundariesPath;
   }
-  // do not send a local path to the API, use a fixed boundary file instead.
-  if (isLocalhost) {
-    return defaultBoundariesPath;
```
ericboucher, line 106: I think this is the only part that should change, actually

src/context/analysisResultStateSlice.ts
```diff
@@ -101,10 +99,6 @@ function getAdminBoundariesURL() {
   if (adminBoundariesPath.startsWith('http')) {
     return adminBoundariesPath;
   }
-  // do not send a local path to the API, use a fixed boundary file instead.
-  if (isLocalhost) {
```
ericboucher, line 105: this will break when the API is being run locally, no?

src/config/cuba/index.ts
```diff
@@ -9,5 +9,6 @@ export default {
   rawLayers,
   rawTables,
   translation,
+  defaultUrl: 'prism-cuba.org',
```
ericboucher, line 12: This URL does not exist, so why put it in here?

PysarenkoDS-BWT: for tests, sorry, forgot to delete it(

src/context/analysisResultStateSlice.ts
PysarenkoDS-BWT: No, all works fine

PR #527, opened 2022-07-18
Add auto linter for json files

PysarenkoDS-BWT:
Add eslint-plugin-json to lint json files


ericboucher:
When/how do we run it? I'm surprised it doesn't change anything in the existing JSON files

PysarenkoDS-BWT:
Add script to run prettier for JSON files, run only manually (not add automatically), because this not fast operation

ericboucher:
@PysarenkoDS-BWT thanks for updating the files. Please make sure that the CI tests are failing when the files are not properly formated

PysarenkoDS-BWT:
@ericboucher I think we can't check it in CI/CD because for this check need a loot of memory, remove this check?

ericboucher (changes requested):
@PysarenkoDS-BWT actually, let's not run this on files that are in public/data/COUNTRY_NAME but only on config files for now. This is making all the geojson files even bigger, which is clearly not what we want, thanks :)

src/components/MapView/AlertForm/index.tsx
```diff
@@ -111,35 +112,35 @@ function AlertForm({ classes, isOpen, setOpen }: AlertFormProps) {
...
   };
 
-  const onOptionChange = <T extends string>(
-    setterFunc: Dispatch<SetStateAction<T>>,
-  ) => (event: React.ChangeEvent<HTMLInputElement>) => {
-    const value = event.target.value as T;
-    setterFunc(value);
-    return value;
-  };
+  const onOptionChange =
```
ericboucher, line 115: Why are the ts and tsx files changing? Can we limit the changes to  json files?

ericboucher (approved):

PR #529, opened 2022-07-22
Remove oldest layer when intersection of multiple layers is empty

JorgeMartinezG:
When two layers of different types are selected, there is a possibility that there is no date available on both dates arrays at the same time. This make an empty intersection array.

This PR removes the oldest layer and sets the date based on the closest date from the layer that is kept.

**How to test:** Using cambodia version.
- Select _10-day NDVI (MODIS)_ and then _Flood report_ layers.

image


ericboucher:
It works well overall, but I am wondering if we should default to the "latest date" available until the user actually selects a different date. At the moment, when we switch from NDVI to Flood, we load the closest date, but more recent reports are available. I'm also fine to moving this discussion to another issue/PR

Screen Shot 2022-07-22 at 4 26 06 PM

In addition, how about adding a timer to warning messages and remove them automatically after 5 or 10 seconds if they're still there?

src/components/MapView/index.tsx
```diff
@@ -357,14 +367,37 @@ function MapView({ classes }: MapViewProps) {
...
+
+      // Remove layer from url.
+      const urlLayerKey = getUrlKey(layerToRemove);
+      const updatedUrl = removeLayerFromUrl(urlLayerKey, layerToRemove.id);
+      if (updatedUrl === '') {
+        removeKeyFromUrl(urlLayerKey);
+      } else {
+        updateHistory(urlLayerKey, updatedUrl);
+      }
+      dispatch(removeLayer(layerToRemove));
```
ericboucher, line 391: This part feels overly complicated, I see that this logic is used in another place already but I believe that it can be simplified. Again, this can be addressed later but I believe that we can:
1. better separate the responsibilities between holding layers and managing the URL
2. make `removeLayerFromUrl` a bit smarter and move some of the logic in there

wadhwamatic:
> It works well overall, but I am wondering if we should default to the "latest date" available until the user actually selects a different date. At the moment, when we switch from NDVI to Flood, we load the closest date, but more recent reports are available. I'm also fine to moving this discussion to another issue/PR

as mentioned on Slack, I think closest date is the right behavior. We could be looking at a historical flood event (often the case in demos). Toggling to the most recent data would be confusing

> In addition, how about adding a timer to warning messages and remove them automatically after 5 or 10 seconds if they're still there?

Agreed and opened an issue https://github.com/WFP-VAM/prism-frontend/issues/530

ericboucher (approved):

PR #539, opened 2022-07-29
Issue 413: created packages folder and the first package format-url

DanielJDufour:
Inspired by [TurfJS's packages folder](https://github.com/Turfjs/turf/tree/master/packages), I created a packages folder which will hold the code that is shared between the different JavaScript pieces of the application (frontend, alerting, etc.).  I think the packages approach is particularly useful because it encourages modularity and easy testing of small components.  It also makes it easier to spin packages off into separate repos if they become large enough or might be reused in other projects.

Additionally, "vendoring in" in the packages at this time (instead of putting them in a separate repo) makes integration easier and also allows us to easily test integration with PRISM making sure we don't cause any regressions.

The new steps will be to add additional modular functionality, especially around fetching metadata and data from OGC Web Services, but I wanted to start with something simple.

Specifically on the formatUrl function, I tightened up some of the functionality.  Previously a param with an undefined value would render `?param=undefined` because `undefined.toString()` is `"undefined"`.  I also added default sorting of param keys, so the same params will always lead to the same string.  Ordered params could always be accomplished with an array or map if that's something we want to add in the future.

## Testing
 I added the ["flug" testing library](https://github.com/danieljdufour/flug), which is designed to be minimal, reliable and fast to develop with.  I wrestled with Jest for hours before giving up on that.  Once we move the main frontend configuration files to a subfolder, we should then be able to use Jest with more ease.  I like flug for the simplicity, but could definitely understand wanting to standardize around jest.

Open to your thoughts and suggestions :-)


DanielJDufour:
I ran into issue as it seems Create React App (CRA) doesn't allow you to import from outside of the src folder.  I'll see if I can solve this with yarn relative path installs...

DanielJDufour:
I got yarn install to work with relative paths by updating the package.json to `"packages": "./packages",` in dependencies, but the issue is that Create React App doesn't support importing raw typescript files.  It expects imports to already be compiled to JavaScript.  I can try using symbolic links, which should trick the compiler into thinking the folder is inside src, but I'm worried it might not play nice on Windows.

DanielJDufour:
In theory symlinks should work on windows ([source](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/#:~:text=Symlinks%2C%20or%20symbolic%20links%2C%20are,supported%20symlinks%20since%20Windows%20Vista.)).  So I created a symbolic link `ln -s ../packages packages` when inside of the src folder.  However, when I import it with `import formatUrl from '../packages/format-url';
` inside of `src/utils/server-utils.ts`, I get the following errors:
Screen Shot 2022-07-30 at 8 49 32 PM

My best guess is that Create React App is failing to follow the symbolic link and find the index.ts inside of format-url.  What is crazy is that VSCode thinks the import is valid and can track down the index.ts without a problem
Screen Shot 2022-07-30 at 8 51 21 PM

DanielJDufour:
There's a few directions that we can take things from here, but none of them are great:
- Add a build step to each of the packages.  It shouldn't be too difficult, perhaps adding a call to `tsc` or a call to `babel with preset-typescript (and a few other plugins) to translate the typescript code to ESM JavaScript. (Here's an example of that: https://github.com/DanielJDufour/pixel-utils/blob/main/package.json#L72)
- Add an install script to the frontend package.json file that just hard copies the source files from the top-level packages folder to the src/packages folder.  The drawback here is that they could get out of sync and would be a pain to develop with because we'd have to rerun the code on each change.

It looks like we'll eventually have to add a build step into each package because CRA doesn't support importing external raw typescript files (unless we can discover a clean hack around this).

I'll do more research and report back later.

DanielJDufour:
I think I have a workable solution now, but we should of course discuss it more, because it's not ideal.  I added a build step to the format-url package (and we'll have to add a build step to all the shared packages probably).  It's fairly painless just `npx tsc index.ts`.  It's just annoying having to run it and compile all the packages to JS when we are running a typescript project.  Even after compiling you have to run `yarn install --check-files` to make sure that yarn updates the "installed" packages folder in the node_modules with the new changes.

Open to your thoughts and suggestions :-)

DanielJDufour:
Jest wants to use the source index.ts instead of the built index.js inside of format-url.  I'll investigate and report back.  Perhaps an update to the jest configuration could fix it.
Screen Shot 2022-07-30 at 9 33 36 PM

packages/ogc/src/wms/layer.ts
```diff
@@ -0,0 +1,39 @@
...
+    const url = await this.getImageUrl(options);
+    const response = await this._fetch(url);
+    const arrayBuffer = await response.arrayBuffer();
+    return {
+      ...options,
+      image: arrayBuffer
+    };
+  }
+
+  // getFeatureInfo({ x, i, j, y }) {}
```
ericboucher, line 38: 👣

packages/ogc/src/wms/layer.ts
```diff
@@ -0,0 +1,39 @@
+import { Layer } from "../layer";
+
+// import { parseName } from "../base/utils";
```
ericboucher, line 3: 👣

packages/ogc/src/common/index.ts
```diff
@@ -0,0 +1,16 @@
+export { default as bboxToString } from "./bbox-to-string";
```
ericboucher, line 1: Not a huge fan of having one file per function like this. How about merging them into a few files by topics? Some of them also don't seem to be ogc specific?

.gitignore
```diff
@@ -33,3 +34,6 @@ yarn-error.log*
 .idea/
 .log/
 *.bak
+
+# ignore shared code packages build files
+packages/*/*.js
```
ericboucher, line 39: Could we exclude the build folders instead if it's easy?

ericboucher (changes requested):
This looks good overall. Some feedback here and inline.

- I think the "packages" folder should simply be named "common"
- The structure could be simplified
- Plus, let's use this PR to move the frontend code inside a "website" folder

alerting/src/utils/raster-utils.ts
```diff
@@ -1,4 +1,4 @@
-import { formatUrl } from './server-utils';
+import formatUrl from '../../../packages/format-url';
```
ericboucher, line 1: Once "packages" becomes "common" and is built as an independent package, we should ideally be able to do the import like this instead: `import formatUrl from 'common/format-url'`;

We have a similar structure for our platform `ovio.org`, I can give you access to the code if it helps @DanielJDufour

packages/format-url/index.ts
```diff
@@ -0,0 +1,19 @@
+export default function formatUrl(
```
ericboucher, line 1: Having format-url as an independent package feels like an overkill. Let's try to aggregate things a bit more

packages/ogc/src/common/is-nil.ts
```diff
@@ -0,0 +1,3 @@
+export default function isNil(it: undefined | null | "" | any): boolean {
```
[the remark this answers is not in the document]
DanielJDufour: isNil was replaced with lodash's isEmpty via https://github.com/DanielJDufour/prism-frontend/commit/12e3d617ab24c3341a1f18eafc9c25b6378b53b2

I'll look to replace any other lodash-like functions in addition to this

alerting/src/utils/raster-utils.ts
DanielJDufour: I reworked things per your suggestion with a slight modification.  I updated the package.json to install the `packages/common` as `prism-common`.  I'm thinking it will help clarify for other developers where the package is coming from and that it is not https://www.npmjs.com/package/common.  Happy to change to just `"common"` if you prefer that though.

packages/format-url/index.ts
DanielJDufour: this is complete.  I removed the format-url package and people can use the formatUrl export of the common package

packages/ogc/src/wms/layer.ts
DanielJDufour: good catch. fixed with https://github.com/DanielJDufour/prism-frontend/commit/81b289a7ab13e84a78205cc73add01f29a3f2b55

packages/common/src/common/set-timeout-async.ts
```diff
@@ -0,0 +1,16 @@
+export default function setTimeoutAsync(
+  seconds: number,
+  callback: () => any
+): Promise<any> {
+  // how do I error out if promise rejected
```
ericboucher, line 5: 👣  or ❓ ?

packages/common/src/wcs/utils/index.ts
```diff
@@ -0,0 +1,288 @@
...
+
+  // version 1.1.x
+  const identifier = findCoverageIdentifier(xml);
+  if (identifier) return identifier;
+
+  // version 1.0
+  const name = findCoverageName(xml);
+  if (name) return name;
+
+  console.log("===========");
```
ericboucher, line 70: should this be a `console.debug`?

packages/common/src/wcs/utils/index.ts
```diff
@@ -0,0 +1,288 @@
...
+
+export function findDescribeCoverageUrl(xml: string): string | undefined {
+  return findCoverageOpUrl(xml, "DescribeCoverage");
+}
+
+export function findGetCoverageUrl(xml: string): string | undefined {
+  return findCoverageOpUrl(xml, "GetCoverage");
+}
+
+// export function formatGetCoverageUrl(
```
ericboucher, line 168: 👣 ?

packages/common/src/wfs/utils.ts
```diff
@@ -0,0 +1,236 @@
...
+    if (format === "geojson") {
+      return response.json();
+    } else if (format === "xml") {
+      return response.text();
+    }
+  };
+  return setTimeoutAsync(wait, run);
+}
+
+// export async function getLayerDates
```
ericboucher, line 236: 👣

packages/common/src/wms/WMS/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  async getLayer(layerId: string): Promise<WMSLayer> {
+    this.checkLayer(layerId);
+    return new WMSLayer({
+      capabilities: this.capabilities,
+      id: layerId,
+      fetch: this._fetch
+    });
+  }
+}
```
ericboucher, line 24: ↩️

packages/common/src/wms/index.ts
```diff
@@ -0,0 +1,2 @@
+// export { wms } from "./WMS";
```
ericboucher, line 1: 👣 ?

packages/common/src/wms/index.ts
```diff
@@ -0,0 +1,2 @@
...
+export { WMS } from "./WMS";
```
ericboucher, line 2: ↩️

packages/common/src/wms/layer.ts
```diff
@@ -0,0 +1,38 @@
...
+    const url = await this.getImageUrl(options);
+    const response = await this._fetch(url);
+    const arrayBuffer = await response.arrayBuffer();
+    return {
+      ...options,
+      image: arrayBuffer
+    };
+  }
+
+  // getFeatureInfo({ x, i, j, y }) {}
```
ericboucher, line 37: 👣 ?

packages/common/src/wms/utils/index.ts
```diff
@@ -0,0 +1,203 @@
...
+  srs: string[];
+  bbox?: Readonly<[number, number, number, number]>;
+  attribution?: {
+    title?: string;
+  };
+  styles: Style[];
+  dates: string[];
+};
+
+// export function findLayerTags(xml: string): { inner: string; outer: string }[] {
```
ericboucher, line 49: 👣 ?

packages/common/yarn.lock
```diff
@@ -0,0 +1,958 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
ericboucher, line 1: Not linked to this file but is `packages/common/src/wmts/.gitkeep` needed? Or we can add the folder when we have something to put in it, no?

ericboucher (approved):
Looking good, just a few cleanup suggestions inline :)

packages/common/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "ogc",
```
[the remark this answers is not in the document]
DanielJDufour: this is fixed. thanks for catching that.

DanielJDufour:
I reformatted the common functions so they are grouped by topic

packages/common/src/common/index.ts
```diff
@@ -0,0 +1,15 @@
+export { default as bboxToString } from "./bbox-to-string";
```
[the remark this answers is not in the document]
DanielJDufour: yes, sounds good. I renamed folder to utils and also grouped functions by topic.

packages/common/src/wcs/utils/index.ts
DanielJDufour: I removed the console.log.  Accidentally left it in from prototyping phase.

packages/common/src/wms/index.ts
DanielJDufour: done. removed

packages/common/src/wms/utils/index.ts
DanielJDufour: I have removed findLayerTags as it is not being used.

packages/common/yarn.lock
DanielJDufour: not needed.  I removed it.

packages/common/src/wms/index.ts
DanielJDufour: I removed unused comment

packages/common/src/wms/layer.ts
DanielJDufour: removed

packages/common/src/utils/image.ts
```diff
@@ -0,0 +1,35 @@
+import { checkExtent } from "./bbox";
+import type { BBOX } from "../types";
+
+export function scaleImage(extent: BBOX,
+  {
+    check_extent = true,
+    resolution = 256,
+    max_pixels = 5096
+  }: {
+    check_extent?: boolean;
```
ericboucher, line 10: is it possible to have check_extent in camel case?

packages/ogc/src/wms/layer.ts
DanielJDufour: done. this was removed

.gitignore
DanielJDufour: yup. done. good idea

DanielJDufour:
@ericboucher , this is ready for a final review.  After this is merged in, I can start to swap in the common functions/classes in another PR.

alerting/src/utils/server-utils.ts
```diff
@@ -2,6 +2,7 @@ import fetch from 'node-fetch';
 import moment from 'moment';
 import { xml2js } from 'xml-js';
 import { get, isEmpty, isString, union } from 'lodash';
+import { formatUrl } from "../../../packages/common";
```
ericboucher, line 5: I would expect this import to be something like:

`import { formatUrl } from "common";`

Not a blocker for this PR

packages/common/README.md
```diff
@@ -0,0 +1,25 @@
+# ogc
```
ericboucher, line 1: This readme seems to be a bit outdated

ericboucher (approved):

alerting/src/utils/server-utils.ts
DanielJDufour: okay.  I'll look into this.

packages/common/README.md
DanielJDufour: I updated README

packages/common/src/wcs/utils/index.ts
DanielJDufour: removed

PR #540, opened 2022-08-02
Remove other countries from build folder

PysarenkoDS-BWT:
Rewrite build script for deleting unused countries from build

Script get selected script from `env` - `REACT_APP_COUNTRY` and then delete other countries, if country not providing do nothing

### DIFF
--- a/config-overrides.js
+++ b/config-overrides.js
@@ -1,0 +1,30 @@
+const {
+  override,
+  disableChunk,
+  useEslintRc,
+  addWebpackPlugin,
+} = require('customize-cra');
+const path = require('path');
+const RemovePlugin = require('remove-files-webpack-plugin');
+
+const country = process.env.REACT_APP_COUNTRY;
+
+module.exports = override(
+  useEslintRc(path.resolve(__dirname, '.eslintrc')),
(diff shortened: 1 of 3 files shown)


config-overrides.js
```diff
@@ -0,0 +1,30 @@
+const {
+  override,
+  disableChunk,
+  useEslintRc,
+  addWebpackPlugin,
+} = require('customize-cra');
+const path = require('path');
+const RemovePlugin = require('remove-files-webpack-plugin');
+
+const country = process.env.REACT_APP_COUNTRY;
```
ericboucher, line 10: Should we take the lowercase version to be safe?

ericboucher, line 10: Should we take the lowercase version?
Also, you need to have a default country, or actually, do not filter the data folder if no country is specified via this environment variable

PysarenkoDS-BWT: I type it in description, it's already implemented on 15th line, if environment variable not provided RemovePlugin not added

ericboucher: perfect, thanks

ericboucher (approved):

PR #541, opened 2022-08-03
Add layer and date to direct link

PysarenkoDS-BWT:
Add `hazardLayerIds` and `date` params to direct link


alerting/src/alert-worker.ts
```diff
@@ -4,7 +4,11 @@ import { ANALYSIS_API_URL } from './constants';
...
 import { sendEmail } from './utils/email';
-import { getWCSCoverage, getWMSCapabilities } from './utils/server-utils';
+import {
+  formatUrl,
+  getWCSCoverage,
+  getWMSCapabilities,
+} from './utils/server-utils';
 
 async function processAlert(alert: Alert, alertRepository: Repository<Alert>) {
   const { baseUrl, serverLayerName, type } = alert.alertConfig;
```
ericboucher, line 14: I think hazardLayerIds should come from the `id` in alertConfig, and not serverLayerName, which is the representation we use to query the data.

alerting/src/alert-worker.ts
```diff
@@ -40,14 +44,19 @@ async function processAlert(alert: Alert, alertRepository: Repository<Alert>) {
...
+  const urlWithParams = formatUrl(prismUrl, {
+    hazardLayerIds: serverLayerName,
+    date: maxDate.toISOString().slice(0, 10),
+  });
+
   if (alertMessage) {
     const emailMessage = `
         Your alert${alertName ? ` ${alertName}` : ''} has been triggered.
 
         Layer: ${serverLayerName}
```
ericboucher, line 56: Let's use the `title` from the alertConfig instead of `serverLayerName` as the "Layer: ...."

PysarenkoDS-BWT: alertConfig don't have any `title`

type WCSConfig = {
  scale?: number;
  offset?: number;
  pixelResolution?: number;
};

export type AlertConfig = {
  type: string;
  serverLayerName: string;
  baseUrl: string;
  wcsConfig: WCSConfig;
};

what do you mean?

ericboucher: Technically, the AlertConfig type is the same as `WMSLayerProps` that you can find in the website code. So it might be worth adding `title` and `id` for this usecase

alerting/src/alert-worker.ts
```diff
@@ -45,7 +45,7 @@ async function processAlert(alert: Alert, alertRepository: Repository<Alert>) {
   url.searchParams.append('email', email);
 
   const urlWithParams = formatUrl(prismUrl, {
-    hazardLayerIds: serverLayerName,
+    hazardLayerIds: title,
```
ericboucher, line 48: This should be "id", not title

alerting/src/alert-worker.ts
```diff
@@ -40,21 +44,28 @@ async function processAlert(alert: Alert, alertRepository: Repository<Alert>) {
...
 
-    const emailHtml = `${emailMessage.replace(
-      /(\r\n|\r|\n)/g,
-      '<br>',
-    )} <br><br>To cancel this alert, click <a href='${url.href}'>here</a>.`;
+    const emailHtml = `${emailMessage
+      .replace(/(\r\n|\r|\n)/g, '<br>')
+      .replace(
+        urlWithParams,
+        `<a href="${urlWithParams}">${title}</a>`,
```
ericboucher, line 67: Could we simply insert this in the string formatting above instead of doing a replacement?

alerting/src/alert-worker.ts
```diff
@@ -40,21 +44,28 @@ async function processAlert(alert: Alert, alertRepository: Repository<Alert>) {
   url.searchParams.append('deactivate', 'true');
   url.searchParams.append('email', email);
 
+  const urlWithParams = formatUrl(prismUrl, {
+    hazardLayerIds: id,
```
ericboucher, line 48: This needs to be the alertConfig.id, not the alert.id

alerting/src/alert-worker.ts
PysarenkoDS-BWT: We can't insert in the string link with title different from URL

ericboucher (approved):

ericboucher (approved):

PR #542, opened 2022-08-03
Adding ICA and multidimensional risk for COUNTRY=rbd

wadhwamatic:
This PR configures new risk analysis data layers to the RBD deployment. All new layers are admin_level datasets
Screen Shot 2022-08-03 at 09 13 40


ericboucher (approved):
Looks good. Just as a general feedback, let's try to use `_ ` instead of `-` in file names for consistency :)

ericboucher (approved):

PR #544, opened 2022-08-08
Dates of validity for WMS layers COUNTRY=cambodia

JorgeMartinezG:
This PR enables PRISM to add additional dates based on the **validity** parameter. On the calendar rendering event, the feature will add more dates on the dates Array object, and will add the correct value to be requested to the WMS server. For instance, for dates _2022-01-11_ and _2022-01-21_ and a validity parameter of 1. The calendar array will have dates 2022-01-10, 2022-01-11, 2022-01-12 and 2022-01-20, 2022-01-21 and 2022-01-22 respectively

image


ericboucher:
@JorgeMartinezG I like the general direction and it seems to be working in the demo URL. However, I think we need to clean up the code a bit more to:
1. make it faster
2. make it more explicit (add comments, find more explicit names for the different variables)

wadhwamatic:
@JorgeMartinezG & @ericboucher:
I'm testing locally. When I tried to load an initial layer (rainfall anomaly), the request fails: "Time dimension value '2022-08-28' not valid for this layer".
When I then load another layer (10-day rainfall), the layer loads with 2022-08-21, but the calendar shows multiple other dates available from the kobo layer. Selecting a non valid date for the current layer does nothing though:
Screen Shot 2022-08-28 at 22 47 38

src/config/types.ts
```diff
@@ -384,6 +384,9 @@ export class WMSLayerProps extends CommonLayerProps {
 
   @optional
   chartData?: DatasetProps; // If included, on a click event, prism will display data from the selected boundary.
+
+  @optional
+  validity?: number;
```
ericboucher, line 389: Add a comment here explaining what this parameter does

src/utils/server-utils.ts
```diff
@@ -302,15 +338,49 @@ export async function getLayersAvailableDates(): Promise<AvailableDates> {
...
+  const layerDates = await Promise.all([
     ...wmsServerUrls.map(url => getWMSCapabilities(url)),
     ...wcsServerUrls.map(url => getWCSCoverage(url)),
     ...pointDataLayers.map(async layer => ({
       [layer.id]: await getPointDataCoverage(layer),
     })),
   ]);
 
-  return merge({}, ...layerDates);
+  const mergedLayers: { [key: string]: number[] } = merge({}, ...layerDates);
```
ericboucher, line 349: Let's add some comments in this part of the code as well

ericboucher: Thinking about it more, the structure might be simpler if we do something like this:

const mergedLayers: { [key: string]: number[] } = merge({}, ...layerDates.map(layerDate -> expandDatesOverValidityPeriod(layerDate));

Then expandDatesOverValidityPeriod finds the corresponding LayerDefinition, if validity is a number -> it adds new dates. And for all layers, it transforms dates as number[] into DateItem[]

I believe this would make the code simpler and easier to expand upon once we add this feature for other layers.

wadhwamatic:
@JorgeMartinezG & @ericboucher -  I'd like your thoughts on how this was implemented. I'm don't think the validity value should be enforced as  (+/-) days, but instead just (+) or (-) which should be part of the configuration. I.e., if it's valid for 5 days after the 'start' date, the validity value should be 5. If it's for 5 days leading up to the 'start' date, the validity value should be -5. 

Also noting here that we should discuss how easily the WMS validity days concept can be applied to other layer types.

ericboucher:
@wadhwamatic

We could do something simple and have 3 different cases:
- `5` both ways
- `-5` leading to
- `+5` after
 
For other layers, I think what @JorgeMartinezG implemented would work for as well. The issue at the moment seems to be around performance

wadhwamatic:
That works @ericboucher. I don't think the +/- scenario exists at the moment, but good to have it for the future

src/config/types.ts
```diff
@@ -356,6 +356,17 @@ interface FeatureInfoProps {
   label: string;
 }
 
+export enum DatesPropagation {
+  FORWARD = 'forward',
+  BACKWARDS = 'backwards',
```
ericboucher, line 361: let's remove the "s"

src/config/cambodia/layers.json
```diff
@@ -126,6 +126,7 @@
   "rainfall_dekad": {
     "title": "10-day rainfall estimate (mm)",
     "type": "wms",
+    "validity": 1,
```
ericboucher, line 129: Don't forget to update this using your new structure @JorgeMartinezG

ericboucher (approved):

wadhwamatic:
@JorgeMartinezG - I don't know if it's just on my side, but the performance on this branch is very slow. Toggling on any layer is slugish. 

I also wanted to test this on the kobo flood layer which has more test data available (see my commit). I loaded the 10-day rainfall layer, followed by kobo flood report. The screen blanks out and never renders again even though the network tool seems to show that requests were complete.  

Screen Shot 2022-09-12 at 18 19 21

Screen Shot 2022-09-12 at 18 14 47

Issue #270, opened 2021-09-28
Allow 'run analysis' if no baseline layers are present


wadhwamatic:
If a deployment doesn't have a 'baseline' layer, the run analysis feature cannot be executed. We should disable the run analysis feature in this case. See screenshot below

![image](https://user-images.githubusercontent.com/3343536/135165659-192d1b28-b553-4236-8112-c0a5df79b68c.png)

wadhwamatic:
> How about running the analysis and giving the result by admin boundary? I think we should make that an option, even when there are other baseline layers

Yes, that would definitely work, though I think we're no longer 'running an analysis' per se. We're just calculating zonal stats, which is something we wanted to facilitate in general. We should revisit the UX for this and consider having zonal stats available at the layer level for raster data.

PR #545, opened 2022-08-09
Allow 'run analysis' if no baseline layers are present

PysarenkoDS-BWT:
if hazardDataType === RasterType.Raster now we check count of finding categories in `src/components/MapView/Layers/LayerDropdown.tsx` and if categories.length === 0 we switch to using admin boundary instead of Layer with admin_level_data type


ericboucher:
@PysarenkoDS-BWT in the issue https://github.com/WFP-VAM/prism-frontend/issues/270, we actually decided to make the baseline layer optional, which means that even if baseline layers exist, we still want to be able to select no.

So instead I think we should simply do this:
- make the baseline layer optional
- hide the baseline layer selection if none are available

ericboucher (changes requested):
See comments above

PR #455, opened 2022-04-16
Feature/fast api

vikhyatsharma:
From Flask to Fast API

PR #548, opened 2022-08-18
Migrate the api to fastAPI

laurentS:
This PR replaces #455.

The goal is also to replace `flask` with `fastAPI`, but it also:
- adds typing annotations (almost) everywhere
- adds extensive endpoint testing via [schemathesis](https://schemathesis.readthedocs.io/) so that all endpoints are tested against their openAPI schema description
- adds some reference tests with "known good" requests to avoid breaking changes.

Some of the tests are still broken.

TODO: add GDAL to docker images.


api-flask/app/main.py
```diff
@@ -1,58 +1,61 @@
...
-stats_model = api.model("Stats", stats_dic)
+@app.get("/")
+def healthcheck() -> str:
+    """Verify that the server is healthy."""
+    return "All good!"
 
 
 @timed
-@cache.memoize(3600)
+# @cache.memoize(3600)
```
ericboucher, line 58: What would you recommend to replace memoization?

ericboucher: The return value is expected to be the same when the args are the same

api-flask/app/errors.py
```diff
@@ -14,27 +17,29 @@ def logError(description, traceback=None):
...
+    status = ex.code if isinstance(ex, HTTPException) else 500  # type: ignore
+    response: Response = jsonify(error=str(ex), status=status)  # type: ignore
     response.status_code = status
     logError("HTTP Error %s: %s" % (status, str(ex)))
     return response
 
 
-def handle_error(e: Exception):
+# TODO: this is not used with fastapi
+def handle_error(e: Exception) -> Response:
```
ericboucher, line 36: if it's not used, let's remove it completely

api-flask/app/zonal_stats.py
```diff
@@ -2,24 +2,35 @@
...
+from shapely.ops import cascaded_union  # type: ignore
 
 logger = logging.getLogger(__name__)
 
 
 DEFAULT_STATS = ["min", "max", "mean", "median"]
 
 
-def get_wfs_response(wfs_params):
+class GroupBy(str, Enum):
```
laurentS, line 24: Does this make sense? I was trying to constrain the data in there more than with a mere `str`, but I don't know how many possible values exist for this?

api-flask/app/database/alert_model.py
```diff
@@ -1,8 +1,10 @@
 """Alert database model."""
 import datetime
+
+# from flask import json
```
ericboucher, line 4: 👣

api-flask/app/kobo.py
```diff
@@ -122,23 +120,25 @@ def parse_form_response(
...
     return form_data
 
 
-def parse_datetime_params():
+def parse_datetime_params(
+    begin_datetime_str: str, end_datetime_str: str | None = None
+) -> tuple[datetime, datetime]:
     """Transform into datetime objects used for filtering form responses."""
-    begin_datetime_str = request.args.get("beginDateTime", "2000-01-01")
+    # begin_datetime_str = request.args.get("beginDateTime", "2000-01-01")
```
ericboucher, line 138: 👣 ?

ericboucher: should we keep the default though?

laurentS: Yep, I've kept the default, but it's now explicitly on the api endpoint declaration in main.py:149, which means it also shows up in the docs

api-flask/app/zonal_stats.py
laurentS: Ok, I've switched this to a `NewType` definition, so that we still get strong type validation in the code, but the api will allow any string.

ericboucher (approved):

api-flask/docker-compose.test.yml
```diff
@@ -0,0 +1,23 @@
...
+      - POSTGRES_HOST=testdb
+      - POSTGRES_USER=postgres
+      - POSTGRES_DB=postgres
+      - POSTGRES_PORT=5432
+      - POSTGRES_PASSWORD=!ChangeMe!
+    command: uvicorn app.main:app --host 0.0.0.0 --port 80 --reload
+    # Infinite loop, to keep it alive, for debugging
+    # command: bash -c "while true; do echo 'sleeping...' && sleep 10; done"
+    #
+# run a full battery of test on all endpoints using https://pypi.org/project/schemathesis/
```
ericboucher, line 16: should this be at the top?

laurentS: I'm not even sure why I put it there in the first place, to be honest :thinking: but yes, top of the file with some extra explanations. Added.

ericboucher (approved):

api-flask/app/zonal_stats.py
```diff
@@ -28,39 +37,37 @@ def get_wfs_response(wfs_params):
     https://docs.geoserver.org/stable/en/user/services/wfs/reference.html
     """
     cql_filter = []
-    if "time" in wfs_params.keys():
-        from_date = datetime.strptime(wfs_params.get("time"), "%Y-%m-%d")
+    if "time" is not None:
```
ericboucher, line 40: this will always be true, no? @laurentS 
did you mean wfs_params.time maybe?

laurentS: Yep, I think you're right :)

PR #550, opened 2022-08-24
Use Popper for NavBar

reed1:
This PR changes the Popover component in NavBar to Popper to allow user to open another Menu or interact with Map / DateSelector without having to do additional click to close Popover menu

https://user-images.githubusercontent.com/1814324/186328752-9d7b938d-4f05-4d5a-a3c3-fd00d4f26d9f.mp4


wadhwamatic:
@reed1 - I'm not fully clear on what the issue you are addressing here is. Can you please explain in more detail what is currently not working as desired?

reed1:
@wadhwamatic in the current version: http://prism-549.surge.sh/, if you open flood menu, then you want to switch to rainfall menu, you'd have to click 2 times, one is to close flood menu, then another one to open rainfall menu. Also if you have a menu opened, you can't immediately interact with map (panning and zooming) without clicking once outside menu to close that menu.

In this version: http://prism-menu-popper.surge.sh/, if you open hazards menu, the you want to switch to vulnerability menu, you'd only have to click one time. And if you have a menu opened, you can immediately interact with map (panning and zooming) without having to close the menu first

ericboucher (approved):
Thanks @reed1, I do like that change and think it makes navigation smoother. I'll let @wadhwamatic give the final green light though

wadhwamatic:
Thanks for the review @ericboucher. Good to go from my side.

PR #553, opened 2022-08-31
Add circular progress when loading layer data

### LINKED ISSUE #552: Kobo forms - choosing a new date fails to load data
If a user chooses a WMS layers followed by a kobo layer, then deselects the WMS layer, date selection no longer works on the kobo layer.

Steps to reproduce:
1. On master, REACT_APP_COUNTRY=cambodia
2. Activate layer Rainfall amount > Rainfall aggregate 10-day
3. Activate layer Exposure > Flood report
4. De-activate rainfall aggregate layer
5. From the date picker, choose another date. The URL correctly updates, but no request is sent and no data is loaded

All ok through step 3:
Screen Shot 2022-08-29 at 18 12 42

This is what I see for the date picker after de-activating the WMS layer (step 4):
Screen Shot 2022-08-29 at 18 13 19

After step 5, no data is loaded. Note the URL is correctly updated to Aug 10, but no request is sent (confirmed in the network tool)
Screen Shot 2022-08-29 at 18 13 42

cc @ericboucher

JorgeMartinezG:
This PR is a proposed solution to close #552 

boom

On a loading layer data event, PRISM will display a circular progress to let the user know that data is loading.
**How to test:** Using cambodia as country deployment
- Select **Exposure > Flood report**.
- Switch across dates in the calendar


ericboucher:
As a quick comment, we've been moving away from blocking circular progress indicators to use legend bars instead, see https://github.com/WFP-VAM/prism-frontend/pull/464

ericboucher (changes requested):
I don't think this change is needed since we already have the loading bar in the legends.
@wadhwamatic what do you think?

PR #554, opened 2022-08-31
Fix url replace error COUNTRY=cambodia

### LINKED ISSUE #552: Kobo forms - choosing a new date fails to load data

JorgeMartinezG:
fixes #552 

**How to test:** Using cambodia
- Select a WMS Layer.**rainfall_dekad**
- Select a PointDataLayer, like Kobo or EWS.
- Verify that the Url is correctly generated

### DIFF
--- a/src/config/cambodia/prism.json
+++ b/src/config/cambodia/prism.json
@@ -37,9 +37,6 @@
       ],
       "early_warning": [
         "ews_remote"
-      ],
-      "flood_events": [
-        "hydra_s1"
       ]
     },
     "rainfall": {
--- a/src/utils/url-utils.ts
+++ b/src/utils/url-utils.ts
@@ -59,15 +59,29 @@ export const useUrlHistory = () => {
     return updatedUrl.join(',');
   };
 
-  const removeLayerFromUrl = (
-    layerKey: UrlLayerKey,
-    layerId: string,
-  ): string => {
+  const removeLayerFromUrl = (layerKey: UrlLayerKey, layerId: string) => {
+    const removeKey = (): string => {
+      urlParams.delete(layerKey);
+
+      if (layerKey === UrlLayerKey.HAZARD) {
+        urlParams.delete('date');
+      }
+
+      return urlParams.toString();
+    };
+
     const urlLayers = urlParams.get(layerKey);
 
     const selectedLayersUrl = urlLayers !== null ? urlLayers.split(',') : [];
 
-    return selectedLayersUrl.filter(l => l !== layerId).join(',');
+    const filteredSelectedLayers = selectedLayersUrl
+      .filter(l => l !== layerId)
+      .join(',');
+
+    const urlToReplace =
+      filteredSelectedLayers === '' ? removeKey() : filteredSelectedLayers;
+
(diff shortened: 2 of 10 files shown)


wadhwamatic:
Working as expected. Thanks @JorgeMartinezG . Requested a review from @ericboucher too

src/utils/url-utils.ts
```diff
@@ -60,28 +60,24 @@ export const useUrlHistory = () => {
...
+    if (filteredSelectedLayers === '') {
+      urlParams.delete(layerKey);
 
-    replace({ search: urlToReplace });
+      if (layerKey === UrlLayerKey.HAZARD) {
+        urlParams.delete('date');
+      }
+    } else {
+      urlParams.set(layerKey, filteredSelectedLayers);
+    }
```
ericboucher, line 78: Can you add some comments to this part, please?

src/config/cambodia/prism.json
```diff
@@ -37,9 +37,6 @@
       ],
       "early_warning": [
         "ews_remote"
-      ],
-      "flood_events": [
```
ericboucher, line 41: no more flood events?

ericboucher (approved):
Looks good, let's add some comments to the function and you can merge after that @JorgeMartinezG

src/utils/url-utils.ts
```diff
@@ -60,28 +60,27 @@ export const useUrlHistory = () => {
...
       .join(',');
 
-    const urlToReplace =
-      filteredSelectedLayers === '' ? removeKey() : filteredSelectedLayers;
+    // If the list of layers is empty, remove the layerKey from the url.
+    if (filteredSelectedLayers === '') {
+      urlParams.delete(layerKey);
 
-    replace({ search: urlToReplace });
+      // For hazard layer, remove also the date.
```
ericboucher, line 75: Could we simply delete the date every time when the list is empty and a date exists?

wadhwamatic: > there is a possibility that a baselinelayer (admin level data) might have dates. I think there has been previous discussions about including dates in an admin level data layer

correct - https://github.com/WFP-VAM/prism-frontend/pull/555

wadhwamatic: we'll need to revisit the logic once admin level dates are incorporated

PR #555, opened 2022-08-31
Allow admin level with dates; COUNTRY=rbd

wadhwamatic:
This PR adds date parameters to admin level data layers in PRISM. Available dates are listed in a json file and individual datasets for given date are stored with a filename following YYYY-MM-DD format. These attributes are included in layers.json. For example:
  "ch_2021_03": {
    "title": "Phase classification",
    "type": "admin_level_data",
    "path": "data/rbd/ch/{YYYY_MM_DD}.json",
    "date_url": "data/rbd/ch/dates.json",
    "data_field": "phase_class",

The result is a admin level layer which includes date selection. See screenshot below.

Note, as of today (31-Aug), this feature breaks date intersections and the calendar picker. @reed1 to work with @JorgeMartinezG and @ericboucher to address this aspect

Screen Shot 2022-09-14 at 20 03 47


src/utils/server-utils.ts
```diff
@@ -289,6 +292,18 @@ async function getPointDataCoverage(layer: PointDataLayerProps) {
   return possibleDates;
 }
 
+async function getAdminLevelDataCoverage(layer: AdminLevelDataLayerProps) {
+  const url = layer.dateUrl;
```
ericboucher, line 296: let's do `const {dateUrl} = layer;`

public/data/myanmar/pop/pop_2022-07-21.json
```diff
@@ -0,0 +1,8 @@
+{
```
ericboucher, line 1: Let's use underscore in file names when possible -> `2022_07_21`

src/components/MapView/Layers/AdminLevelDataLayer/index.tsx
```diff
@@ -34,7 +35,10 @@ function AdminLevelDataLayers({ layer }: { layer: AdminLevelDataLayerProps }) {
 
   const boundaryId = layer.boundary || firstBoundaryOnView(map);
 
-  const layerData = useSelector(layerDataSelector(layer.id)) as
+  const selectedDate = useDefaultDate(layer.id);
+  // use undefined date for layers without date dimension
+  const date = layer.dateUrl ? selectedDate : undefined;
```
ericboucher, line 40: I don't think this step is needed. Technically, you could always pass in `selectedDate` and handle the missing layer.dateUrl in `loadLayerData`, no?

public/data/myanmar/pop/pop_2022-07-21.json
wadhwamatic: removed the myanmar test data and updated rbd to use underscores

wadhwamatic:
> @wadhwamatic are we expecting a lot of dates for admin layer data? If not, which would be my guess, we could simplify things and have the dates directly in the config file as:
> 
> `dates: [date1, date2]`, instead of requiring a different json file to hold the dates.
> 
> The performance also seems to be sub-par, but it might be due to the size of the geojsons for RBD.

Good point re: dates. I'll shift it to the config. Thanks!

Loads ok when deployed

ericboucher (approved):

src/utils/server-utils.ts
```diff
@@ -420,12 +432,20 @@ export async function getLayersAvailableDates(): Promise<AvailableDates> {
     (layer): layer is PointDataLayerProps => layer.type === 'point_data',
   );
 
+  const adminWithDateLayers = Object.values(LayerDefinitions).filter(
```
ericboucher, line 435: Maybe this could be merged with pointDataLayers to avoid looping over the same data twice?

PR #557, opened 2022-09-04
Add basic auth for kobo endpoints using a database table

ericboucher:
This PR adds a basic "user_info" table to contain username, password, and access parameters for users and a modal to request authentication in the frontend.

A few notes and potential improvements:
- better handle wrong password input and allow users to re-enter it
- we should change the message that says `Data not found for provided date: 2022-08-19`. Would we have an easy way to filter available dates by admin area as well? So that we could avoid this message altogether?

TODO:
- [x] activate salting
- [x] update Kobo urls (currently pointing to localhost)


api-flask/app/generate_jwt.py
```diff
@@ -0,0 +1,55 @@
+import json
```
ericboucher, line 1: @JorgeMartinezG we could repurpose that to generate a list of users with passwords and add them to the DB

ericboucher: The bbox needs to be entered into the "acccess" json:

we then get is as `user_info["access"]["bbox"]`

api-flask/docker-compose.yml
```diff
@@ -11,6 +11,7 @@ services:
       - WORKERS_PER_CORE=1
       # Kobo credentials
       - KOBO_USERNAME=${KOBO_USERNAME:-kobo_user}
+      - JWT_SECRET_KEY=${JWT_SECRET_KEY:-random_string}
```
ericboucher, line 14: not needed in this version

api-flask/requirements.txt
```diff
@@ -21,5 +21,6 @@ psycopg2==2.9.3
 sqlalchemy==1.4.40
 sqlalchemy2-stubs==0.0.2a25
 python-dateutil==2.8.2
+PyJWT==2.4.0
```
ericboucher, line 24: not needed in this version

src/components/AuthModal/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  const boundaryData = useSelector(layerDataSelector(boundaryLayer.id)) as
+    | LayerData<BoundaryLayerProps>
+    | undefined;
+
+  const { t } = useSafeTranslation();
+
+  useEffect(() => {
+    const layersWithAuthRequired = selectedLayers.find(
+      layer => layer.type === 'point_data' && layer.tokenRequired,
```
ericboucher, line 53: rename token_required to auth_required

api-flask/app/database/user_info_model.py
```diff
@@ -0,0 +1,26 @@
...
+from sqlalchemy.ext.declarative import DeclarativeMeta, declarative_base
+from sqlalchemy.sql.sqltypes import Boolean
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+Base = declarative_base()
+
+
+class UserInfoModel(Base):  # type: ignore
```
ericboucher, line 16: @wadhwamatic can you think of other useful info/columns we might want to keep in here?
`country` maybe? noting that we want to limit usage of this functionality, but it might come in handy still.

I made "access" a json so that we can use this table for other config/limitations in the future

ericboucher:
@JorgeMartinezG to test you can simply add this user to your local user_info table:

Screen Shot 2022-09-04 at 11 46 34 PM

api-flask/app/auth.py
```diff
@@ -0,0 +1,70 @@
...
+auth_db = AuthDataBase()
+
+if auth_db is None:
+    depends = lambda *_: True
+
+
+def verify_hash(password: str, saved_salt: str) -> bytes:
+    """Verify password hash."""
+    # Salt is in utf-8 string I need to encode it in Base64 and then decode the Base64 to bytes
+    if saved_salt.lower() == "false":
```
ericboucher, line 22: Temporarily bypass the salting process for development purposes.

api-flask/app/auth.py
```diff
@@ -0,0 +1,70 @@
...
+    """Validate user info."""
+    if auth_db is None:
+        return {"access": None}
+
+    user_info = auth_db.get_by_username(username=credentials.username)
+    print(user_info)
+    if not user_info:
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="User not found",
```
laurentS, line 46: Returning a different error message for username vs password allows an attacker to check if a username is registered. Ideally, you would return the same error message in both cases.
Also, since you do less processing when the username is wrong vs valid, it is theoretically possible to use timing attacks to determine if a username is registered on the site.

api-flask/app/auth.py
```diff
@@ -0,0 +1,70 @@
...
+            headers={"WWW-Authenticate": "Basic"},
+        )
+
+    is_correct_password = False
```
laurentS, line 50: The typing of `is_correct_password` is a bit confusing, it's a `bool` here, but `verify_hash` returns `bytes`.

api-flask/app/database/user_info_model.py
```diff
@@ -0,0 +1,26 @@
...
+    """Alert ORM that defines a table."""
```
laurentS, line 17: Leftover copy/paste from another model?

laurentS (commented):
I left a few comments. I would also add typing annotations where possible, at least on return values of functions to ensure `mypy` flags problems.

api-flask/app/auth.py
ericboucher: the result of verify_hash is compared to another value -> bool

src/context/layers/point_data.ts
```diff
@@ -80,13 +82,19 @@ export const fetchPointLayerData: LazyLoader<PointDataLayerProps> = () => async
     dataUrl.includes('?') ? '&' : '?'
   }${dateQuery}&${queryParamsToString(additionalQueryParams)}`;
 
+  const fetchUrl = authRequired
```
ericboucher, line 85: @JorgeMartinezG these needs to be passed in the authorization header below

ericboucher: done

api-flask/app/kobo.py
```diff
@@ -202,6 +204,22 @@ def get_responses_from_kobo(
     return form_responses, form_labels
 
 
+def get_form_dates(form_url: HttpUrl, form_name: str, datetime_field: str):
```
ericboucher, line 207: @laurentS is there an easy way to add timed caching to this function? Say for 6hours?

api-flask/app/database/database.py
```diff
@@ -34,8 +40,15 @@ class AlertsDataBase:
     def __init__(self):
         """Alerts Database initializer."""
         _eng = create_engine(DB_URI)
-        self.session: Session = sessionmaker(_eng)()
-        logger.info("DB connection is initialized.")
+        try:
```
ericboucher, line 43: @laurentS I can't find a good way to test the db connection. When I do this and the database is down, Python simply hangs and the app doesn't start. Any ideas?

laurentS: I can think of a couple of ways:
- if you're running everything in docker compose, I think you can use the [`healthcheck` ](https://docs.docker.com/compose/compose-file/#healthcheck) option for the `postgres` container, and then `depends_on` with `service_healthy` for your python container. That way, you're sure that postgres is ready to accept connections when python is starting (but it doesn't mean they can talk to each other)
- otherwise, I think I would check access to the port for pg before creating the session. Something like this [SO answer](https://stackoverflow.com/a/19196218/1138710) would allow a quick check at python startup time.
- or maybe the SQLAlchemy [timeout](https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.create_engine.params.pool_timeout) option: `create_engine(..., connect_args={"timeout": 30})` before calling `sessionmaker`?

I guess ideally you'd combine all of them so that python code only starts when pg is ready, then at least have a reasonable timeout, and ideally poke the port first to check that connectivity is available?

ericboucher: I don't want this to be blocking the app, so the timeout option is actually the only one we need. Thanks for the tips!

wadhwamatic:
@ericboucher - thanks for the help to figure out DB setup locally. After testing a bit, just two pieces of feedback:
1. After entering incorrect credentials, it would be better to keep the login modal active and specifically say that the credentials are invalid. Right now, it just closes the modal and shows a generic failed to load message (screenshot 1 below)
2. If the login fails, can we not show the layer as toggled on in the navigation and legend? Currently the UI still shows the layer as active even though it shouldn't have loaded (screenshot 2)

Screen Shot 2022-09-13 at 17 40 06

Screen Shot 2022-09-13 at 17 41 43

ericboucher:
@wadhwamatic agreed. @JorgeMartinezG could you give this a shot? :)

wadhwamatic (approved):
Good to go! Let's create a new issue for the credential frontend changes needed that were mentioned earlier

PR #559, opened 2022-09-12
Configurable API URLs

ericboucher:
Make it easier to configure the analysis API using two environment variables:
Setting `REACT_APP_LOCAL` to true (or any value really), will make the frontend point to `localhost`
Otherwise, you can set `REACT_APP_API_URL` directly to replace the default URL of `prism-api.ovio.org`.

For the alerts backend, you can now override the backend endpoint using the `API_URL` environment variable.


alerting/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:12
+FROM node:16
```
JulienPathtech, line 1: Did you tried node:20 ? ^^

ericboucher: I have not, but we upgraded the rest of the app is using node:16 so I upgraded the alerting module to be consistent with the rest

laurentS (approved):
Looks good to me. My only question is about whether there is a way to avoid having to set 2 variables to do local dev. In particular, hitting prism-api.ovio.org as a default implies not knowing which version of code is running there, meaning possible bugs due to version mismatches. My gut says the default should be localhost, but maybe I'm missing something...

ericboucher:
Fair point. The API changes a lot less often than the frontend and usually "adds" functionalities so this allows most devs that only touch on the frontend to not have to think about the backend at all. We can always change this in the future if it becomes cumbersome

PR #565, opened 2022-09-19
initial configuration for prism south sudan COUNTRY=southsudan

wadhwamatic:
Initial configuration for a demo for South Sudan. Deployed here: https://prism-demo-ssd.surge.sh/

To do:
- [x] Add admin 1 boundaries
- [x] Troubleshoot issue with IPC data layer (admin_level_data)

Screen Shot 2022-10-10 at 17 25 13


src/config/southsudan/layers.json
```diff
@@ -0,0 +1,1200 @@
...
+    "date_interval": "days",
+    "wcsConfig": {
+      "scale": 0.02,
+      "offset": -273,
+      "pixelResolution": 64
+    },
+    "opacity": 0.7,
+    "legend_text": "Daytime land surface temperature",
+    "legend": [
+      { "value": 0, "label": "0C - 2C", "color": "#1d2e83" },
```
ericboucher, line 1037: In general, since the colors are gradual, I think we should limit the number of values in the legend. In this case in particular, le's try to limit to 5-10 values top.

It's also worth doing that in a few other legends as well imho.

Screen Shot 2022-10-03 at 12 14 04 PM

ericboucher (approved):
Looks good overall. I would recommend spending some time to cleanup the legends while we're at it but I didn't see any issues otherwise

src/config/southsudan/layers.json
wadhwamatic: yup, working to refine this

PR #566, opened 2022-09-20
Fix bug with dates in URL in getPossibleDatesForLayer

JorgeMartinezG:
This PR fixes the following error on master when loading prism using an analysis url:

image

**How to test:** Using mozambique or any country deployment that contains adamts_buffers layer:

- Open the following link: http://localhost:3000/?analysisHazardLayerId=adamts_buffers&analysisAdminLevel=1&analysisStartDate=2020-08-02&analysisEndDate=2022-05-11&analysisStatistic=mean

### DIFF
--- a/src/utils/server-utils.ts
+++ b/src/utils/server-utils.ts
@@ -90,7 +90,7 @@ export const getPossibleDatesForLayer = (
     }
   };
 
-  return datesArray().map(d => d.displayDate);
+  return datesArray()?.map(d => d.displayDate);
 };
 
 export function formatUrl(
--- a/src/utils/server-utils.ts
+++ b/src/utils/server-utils.ts
@@ -90,7 +90,7 @@ export const getPossibleDatesForLayer = (
     }
   };
 
-  return datesArray()?.map(d => d.displayDate);
+  return datesArray()?.map(d => d.displayDate) ?? [];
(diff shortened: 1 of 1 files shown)


src/utils/server-utils.ts
```diff
@@ -90,7 +90,7 @@ export const getPossibleDatesForLayer = (
     }
   };
 
-  return datesArray().map(d => d.displayDate);
+  return datesArray()?.map(d => d.displayDate);
```
ericboucher, line 93: To be consistent, you should either retype the return of the function to be number[] | undefined, or return an empty array, eg.

`return datesArray()?.map(d => d.displayDate) ?? []`

ericboucher (changes requested):
Small typing request

src/utils/server-utils.ts
JorgeMartinezG: but the function should return undefined, based on this condition: https://github.com/WFP-VAM/prism-app/blob/master/src/components/MapView/index.tsx#L416

JorgeMartinezG: included in 814af14d895085c45569af957e8a1ecc05fe1623

ericboucher (approved):

PR #567, opened 2022-09-20
Update ODC endpoints to use WFP's prod environment

wadhwamatic:
This PR updates the configuration of all PRISM deployments to replace references to WFP's Data Cube deployment to make use of the final production URL. Replaces https://odc.ovio.org/ with https://api.earthobservation.vam.wfp.org/ows/

Two test files were also updated


ericboucher (approved):

ericboucher:
Looks great and staging is working well 🎉

PR #569, opened 2022-09-26
Get data from HDC country=mozambique

JorgeMartinezG:
This PR modifies the request to get data from the earth observation API (humanidatian data cube) instead of the static json file used for development.

**How to test:** Using mozambique country deployment.
- Select the **Rainfall Aggregate** or **NDVI** layer and click on any district. the chart will be rendered after success response from api according to the chart configuration.

**Notes:** The boundary geojson file was changed with the purpose of including the administrative pcodes that the api requires

image

### DIFF
--- a/src/context/datasetStateSlice.ts
+++ b/src/context/datasetStateSlice.ts
@@ -44,7 +44,7 @@ const initialState: DatasetState = {
 
 type BoundaryProps = {
   code: number;
-  urlPath: string;
+  level: string;
   name: string;
 };
 
@@ -55,6 +55,7 @@ export type AdminBoundaryParams = {
   url: string;
   serverLayerName: string;
   id: string;
+  dataField: string;
 };
 
 export type AdminBoundaryRequestParams = AdminBoundaryParams & {
@@ -74,33 +75,6 @@ export enum TableDataFormat {
   DATE = 'date',
   TIME = 'time',
 }
-
-const getDatasetFromUrl = async (
-  year: number,
-  params: AdminBoundaryRequestParams,
-  startDate: number,
-  endDate: number,
-): Promise<DataItem[]> => {
-  const { serverLayerName, url, id, boundaryProps } = params;
-
-  const { code: adminCode, urlPath } = boundaryProps[id];
-
-  const serverUrl = `${url}/${urlPath}/${year}.json`;
-
-  const resp = await fetch(serverUrl);
-  const results = await resp.json();
-
-  const filteredRows = results.DataList.filter(
-    (item: any) => item[id] === adminCode,
-  )
-    .map((item: any) => ({
-      date: moment(item.time).valueOf(),
-      value: item[serverLayerName],
-    }))
-    .filter(({ date }: DataItem) => date >= startDate && date <= endDate);
-
-  return filteredRows;
-};
 
 const createTableData = (
   results: DataItem[],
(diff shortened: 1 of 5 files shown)


ericboucher:
@JorgeMartinezG so no other countries are using this for now? Just making sure that this won't be breaking anything for other deployments

wadhwamatic:
We are starting with Mozambique but the intention is to roll this out for
all deployments. But the boundary files need to be reviewed case by case
before we deploy this feature.

wadhwamatic:
@JorgeMartinezG - as mentioned, there are a few things to address:

1. Add a legend to the chart
2. Allow for configuration of labels in the chart; i.e. we want to be able to use 'rainfall amount' and 'average' instead of 'rfh' and 'rfh_avg'
3. I still get an error message when I click on an admin area for rainfall anomaly / NDVI anomaly: "Cannot read properties of undefined (reading '0')"

Please also make sure to get changes from this PR before attempting to merge as there are multiple config changes forthcoming for Mozambique. https://github.com/WFP-VAM/prism-app/pull/572

wadhwamatic:
@JorgeMartinezG - it seems we're not correctly handling all four possible data layers with time-series data extraction available. This becomes visible when loading NDVI data, which is available up until 2022-11-09. But when I click on the chart, I see data until 2022-08-21, and the response shows data until 2022-01-09. Seems we're not getting the correct response, but also not plotting correctly? See screenshots below.

In addition, NDVI anomaly doesn't load at all.

Some additional nitpicking:
1. Reduce size of point for line charts.
2. Nice to have: allow for configuration of color for each data series

Screen Shot 2022-10-03 at 18 01 38

Screen Shot 2022-10-03 at 18 03 39

src/context/datasetStateSlice.ts
```diff
@@ -183,24 +160,67 @@ export const loadEWSDataset = async (
...
+  const { code: adminCode, level } = boundaryProps[id];
+
+  const endDateStr = endDate.format(DEFAULT_DATE_FORMAT);
+  const startDateStr = startDate.format(DEFAULT_DATE_FORMAT);
+
+  const hdcRequestParams = {
+    level,
+    admin_id: adminCode,
+    coverage: 'full',
+    vam: serverLayerName.includes('vim') ? 'vim' : 'rfh',
```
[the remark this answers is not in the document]
JorgeMartinezG: viq and rfq data is within the api response. API documentation specifies only **vim** and **rfh** as request parameters

JorgeMartinezG:
@wadhwamatic by making multiple requests, the api is always returning 2022-09-01 as most recent date: You can try by changing the **end** parameter within the request

https://api.earthobservation.vam.wfp.org/stats/admin/fetch?level=2&admin_id=1010614&coverage=full&vam=vim&start=2021-09-11&end=2022-12-31 gives

image

valpesendorfer:
As the API is currently under development, we can't guarantee that the timesteps served through the API are in lockstep with what's published through OWS.

Currently the stats API serves until `2022-09-01` (which is specified in the API response) while OWS has data up to `2022-09-11`.

Please note that specifying an end date is optional, if omitted you get the full range of the timeseries from the specified start date.

wadhwamatic:
thanks for the clarifications @valpesendorfer!

valpesendorfer:
One related thing: we're in the process of moving the API behind the WFP API gateway. I'd think that has a bearing on this PR.

wadhwamatic:
@JorgeMartinezG - thanks for the changes to make chart styles more configurable. as mentioned yesterday, NDVI anomaly fails to load any chart data. See screenshot

Screen Shot 2022-10-04 at 10 09 32

JorgeMartinezG:
@wadhwamatic pushed fix

wadhwamatic:
thanks @JorgeMartinezG. looks good from my side. requesting a review from @ericboucher

src/context/datasetStateSlice.ts
```diff
@@ -179,24 +156,70 @@ export const loadEWSDataset = async (
   return new Promise<TableData>(resolve => resolve(tableDataWithEWSConfig));
 };
 
+const fetchHDC = async (
```
ericboucher, line 159: Let's add some comments for non-WFP people please :)
What's HDC? What's the expected data format?

src/context/datasetStateSlice.ts
```diff
@@ -67,73 +68,46 @@ export type DatasetRequestParams =
...
-  )
-    .map((item: any) => ({
-      date: moment(item.time).valueOf(),
-      value: item[serverLayerName],
-    }))
-    .filter(({ date }: DataItem) => date >= startDate && date <= endDate);
-
-  return filteredRows;
-};
+export const PREFIXES = { col: 'd', date: 'Date' };
```
ericboucher, line 79: Let's give this a more explicit name please

ericboucher (approved):
@JorgeMartinezG looks good thanks! Just a few comments to make the changes more readable / documented :)

ericboucher (approved):

ericboucher:
Thanks @JorgeMartinezG, gtg!

PR #572, opened 2022-09-30
layer configuration updates for mozambique; COUNTRY=mozambique

wadhwamatic:
This PR includes multiple configuration changes for Mozambique including new color ramps and reorganization of layers within navigation menus.

Deployed here:
https://prism-moz.surge.sh/

Screen Shot 2022-10-05 at 11 32 35


ericboucher:
@wadhwamatic this does not seem to be working as expected with the new API endpoint. Looks like the blended data is not loading

wadhwamatic:
> @wadhwamatic this does not seem to be working as expected with the new API endpoint. Looks like the blended data is not loading

@ericboucher - It seems there's an issue on the OWS side. Dates are listed as available through September 2022, but the response is empty. Data for these layers in August loads as expected. See http://prism-572.surge.sh/?hazardLayerIds=precip_blended_dekad&date=2022-08-21 

cc @valpesendorfer

valpesendorfer:
This happens because all blended datasets for different countries are in the same product, exposed through the same layer.

So there is data after August for Cuba, so the dates show in the timeline but of course no data shows in Mozambique.

src/config/mozambique/layers.json
```diff
@@ -4,8 +4,12 @@
     "path": "data/mozambique/moz_bnd_adm1_WFP.json",
     "opacity": 0.8,
     "admin_code": "source_id",
-    "admin_level_names": ["adm1_name"],
-    "admin_level_local_names": ["adm1_name"],
+    "admin_level_names": [
```
ericboucher, line 7: not a huge deal but it looks like we don't have the same linter rules, which is created a bit of xtra changes

wadhwamatic: @ericboucher - Thanks for the fix. The config files to get unwieldy and challenging to navigate quickly with multiple parameters split over new lines. Thoughts on revisiting the linter rules to make this a bit easier to navigate?

wadhwamatic:
ready to merge @ericboucher ?

src/context/layers/admin_level_data.ts
```diff
@@ -48,6 +49,13 @@ export function getAdminLevelDataLayerData(
   const adminBoundariesLayer = layerDataSelector(adminBoundaryLayer.id)(
     getState(),
   ) as LayerData<BoundaryLayerProps> | undefined;
+  // TEMP - for Mozambique, add a 10s wait time to load admin boundaries which are very large
+  if (
+    safeCountry === 'rbd' &&
```
wadhwamatic, line 54: does this need to be updated for mozambique?

ericboucher (approved):

PR #573, opened 2022-10-03
TypeError bug fix in ReportDoc

echaidemenos:
This PR:

- Fixes warning `Image should receive either a "src" or "source" prop`

### DIFF
--- a/src/components/MapView/Download/report.tsx
+++ b/src/components/MapView/Download/report.tsx
@@ -14,7 +14,7 @@ import {
 } from '@material-ui/core';
 import { useSelector } from 'react-redux';
 import { ArrowBack } from '@material-ui/icons';
-import { PDFViewer, PDFDownloadLink } from '@react-pdf/renderer';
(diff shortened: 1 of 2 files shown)


src/components/MapView/Download/report.tsx
```diff
@@ -40,23 +40,26 @@ function Report({ classes, open, reportType, handleClose }: ReportProps) {
...
-        reportType === ReportType.Storm
-          ? `Storm Report (${eventDate})`
-          : `Flood Report (${eventDate})`
-      }
-      mapImage={mapImage}
-      tableData={analysisData}
-    />
-  );
+  const document =
+    mapImage !== null ? (
```
ericboucher, line 44: Could this happen in `ReportDoc` instead at the image level? It might make things a bit cleaner. wdyt?
It's fine like that otherwise, thanks for fixing this

ericboucher (approved):

PR #576, opened 2022-10-04
add CH data for testing; COUNTRY=rbd

wadhwamatic:
This PR adds additional data layers for Cadre Harmonisé (CH) data for the regional deployment of PRISM in Dakar (RBD). There are two things I will be testing with this PR:
1. Add validity concept to admin_level_data layers
2. Group data layers so that admin 1 and admin 2 level layers can be showed as stacked layers. The idea is that the admin 2 layer will be on top, and where data doesn't exist at admin 2, the admin 1 layer below is visible


src/components/MapView/Layers/AdminLevelDataLayer/index.tsx
```diff
@@ -28,15 +28,21 @@ import {
 import { getRoundedData } from '../../../../utils/data-utils';
 import { useSafeTranslation } from '../../../../i18n';
 import { fillPaintData } from '../styles';
+import { availableDatesSelector } from '../../../../context/serverStateSlice';
```
ericboucher, line 31: @JorgeMartinezG can you review this fix for the validity period and confirm it makes sense to you?

src/utils/server-utils.ts
```diff
@@ -52,7 +52,7 @@ export const getRequestDate = (
     date => date.displayDate === selectedDate,
   );
   if (!dateItem) {
-    return selectedDate;
+    return layerAvailableDates[layerAvailableDates.length - 1].queryDate;
```
ericboucher, line 55: @JorgeMartinezG what do you think of this change?

Also, in this case `if (!layerAvailableDates)`, should we return `undefined` instead?

wadhwamatic:
@sifelix and @ajanim - I added a couple of changes to the Cambodia deployment in this branch to test multiple dates with a few scenarios:
1. I added a time dimension to the admin_level_data layer called 'idpoor'. It includes two rounds of data, each are valid for 150 days
2. I also added days of validity to the 'rainfall_dekad' layer (a WMS layer)
3. Cambodia already has a point data layer called 'ews_remote'

With these three layers, and any existing WMS layer, we've got a few scenarios to test dates. Let me know if you need anything else for this. Thanks!

wadhwamatic:
Closing this and using https://github.com/WFP-VAM/prism-app/pull/678 instead

PR #589, opened 2022-10-13
Update sierra leone configuration; COUNTRY=sierraleone

wadhwamatic:
This PR makes updates to the demo version of PRISM Sierra Leone. Color ramps have been updated to match global styles and the menu has reorganized by grouping similar climate layers. The admin 1 boundary file has also been added.

Deployed here: https://prism-sierra-leone-demo.surge.sh/
Screen Shot 2022-10-12 at 18 05 03


ericboucher (approved):

src/config/sierraleone/layers.json
```diff
@@ -17,6 +17,371 @@
...
+    "server_layer_name": "rfh_dekad",
+    "additional_query_params": {
+      "styles": "rfh_16_0_300"
+    },
+    "base_url": "https://api.earthobservation.vam.wfp.org/ows/",
+    "date_interval": "days",
+    "opacity": 0.7,
+    "legend_text": "Estimate of precipitation over a 10-day period derived from CHIRPS (UCSB Climate Hazards Group). https://www.chc.ucsb.edu/data/chirps",
+    "legend": [
+      { "value": 0, "label": "0 mm", "color": "#fafafa" },
```
ericboucher, line 50: Would it be easy to limit the number of labels in these legends?
Maybe merge a few that are out of usual range for this country?

wadhwamatic: We can't limit this color ramp any further but will limit where we can which is mostly on the temperature layers

PR #601, opened 2022-10-19
Adding new COUNTRY=colombia

wadhwamatic:


package.json
```diff
@@ -150,6 +150,6 @@
     "eslint-import-resolver-typescript": "^2.0.0"
   },
   "engines": {
-    "node": "16.x"
+    "node": "14.x"
```
ericboucher, line 153: Why this change?

wadhwamatic: that was unintentional. thanks for fixing

src/config/colombia/prism.json
```diff
@@ -0,0 +1,41 @@
...
+    "rainfall": "icon_rain.png",
+    "vegetation": "icon_veg.png",
+    "temperature": "icon_climate.png",
+    "risk": "icon_impact.png",
+    "tropical_storms": "icon_tropical_storm.png",
+    "vulnerability": "icon_vulnerable.png"
+  },
+  "serversUrls": {
+    "wms": [
+      "https://geonode.wfp.org/geoserver/prism/wms/",
```
ericboucher, line 14: do we need geonode?

wadhwamatic: yes, colombia will use geonode

wadhwamatic:
@ericboucher - I'm getting an error when I try to run an analysis again (testing locally). 

Here's the payload of the request:
geojson_out: false
geotiff_url: "https://api.earthobservation.vam.wfp.org/ows/?service=WCS&request=GetCoverage&version=1.0.0&coverage=rfh_dekad&crs=EPSG%3A4326&bbox=-81.7%2C-4.2%2C-66.8%2C13.4&width=3220&height=3812&format=GeoTIFF&time=2022-10-11"
group_by: "MPIO_CDPMP"
zones_url: "https://prism-admin-boundaries.s3.us-east-2.amazonaws.com/col_municipios.json"

My guess is that there's a problem with the admin code (MPIO_CDPMP). In this case, it 'looks' like numeric data but is a string and should be treated as such. This is another of those issues where it's difficult to troubleshoot the issues with admin boundaries (https://github.com/WFP-VAM/prism-app/issues/604). 

I think we have some inconsistency with how the group_by function works and the mapping of admin level data to boundaries works.

Screen Shot 2022-10-21 at 11 53 19

ericboucher (approved):

PR #606, opened 2022-10-20
Use correct wcsconfig parameter for rasters without time support COUNTRY=mozambique

### LINKED ISSUE #600: Exposure analysis fails - incorrect handling of WCS time attribute
When I try to run an exposure analysis, I get an error 'g.map is not a function'. Here is the server response:

> The file you requested is not available - https://api.earthobservation.vam.wfp.org/ows/?service=WCS&request=GetCoverage&version=2.0.0&coverageId=wp_pop_cicunadj&subset=Long(30.2156,40.84062)&subset=Lat(-26.86704,-10.47219)&subset=time("2022-01-25")

I see that the 'time' attribute is causing the issue as 'wp_pop_cicunadj' does not have a time attribute. The configuration of the layer is correct with 'time_support' set to false. It seems likely this PR https://github.com/WFP-VAM/prism-app/pull/586 introduced this issue.

To reproduce, go to the PRISM Mozambique demo page, select tropical storm on 2022-01-25 and then click exposure analysis.
https://prism-moz.surge.sh/?hazardLayerIds=adamts_buffers&date=2022-01-25 

Screen Shot 2022-10-19 at 09 04 45

JorgeMartinezG:
Closes #600 

This PR replaces the parameter to remove the time parameter used for the geotiff_url within a /stats request to the API..

**How to test:**
- Select Tropical Storms layer and press the **Exposure analysis** button. Verify within the network tab that the `geotiff_url` does not have the time parameter.
- Run a normal analysis and verify that the parameter appears.


wadhwamatic (approved):
thanks for the fix jorge.

PR #614, opened 2022-10-21
layer config updates; COUNTRY=myanmar

wadhwamatic:
This PR makes a few updates to the live version of PRISM Myanmar:
1. Update color ramps for climate layers from WFP's Humanitarian Data Cube
2. Add new admin level data on women and children
3. Reconfigure navigation, grouping floods and storms, and also grouping rainfall layers
4. Fix admin boundary label configuration

The live version is here (WFP credentials required):
https://prism.mmr.wfp.org/

Screen Shot 2022-10-20 at 23 10 29


ericboucher:
@JorgeMartinezG @wadhwamatic I believe the tests are failing because you added authentication to Myanmar which is the default country, so the frontend tests are failing.

Two easy options I can think of:
- change the default country to one that does not require authentication
- update the tests to allow them to bypass authentication

wadhwamatic:
@ericboucher - yes we should not make myanmar the default. i made that change but test still fail which i assume is because we're still using myanmar in this instance. i think we should 1) bypass auth flag for the test and 2) not auto deploy if the auth flag is true. the point is that these shouldn't be publicly accessible.

wadhwamatic:
thanks for the tests / auth fix @ericboucher. I think this is GTG now

ericboucher (approved):

PR #626, opened 2022-11-08
Use prism-common WMS class and wms functions

DanielJDufour:
This is the next step in consolidating OGC and other common functionality into one package.  This PR implements the WMS class and wms functions found in `packages/common`.

Note: I couldn't use the WMS Class for everything, because React functional components are synchronous.  Technically, I could have implemented a useEffect, but it would have made things more complicated than it is worth it at the moment.  I'll consider how to simplify things even more going forward.


.github/workflows/build.yml
```diff
@@ -78,8 +78,9 @@ jobs:
             ${{ runner.os }}-yarn-
 
       - name: Install dependencies if needed.
-        if: steps.yarn-cache.outputs.cache-hit != 'true'
-        run: yarn install --frozen-lockfile
+        run: |
```
ericboucher, line 81: Why did you need to remove the cache-hit? Maybe you need to include the new lock file hash in the key instead? (line 76)

ericboucher:
Looking good, thanks! Would this be the right PR to move frontend related files and folders into a top-level `frontend` folder?

.github/workflows/build.yml
DanielJDufour: great catch.  I'm not an expert with github workflows, but I'll try to make that change and see if it works.

DanielJDufour:
@ericboucher , I think it'd be best to create the frontend folder in a separate PR

ericboucher (approved):

PR #628, opened 2022-11-08
Create boundary relations for dropdown COUNTRY=cambodia

JorgeMartinezG:
this PR modifies the rendering of the dropdown menu that lists the boundaries at different levels. The levels are created using the number and the order of the elements in the adminLevelNames parameter.

The BoundaryDropdown component enabling multiple interaction with the map, by enhancing the MapInteraction enum. This is needed for the dropdown select within the alert system


src/components/MapView/Layers/BoundaryLayer/index.tsx
```diff
@@ -40,12 +46,30 @@ function BoundaryLayer({ layer, before }: ComponentProps) {
...
+    : layer.adminLevelLocalNames;
+
+  const isPrimaryLayer = isPrimaryBoundaryLayer(layer);
+
+  useEffect(() => {
+    if (!data || !isPrimaryLayer) {
+      return;
+    }
+
+    const boundaryRelationData = loadBoundaryRelations(
```
ericboucher, line 61: You could run this once and use both adminLevelNames and adminLevelLocalNames values in your mapping so that this doesn't have to run when users switch language, wdyt?

ericboucher:
Nice start @JorgeMartinezG, I like where this is going. I do have some feedback:

- The code seems a lot more complicated than it needs to be, or I might be missing something about the different functionalities?
- These changes should also apply to the Alert dropdown selector so that we don't have two distinct objects and the alert selector can benefit from this update
- the UI is worse than before, let's try to stick to the UI, fonts and colors we had before? You can ping Garance on Slack if you need help.
- Let's have the results organized by alphabetical order (on the first admin level)

wadhwamatic:
@JorgeMartinezG - a few inputs from my side:
1. If I double-click in the search area of the go to feature, I can no longer search and have to either close the feature or choose an area by click on it in the list of admin areas
2. I agree with Eric that the new look isn't as clean as what we had before. But we did talk about a way to better show that there are multiple levels now. Perhaps match what was previously there (font size and type) but use indentation and maybe a hyphen prefix to show levels? Something like:

> Screen Shot 2022-11-15 at 16 32 46

JorgeMartinezG:
Sorted elements, simplified logic and stored adminlevelnames in an object

JorgeMartinezG:
> @JorgeMartinezG - a few inputs from my side:
> 
> 1. If I double-click in the search area of the go to feature, I can no longer search and have to either close the feature or choose an area by click on it in the list of admin areas
> 2. I agree with Eric that the new look isn't as clean as what we had before. But we did talk about a way to better show that there are multiple levels now. Perhaps match what was previously there (font size and type) but use indentation and maybe a hyphen prefix to show levels? Something like:
> 
> > Screen Shot 2022-11-15 at 16 32 46

Tried including dashes. IMO it does not look good.

image

wadhwamatic:
@JorgeMartinezG - The UI looks better to me. But the load time is a problem. I tried making a build and testing that (https://medical-inventionz.surge.sh), but also am facing very slow load times, and on many occasions, the UI never fully loads. I also tried to test with Mozambique and had the same problem. We'll likely need to revisit this when you're back unless you can advise on some quick solutions here.

JorgeMartinezG:
> You could run this once and use both adminLevelNames and adminLevelLocalNames values in your mapping so that this doesn't have to run when users switch language, wdyt?

@wadhwamatic  This could be one cause why the dropdown is taking more time to render.

wadhwamatic:
@JorgeMartinezG - you can see a significant improvement in the initial page load now. Toggling to Khmer is fairly slow though. It seems that all components are loaded before the screen renders in Khmer. Are we able to make this sequential and add a loader? I believe this is how it's managed for the English version.

JorgeMartinezG:
commit [5c35056](https://github.com/WFP-VAM/prism-app/pull/628/commits/5c35056e02c8aa8c2f27d348fab0737f4ae9cc31) reduces dropdown loading time from 10 seconds to 500 ms. However, is important to mention that this loading overhead might be present when trying to make the level 2 and level 1 multipolygon boundaries from level 3 ones

wadhwamatic:
> commit [5c35056](https://github.com/WFP-VAM/prism-app/pull/628/commits/5c35056e02c8aa8c2f27d348fab0737f4ae9cc31) reduces dropdown loading time from 10 seconds to 500 ms. However, is important to mention that this loading overhead might be present when trying to make the level 2 and level 1 multipolygon boundaries from level 3 ones

@ericboucher - the change Jorge made here made this process 20x faster but from what I understand, the approach will slow down future attempts to construct admin boundaries based on unions of lower level boundaries. Since we're not doing this yet, and this PR has been pending for a long time, I'm hoping we can go ahead and merge this. Thanks for adding your review when you can.

ericboucher:
@JorgeMartinezG could you add a few screenshots to this PR's first comment?
This is updating the design for both the GoTo and alert functions, correct?

frontend/src/context/mapStateSlice/index.ts
```diff
@@ -21,6 +22,7 @@ export type MapState = {
   // Note that layerData is mainly for storing vector map data.
   // Tile image loading for raster layer is tracked separately on mapTileLoadingStateSlice
   loadingLayerIds: LayerKey[];
+  relationData: BoundaryRelationsDict;
```
ericboucher, line 25: let's call this boundaryRelationData maybe? relationData isn't really telling out of context

ericboucher (approved):
Looks good, thanks!

I would recommend taking the time to add a few docstrings around functions and some comments as some of the steps will not be very clear to a new dev.

frontend/src/context/mapStateSlice/index.ts
ericboucher: @echaidemenos could you make this change when you have a moment please? Thanks :)

PR #638, opened 2022-11-18
fix wms issues

DanielJDufour:
prism-app will call createGetMapUrl with a string for a `styles` value.  I had assumed incorrectly that we were passing in an array of style strings.  This was the wrong assumption, so I updated the code to be extra cautious and only call `.join(',')` if styles is an array.


wadhwamatic:
Thanks @DanielJDufour for the work on this. I tested various WMS, WFS and WCS requests without an issue after your latest commit. Over to @ericboucher for review from his side

ericboucher (approved):

src/utils/server-utils.ts
```diff
@@ -504,13 +504,17 @@ export async function fetchWMSLayerAsGeoJSON(options: {
...
-    const wfs = new WFS(lyr.baseUrl);
+    const wfsServerURL = `${lyr.baseUrl}/wfs`;
+
+    const wfs = new WFS(wfsServerURL);
 
     const wfsLayer = await wfs.getLayer(lyr.serverLayerName);
 
     const featureCollection: GeoJSON.FeatureCollection = await wfsLayer.getFeatures(
       {
+        count: Infinity,
```
ericboucher, line 515: Is there a risk in setting this to Infinity? What would be a reasonable finite value otherwise?

DanielJDufour: Hey, good question.  I'll set it to 100,000.  If we run into a situation where we are using a WFS service that has tens of thousands of features (within a given time frame), we might want to consider implementing paging to improve performance and network reliability.

DanielJDufour: done

PR #645, opened 2022-11-23
Update download button for analysis to be able to download geoJson and csv

sifelix:
Linked to https://github.com/WFP-VAM/prism-app/issues/209

To test the feature:
- Generate an analysis
- Once it is generated, check that in the legend you have a multi-options button
- If you click on CSV or GeoJSON you should download the corresponding format
- In the analysis pop up check that the download button has been renamed DOWNLOAD CSV

image


src/components/MapView/Legends/index.tsx
```diff
@@ -78,21 +82,36 @@ function Legends({ classes, layers, extent }: LegendsProps) {
...
+          },
+          analysisResult?.getTitle() ?? 'prism_extract',
+          'application/json',
+        );
+        break;
+      case 'csv':
+        if (
+          analysisResult &&
+          (analysisResult instanceof BaselineLayerResult ||
+            analysisResult instanceof PolygonAnalysisResult)
```
ericboucher, line 107: Since this is used twice, it might be worth extracting it into a variable with a clear name. Maybe `layerAcceptsCSVDownload` or something like that?

src/components/MapView/Legends/index.tsx
```diff
@@ -15,32 +14,37 @@ import {
...
   ExposedPopulationDefinition,
+  LayerType,
   LegendDefinitionItem,
 } from '../../../config/types';
 import {
   analysisResultSelector,
   isAnalysisLayerActiveSelector,
 } from '../../../context/analysisResultStateSlice';
+import { mapSelector } from '../../../context/mapStateSlice/selectors';
+import { Extent } from '../Layers/raster-utils';
```
ericboucher, line 31: I see a few order changes, are we using a different linter? I don't mind too much either way but we should make sure that we use a consistent heuristic to avoid unnecessary line changes

ericboucher:
Looking great @sifelix, I left some small comments inline

src/components/MapView/Legends/index.tsx
sifelix: It has been changed by the last commits

src/components/MapView/Legends/index.tsx
sifelix: I think that this file has been linted for the imports

frontend/src/components/MapView/Legends/index.tsx
```diff
@@ -94,6 +98,22 @@ function Legends({ classes, layers, extent }: LegendsProps) {
       'application/json',
     );
   };
+  const handleAnalysisDownloadCsv = (): void => {
```
ericboucher, line 101: Could simplify this by extracting the definition of an analysisDate? Since all the other params overlap?

frontend/src/components/MapView/Legends/index.tsx
```diff
@@ -145,15 +165,24 @@ function Legends({ classes, layers, extent }: LegendsProps) {
...
+                mainLabel="Download"
+                options={[
+                  {
+                    label: 'GEOJSON',
+                    onClick: handleAnalysisDownloadGeoJson,
+                  },
+                  {
+                    label: 'CSV',
+                    onClick: handleAnalysisDownloadCsv,
+                    disabled: !(
```
ericboucher, line 178: Does this mean that it will be shown but greyed out? Maybe we should hide it altogether instead? @wadhwamatic what do you think?

ericboucher (approved):
Looks great,

@wadhwamatic let's make sure it works as you expected, including the UX. See my comment in the PR.

@sifelix I just left a small refactor comment

frontend/src/components/MapView/Legends/index.tsx
sifelix: Yes it will be greyed out for now

frontend/src/components/MapView/Legends/index.tsx
sifelix: Done :)

frontend/src/components/MapView/Legends/index.tsx
sifelix: Just pushed a commit to hide the option

wadhwamatic:
@sifelix - nice work! I noticed a few things that need to be addressed"
1. The filename for the geojson should be the same as the CSV file, just change the extension
2. The GeoJSON seems to be formatted incorrectly and fails to load in local GIS software (I'm using QGIS). I believe the GeoJSON should be a 'FeatureCollection' with an array of features. You can look at any of the existing boundary files in PRISM as an example
3. The date label in the filename is incorrect. It's taking the layer date and adding one to it. Note, the 'download CSV' button in the analysis results table gets the right date

ajanim:
> 



> @sifelix - nice work! I noticed a few things that need to be addressed:
> 
> 1. The filename for the geojson should be the same as the CSV file, just change the extension
> 2. The GeoJSON seems to be formatted incorrectly and fails to load in local GIS software (I'm using QGIS). I believe the GeoJSON should be a 'FeatureCollection' with an array of features. You can look at any of the existing boundary files in PRISM as an example
> 3. The date label in the filename is incorrect. It's taking the layer date and adding one to it. Note, the 'download CSV' button in the analysis results table gets the right date

1 & 2 are fixed. As we discussed, 3 is a separate bug that will be handled [here](https://github.com/WFP-VAM/prism-app/issues/652)

wadhwamatic (approved):
Looks good, thanks Felix and Ajani

PR #646, opened 2022-11-24
Use adminlevel name instead of admincode for groupby COUNTRY=mozambique

### LINKED ISSUE #631: Exposure analysis and report use incorrect admin attributes in the table view
When I run an exposure analysis and export a report of that analysis, the table that is generated uses the incorrect admin attributes for the name of the area.  

Steps to reproduce:
  2. Open the tropical storm layer
  3. Go to date 2022-03-12
  4. Click the 'exposure analysis' button
  5. Toggle 'table view' on
  6. Select 'export' and then 'report' 

Here's a screenshot of the current values in the table view in the app which are meaningless to users. It's also missing the total row.
Screen Shot 2022-11-08 at 21 47 48

And here's a screenshot of the values seen in the report. Note, the total row is also missing commas in the report
Screen Shot 2022-11-08 at 21 48 32

Please use the same approach that is used in the 'run analysis' output table. The header should say 'Name', and the values in the first column should be the name of the admin area, using commas to separate admin levels. See screenshot below for an example of a run analysis output in Mozambique:
Screen Shot 2022-11-08 at 21 45 45

JorgeMartinezG:
…population analysis

This PR closes #631.

image

**How to test**
- Run an exposed population analysis for mozambique (date: 2022-01-25)
- Click on Table view


ericboucher:
@JorgeMartinezG thanks for looking into this! A few things to think about imho:

I think we could keep the group by as is, but fetch the admin boundary names some other way. Ideally it would work similarly to the raster analysis table that shows coma-separated admin levels. And that gets translated when we change the language.

Finally, it should also work for flood exposure analysis

JorgeMartinezG:
@ericboucher I agree with you. However, I believe this PR fixes an urgent issue. Maybe creating an issue and addressing it separately would be another option. cc @wadhwamatic

ericboucher:
Ok, then can you make the change for flood exposure analysis as well and open an issue for the rest? Thanks :)

wadhwamatic:
> @ericboucher I agree with you. However, I believe this PR fixes an urgent issue. Maybe creating an issue and addressing it separately would be another option. cc @wadhwamatic

A timely fix is important, but let's make sure we're covering flood and storm exposure. I'm also not sure how we can use the title for the admin column header. I.e., the header should say 'District' or 'Admin name' instead of 'adm2_name'.

Agreed with @ericboucher - please create an issue to capture a more complete fix that is inline with how the run analysis table result works.

JorgeMartinezG:
closing in favor of #665

PR #655, opened 2022-11-30
Automated deployments from the master branch

sifelix:
This PR adds a CD for master.
It creates 3 deployments on push to master:
- Cambodia https://prism-master-cambodia.surge.sh/
- Mozambique https://prism-master-mozambique.surge.sh/
- RBD https://prism-master-rbd.surge.sh/

image

Those deployments are not triggered by a pull request:
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,6 +52,7 @@ jobs:
 
   frontend_build_and_deploy:
     needs: frontend_tests
+    if: github.event_name == 'pull_request'
     name: frontend_build_and_deploy
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -113,6 +114,55 @@ jobs:
...
             Build succeeded and deployed at http://${{ env.surge_url }} 
             (hash ${{ github.event.pull_request.head.sha }} deployed at ${{ steps.date.outputs.date }})
+
+  master_frontend_build_and_deploy:
+    needs: frontend_tests
+    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
+    name: master_frontend_build_and_deploy
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
ericboucher, line 124: The inconvenient of a matrix is that it will run all the tasks times 3, including checkout and dependency building. Not a big deal but we could probably clean things up a bit. Could be a todo for someone else later if you don't see how right now.

.github/workflows/build.yml
```diff
@@ -113,6 +114,55 @@ jobs:
...
+        country: [mozambique, cambodia, rbd]
+    env:
+      surge_url: prism-master-${{ matrix.country }}.surge.sh
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Get yarn cache directory path
```
ericboucher, line 132: Is this step needed? Maybe not needed in the other deploy either?

ericboucher (approved):
Looks good overall and gives us the expected output. Though I think we should try to think about ways to avoid duplication and keep the actions clean when we can.

.github/workflows/build.yml
```diff
@@ -113,6 +114,55 @@ jobs:
           message: |
             Build succeeded and deployed at http://${{ env.surge_url }} 
             (hash ${{ github.event.pull_request.head.sha }} deployed at ${{ steps.date.outputs.date }})
+
+  master_frontend_build_and_deploy:
+    needs: frontend_tests
+    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```
ericboucher, line 120: Instead of using this and to keep things clean, you could also create a different file/workflo and use the "on:push:banches:master" as the only target.

.github/workflows/build.yml
sifelix: They are cached so it should be good

.github/workflows/build.yml
sifelix: Its output is used by the following step, it was added 5 months ago. Do you know why?

ericboucher:
We could also look into creating composite actions for our common tasks
See https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

Again, not a priority but could be interesting to explore

.github/workflows/build.yml
```diff
@@ -113,6 +114,55 @@ jobs:
...
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+
+      - uses: actions/cache@v2
+        id: yarn-cache
+        with:
+          path: |
+            **/node_modules
+            **/.eslintcache
+            ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/frontend/yarn.lock') }}-${{ hashFiles('**/packages/common/yarn.lock') }}
```
ericboucher, line 143: packages/common -> common

.github/workflows/build.yml
sifelix: From our discussion: for this short term solution let's keep it here

.github/workflows/build.yml
sifelix: From our discussion: we can benefit from the cache

PR #662, opened 2022-12-02
Integrating Common WCS Functionality COUNTRY=mongolia

DanielJDufour:
- [X] Related issue: https://github.com/WFP-VAM/prism-app/issues/413
- [ ] How to test the feature:
  - [ ] we will need to deploy a test instance, and check if the scheduled alerting works 
  - [ ] run REACT_APP_COUNTRY=mongolia yarn start and check if analysis works for hazard layers (with vector and raster zones)  
- [ ] Screenshot/video of feature: N/A (nothing should appear different)

## Notes on Changes to Docker Configuration in Alerting
I modified the Dockerfile and docker-compose.yml for alerting in order to install the common folder.  The issue is basically that Docker by default doesn't accept links up the directory path (i.e. to parent folders).  That's why I had to change the context to the parent folder.

## Testing Against Regressions
I would recommend testing existing WMS functionality just to make sure none of the changes in this PR led to any regressions.  fwiw, I tested running analysis on http://localhost:3000/?hazardLayerIds=adamts_buffers&date=2019-03-15&analysisHazardLayerId=adamts_buffers&analysisAdminLevel=1 with date 2019-03-15


alerting/src/utils/analysis-utils.ts
```diff
@@ -1,11 +1,14 @@
...
 import { get, isNil } from 'lodash';
 import bbox from '@turf/bbox';
+import { createGetCoverageUrl } from 'prism-common';
 import { Extent } from './raster-utils';
-import { getWCSLayerUrl } from './server-utils';
 import { ANALYSIS_API_URL } from '../constants';
 import { Alert } from '../entities/alerts.entity';
 
+// eslint-disable-next-line fp/no-mutation
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
ericboucher, line 10: Why is this change needed here?

DanielJDufour: Good question.  I'm not sure.  I just copied it over from the server-utils.ts file.  I see that it was originally added with this PR: https://github.com/WFP-VAM/prism-app/pull/149/files#diff-1f860e81328aef6dfd726374f959ffb6ff8fa811ba25c9ca7fba8c665770d365R9

The direct reason is I was trying to preserve any prior functionality.  I think because analysis-utils imported from server-utils, NODE_TLS_REJECT_UNAUTHORIZED was essentially set to '0' inside analysis-utils, too.

The only original reason I can think of is that one of the WCS services had a certificate issue, so making https requests was rejecting and this was added as a workaround.  It might no longer be necessary.

Let me know if you'd like me to remove it.

ericboucher (approved):

PR #669, opened 2022-12-08
Download geotiff for any rasters from wfp

sifelix:
- [ ] Related issue: https://github.com/WFP-VAM/prism-app/issues/618
- [ ] How to test the feature:
  - [ ] Select a raster layer from WFP
  - [ ] In the legend, you should see a button "DOWNLOAD WFP GEOTIFF"
  - [ ] On click, you should download the associated GeoTiff
- [ ] Screenshot/video of feature:
image


ericboucher:
@sifelix great start. 

To take this a step further, consider the fact that we won't have the "catalog" stored in the backend.
So we need to create a function that takes the minimum parameters possible that we can reasonably send from the frontend, to create the expected output.

Looking at your code, the list of parameters sent between the frontend and the backend could look like this:

- stac_url
- layer_name
- layer_config
- bbox
- date

Ideally, we want to get the best resolution possible. Wdyt?

api/README.md
```diff
@@ -79,6 +79,18 @@ Returns all form responses using Kobo API
 ```
 curl -X GET 'http://localhost/kobo/forms?nameField=Test%20MMR&datetimeField=_submission_time&geomField=Location&measureField=The_number&beginDateTime=2021-09-15&endDateTime=2021-09-29'
 ```
+### `/rasterGeotiff` (POST)
```
ericboucher, line 82: Let's make this endpoint snake_case please

ericboucher: As well as the variable names below

api/app/geotiff_from_stac_api.py
```diff
@@ -0,0 +1,84 @@
...
+from uuid import uuid4
+
+import boto3
+from cachetools import TTLCache, cached
+from fastapi import HTTPException
+from odc.geo.xr import write_cog
+from odc.stac import configure_rio, stac_load
+from pystac_client import Client
+
+STAC_URL = "https://api.earthobservation.vam.wfp.org/stac"
```
ericboucher, line 11: How about making this a bit more future-proof and accepting the STAC_URL as a parameter?

Not urgent but this is something that will be needed if we want to download data from other sources.

ericboucher (approved):
@sifelix looks good overall, nice work.

I left a few comments inline but overall I would recommend making this a bit more future-proof by accepting a STAC URL as a parameter and reformating the frontend part a bit so that the stac_url can be set for a layer as this is not specific to WFP necessarily. You could then rename the button to `StacDownloadButton`. Wdyt?

frontend/src/components/MapView/Legends/WfpWmsDownloadButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  return (
+    <Button
+      variant="contained"
+      color="primary"
+      size="small"
+      onClick={() => handleDownload()}
+      fullWidth
+    >
+      {t('Download WFP GeoTiff')}
```
ericboucher, line 38: no need to say "wfp" here

api/docker-compose.deploy.yml
```diff
@@ -49,5 +49,7 @@ services:
       - DATABASE_URL=${DATABASE_URL:?"Provide a URL for the alerts database."}
       - KOBO_PW=${KOBO_PW:?"Add a password to access KOBO data."}
       - KOBO_USERNAME=${KOBO_USERNAME:-kobo_user}
+      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
+      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
```
ericboucher, line 53: If this is a requirement in prod, you can make docker-compose complain if it is not set, see "KOBO_PW" above for an example

ericboucher:
@sifelix I deployed this version for now so that you can test and it seems to be working fine (retunring a signed URL and downloading some data). The resulting tif seems empty, however.

api/app/geotiff_from_stac_api.py
sifelix: The issue is that it is not only a URL but also an authenticated S3 bucket

frontend/src/components/MapView/Legends/WfpWmsDownloadButton.tsx
sifelix: @wadhwamatic you recommended having wfp in the button name, didn't you?

wadhwamatic: @sifelix - no... We talked about differentiating downloads delivered through STAC from non-STAC / WCS approaches, which would include WFP's Data Cube layers. That should be a config process, not something a user would ever see or know

wadhwamatic:
Hi @sifelix. Great work, we've just got a few things to address:
1. Change the button to just say 'download'
2. It looks like the button style isn't consistent with other functions in the legend area. Exposure analysis is an example
3. A user shouldn't see the download option for all WMS layers, only those that are accessible via the STAC API. 

The screenshot below shows a WMS layer that can't be downloaded via STAC. The button shouldn't appear in this scenario. The button style also isn't consistent with 'exposure analysis'

Screen Shot 2022-12-19 at 15 26 10

wadhwamatic (changes requested):
See my last comment in the PR for changes needed

wadhwamatic (commented):
Can we rename this file to something like stacRasterDownloadButton? 
frontend/src/components/MapView/Legends/WfpWmsDownloadButton.tsx

This functionality should work with STAC implementations broadly and not be dependent on WFP's implementation only

sifelix:
> 1. Change the button to just say 'download'
> 2. It looks like the button style isn't consistent with other functions in the legend area. Exposure analysis is an example
> 3. A user shouldn't see the download option for all WMS layers, only those that are accessible via the STAC API.

@wadhwamatic 1. and 3. are fixed.

For 2. I have kept the style of the other download buttons with a full width. Do you want me to update the "exposure analysis"
to full width too or to change all the download buttons?

wadhwamatic:
> For 2. I have kept the style of the other download buttons with a full width. Do you want me to update the "exposure analysis" to full width too or to change all the download buttons?

@sifelix - yes, let's update exposure analysis to use the same style as download. Thank you!

wadhwamatic (approved):

wadhwamatic (approved):
Thanks for the UI improvement!

PR #672, opened 2022-12-09
Center the map directly on the target on not on london

sifelix:
[ ] Related issue:
Discussion with Amit
[ ] How to test the feature:
  - [ ] open the website
  - [ ] Yous should directly see the country you want and not London
[ ] Screenshot/video of feature:

### DIFF
--- a/frontend/src/components/MapView/index.tsx
+++ b/frontend/src/components/MapView/index.tsx
@@ -96,9 +96,15 @@ import { DEFAULT_DATE_FORMAT } from '../../utils/name-utils';
 import { firstBoundaryOnView } from '../../utils/map-utils';
 import DataViewer from '../DataViewer';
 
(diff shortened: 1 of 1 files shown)


wadhwamatic:
Thanks @sifelix! This has been much needed. Glad to see if looks like a simple fix too. I see this PR is still a draft. If ready, let's have @ericboucher review. GTG from my side

wadhwamatic (approved):
Good from my side. Thanks!

ericboucher:
@sifelix The problem with this approach is that the app re-centers anytime you select a new layer. Which is not ideal…

ericboucher (changes requested):
Let's find a way for the app not to re-center each time we change a layer

ericboucher (approved):

Issue #582, opened 2022-10-07
Look into a solution to display multiple admin level data layers simultaneously


wadhwamatic:
For the PRISM deployment to WFP's Regional Bureau in Dakar (RBD), the Cadre Harmonisé (CH) data is admin level data produced three times per year. But the admin areas that are described can change from one data round to the next. In some places, data could be for admin 1, for others admin 2, or a description of a distinct population within an admin 2. For users though, we want to show a single data products that shows all levels of data for a given point in time as though they are all part of a single dataset. 

There are three separate datasets / layers that need to be shown together (in order from top to bottom):
- Layer 1: 'special' admin2 level data - inaccessible areas and displaced populations. These are unique admin2 areas where the full admin area wasn't covered. We'll use a different style to make them distinct. See https://github.com/WFP-VAM/prism-app/issues/578 
- Layer 2: admin2 level data - data covers the entire admin area
- Layer 3: admin1 level data - in some instances, data is not collected and analyzed down to admin 2, but only at admin 1 level

In PR https://github.com/WFP-VAM/prism-app/pull/576, I used the group layer feature to show all three layers together with an example dataset without a date attribute (we'll add dates later). It works(!). Earlier, I thought it didn't as I kept a custom boundary file for admin 1 level data. Now we have just one boundary file displaying. This will likely break again when the 'special' admin 3 level data is brought in.

Here are some data examples:
- ['special' Admin 2 level data](https://github.com/WFP-VAM/prism-app/blob/74259814e7fc65dfb6720d5f2b3096938966374f/public/data/rbd/ch/group/ch_admin2_access_issues.json)
- [Admin 2 level data](https://github.com/WFP-VAM/prism-app/blob/74259814e7fc65dfb6720d5f2b3096938966374f/public/data/rbd/ch/group/ch_admin2.json)
- [Admin 1 level data](https://github.com/WFP-VAM/prism-app/blob/74259814e7fc65dfb6720d5f2b3096938966374f/public/data/rbd/ch/group/ch_admin1.json)

The attached CSV file ([ch_restructure_all_AW.csv](https://github.com/WFP-VAM/prism-app/files/9856096/ch_restructure_all_AW.csv)) also shows how data is received and restructured to prepare for import in the format needed for admin_level_data in PRISM. This is the logic for dispersing this data cross layers:
1. If adm2_5_name has a value, that row is a special admin 2 level layer
[...]

PR #680, opened 2022-12-16
Multiple level admin layer display

ajanim:
This fixes issue #582

How to test the feature:

- [ ] Choose a admin layer 
- [ ] Find a district with no admin level 2 data (with admin level 1 data)
- [ ] Click this district on the map
- [ ] Witness the `Admin Level` in the tooltip (matches that of admin level 1)

Screenshot/video of feature:
Screenshot 2022-12-20 at 5 28 05 PM


wadhwamatic:
Hi @sifelix. Great work, we've just got a few things to address:
1. Change the button to just say 'download'
2. It looks like the button style isn't consistent with other functions in the legend area. Exposure analysis is an example
3. A user shouldn't see the download option for all WMS layers, only those that are accessible via the STAC API. 

The screenshot below shows a WMS layer that can't be downloaded via STAC. The button shouldn't appear in this scenario. The button style also isn't consistent with 'exposure analysis'

Screen Shot 2022-12-19 at 15 26 10

wadhwamatic:
This is looking good, thanks @ajanim! For anyone else testing, a couple of countries where the fallback data comes into play are Mauritania and Cote d'Ivoire. Over to @sifelix @ericboucher for technical review, but this works well from my side.

frontend/src/config/rbd/translation.json
```diff
@@ -102,5 +102,6 @@
   "Normalized Difference Vegetation Index (NDVI) derived from MODIS TERRA/AQUA": "Indice de végétation par différence normalisée (NDVI) dérivé de MODIS TERRA/AQUA",
   "Normalized Difference Vegetation Index (NDVI) anomaly when compared to the long-term average": "Anomalie de l'indice de végétation par différence normalisée (NDVI) par rapport à la moyenne à long terme",
   "Daytime land surface temperature": "Température diurne au sol",
-  "Daytime land surface temperature anomaly when compared to long-term average": "Anomalie de la température diurne au sol par rapport à la moyenne à long terme"
+  "Daytime land surface temperature anomaly when compared to long-term average": "Anomalie de la température diurne au sol par rapport à la moyenne à long terme",
+  "Admin Level": "Niveau Administrateur"
```
ericboucher, line 106: I think it should be "Niveau Administratif" instead

frontend/src/config/types.ts
```diff
@@ -439,6 +439,9 @@ export class AdminLevelDataLayerProps extends CommonLayerProps {
 
   @optional
   boundary?: LayerKey;
+
+  @optional
+  backupAdminLevelDataLayers?: string[];
```
ericboucher, line 444: Can we find a better name for this?

frontend/src/context/layers/admin_level_data.ts
```diff
@@ -59,45 +71,76 @@ export async function getAdminLevelDataLayerData(
...
   const adminBoundaries = adminBoundariesLayer.data;
+  const adminBoundaryFeatureProps = adminBoundaries.features.map(feature =>
+    pick(feature.properties, [
+      adminBoundaryLayer.adminCode,
+      ...adminBoundaryLayer.adminLevelNames,
+      ...adminBoundaryLayer.adminLevelLocalNames,
+    ]),
+  );
+  const layerData: DataRecord[] = compact(
+    adminBoundaryFeatureProps.map(adminBoundaryFeatureProp => {
```
ericboucher, line 82: Let's add some comments to explain the different steps.

Should this be a reduce instead to avoid the `eslint-disable-next-line fp/no-mutation` ?

frontend/src/context/layers/admin_level_data.ts
```diff
@@ -111,13 +154,16 @@ export async function getAdminLevelDataLayerData(
...
+             Some properties are missing the admin code ${adminBoundaryLayer.adminCode}`,
           );
         }
+
         const matchProperties = layerData.find(({ adminKey }) =>
           adminBoundaryCode?.startsWith(adminKey),
         );
+
         if (matchProperties && !isNull(matchProperties.value)) {
+          // console.log({ matchProperties });
```
ericboucher, line 166: 👣

frontend/src/context/layers/admin_level_data.ts
ajanim: since the fallback value is conditionally set, I don't believe the reduce will help avoid the `eslint disable`

ericboucher (approved):

PR #696, opened 2023-01-05
Ux revamp add the chart tab

sifelix:
UX REVAMP

How to test the feature:

- [ ] Check the new chart tab
  - [ ] Check that you can select an admin one or 2 and display multiple charts 
- [ ] Check that you can fold the left panel
- [ ] Check that the the chart tab is hidden if no wms chart layers available

Screenshot/video of feature:

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ For each category, you can define sub categories as "subcategorie_name":
     ]
   },
   "alertFormActive": false,
+  "hidePanel": false,
   "icons": {
     "vulnerability": "icon_vulnerable.png",
     "exposure": "icon_basemap.png",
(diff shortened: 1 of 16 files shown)


wadhwamatic:
I know this is work in progress, but sharing what I found so far @sifelix:

1. Adding a 2nd chart after the 1st chart is loaded causes an overlap (see screenshot below)
2. There's no way to deselect admin 2 (and just use admin 1)
3. Charts are too small when there is more than 1. It looks like sizing is adopted from the current data viewer. Suggesting we separate this so the data viewer isn't affected
4. We should consider adding buttons to apply / cancel changes to the chart configuration. Auto-update isn't clear
5. Map view breaks after you toggle between chart and other views
6. There's no ability to download chart data (see Figma design). This isn't a blocker though. We can address it later if we're out of time
7. There's no ability to see charts based on national data. We can address it later if we're out of time

Screen Shot 2023-01-05 at 12 15 13

PR #697, opened 2023-01-06
Ux revamp/exposure analysis

ajanim:
This addresses UX revamp.

How to test the feature:

- [ ] Apply the tropical storm layer
- [ ] Select `Exposure Analysis` in the "layer download" menu
- [ ] Test action buttons at the bottom

Screenshot/video of feature:

https://user-images.githubusercontent.com/88509701/211606687-e7aa15ec-f4d8-4418-974e-3e5430794c23.mov


ericboucher (changes requested):
The flow doesn't seem super intuitive to me. Instead, could we create new rows in the list of hazard layers in the analysis tab with a `people` logo 🧑‍🤝‍🧑 ?

(Out of scope, we should consider better ways to handle date changes)

ericboucher:
Not sure if from this PR, but increasing pagination size breaks the layout:

Screenshot 2023-01-09 at 4 24 49 PM

common/src/base/test.ts
```diff
@@ -41,15 +41,15 @@ test("Base.getCapabilities with bad versions", async ({ eq }) => {
   let msg1;
   try {
     await base.getCapabilities({ debug: true, version: "incorrect" });
-  } catch (error) {
+  } catch (error: any) {
```
ericboucher, line 44: This change suggests that revamp is not up to date with master fyi

frontend/src/context/sidebarStateSlice.ts
```diff
@@ -0,0 +1,30 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import type { RootState } from './store';
+
+type SidebarState = {
```
ericboucher, line 4: Is `sideBar` the proper name for this? sideBar could be many things.

ericboucher (approved):

frontend/src/context/sidebarStateSlice.ts
sifelix: renamed

PR #707, opened 2023-01-12
Ux revamp/fix polygon analysis style

sifelix:
Ux revamp fix polygon analysis font style
and fix display of the table

How to test the feature:

- [ ] got to the analysis table 
- [ ] select a polygon analysis layer
- [ ] the font should be black for all the selectors
- [ ] Check that the table is well displayed
Screenshot/video of feature:

image

### DIFF
--- a/frontend/src/components/MapView/LeftPanel/AnalysisPanel/index.tsx
+++ b/frontend/src/components/MapView/LeftPanel/AnalysisPanel/index.tsx
@@ -490,7 +490,7 @@ function AnalysisPanel({
                       className={classes.analysisParamTitle}
                       variant="body2"
                     >
-                      Admin Level
+                      {t('Admin Level')}
                     </Typography>
                     <SimpleDropdown
                       value={adminLevel}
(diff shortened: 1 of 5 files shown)


frontend/src/components/MapView/LeftPanel/AnalysisPanel/index.tsx
```diff
@@ -486,21 +486,37 @@ function AnalysisPanel({
               {hazardDataType === GeometryType.Polygon && (
                 <>
                   <div className={classes.analysisPanelParamContainer}>
-                    <Typography variant="body2">Admin Level</Typography>
+                    <Typography
+                      className={classes.analysisParamTitle}
+                      variant="body2"
+                    >
+                      Admin Level
```
ericboucher, line 493: While we're at it can we translate the parts that are not translated yet like this one?

ericboucher (approved):

ericboucher:
Quick UI fix, I think that there is too much spacing at the top of the tables. I think we could bring this all the way to the top almost, similar to what we have in the designs.

Screenshot 2023-01-13 at 11 56 25 AM

PR #718, opened 2023-01-23
Create mapping for data  within popup COUNTRY=cambodia

### LINKED ISSUE #491: Allow for configuration of text labels in pop up when you click on a point or admin level layer
We currently use the name of the layer to label a value when a user clicks on a point or admin level layer. We should instead allow the label in the popup to be configured as its not always referring to the layer, but a specific attribute of it. 

The screenshot below is for a kobo form in Cambodia. Rather than show the text 'Disaster impact report: 55', it should show 'Number of people affected: 55' which is the 'measure_field' for this layer. 

Screen Shot 2022-06-21 at 18 21 02

We should also consider displaying the label associated with values that are configured in layer. This would allow us to display text that is associated with a categorical variable. For example, a Kobo form may have values of 1-10, but they are associated with types of disasters (1=flood, 2=drought, etc.). If we can display the label instead of the value, we'll have more useful info in the popup. The screenshot below shows 'incident report' with a value of 5 in the popup. The layer configuration associates the value of 5 with the label 'lightning'.

Screen Shot 2022-07-05 at 11 18 21

JorgeMartinezG:
This fixes issue. Closes #491 

This PR removes the default dataField that is included in the popup window if dataField appears within feature_info_props. Also, a **mapping** type is included to parse values according to the mapping established .

How to test the feature:
- Select the EWS sensor data using cambodia country deployment.

Screenshot/video of feature:
image


JorgeMartinezG:
> This great, thanks @JorgeMartinezG. I added another example for Cambodia.
> 
> I think a more clear name for this attribute could be 'values_labels' instead of 'mapping'. What do you think?

I've modified the variable. Used Values instead of values_labels since PRISM does not allow variables in snake case

wadhwamatic:
@JorgeMartinezG - it looks like this wasn't implemented for admin_level data as well? I tried to add a label for IDPoor data but both the data_field and label show up in the popup.

Screenshot 2023-01-24 at 10 30 08

wadhwamatic:
Looks great @JorgeMartinezG, thanks! @ericboucher - can you provide a review?

frontend/src/utils/server-utils.ts
```diff
@@ -301,11 +301,23 @@ export async function getLayersAvailableDates(): Promise<AvailableDates> {
...
+export function formatFeatureInfo(
+  value: string,
+  type: LabelType,
+  values?: { [key: string]: string },
+): string {
   if (type === LabelType.Date) {
     return `${moment(value).utc().format('MMMM Do YYYY, h:mm:ss')} UTC`;
   }
 
+  if (type === LabelType.ValuesLabels) {
```
ericboucher, line 313: I believe the correct terminology for this is `label map`.

frontend/src/config/types.ts
```diff
@@ -365,11 +365,13 @@ export enum LabelType {
   Date = 'date',
   Text = 'text',
   Number = 'number',
+  ValuesLabels = 'values_labels',
 }
 
 interface FeatureInfoProps {
   type: LabelType;
   label: string;
```
ericboucher, line 373: I think that the current parameter names and types are confusing. Between `value, values, label, ValuesLabels` it's easy to get lost.

Can we find better names?

frontend/src/config/cambodia/layers.json
```diff
@@ -1448,9 +1448,9 @@
     "date_url": "",
     "feature_info_props": {
       "status": {
-        "type": "values_labels",
-        "label": "Sensor status",
-        "values": {
+        "type": "labelMapping",
```
ericboucher, line 1451: to be consistent let's try to keep things in snake_case in the json files and camelCase in the code.

frontend/src/config/types.ts
```diff
@@ -361,17 +361,17 @@ export class BoundaryLayerProps extends CommonLayerProps {
...
   Text = 'text',
   Number = 'number',
-  ValuesLabels = 'values_labels',
+  LabelMapping = 'labelMapping',
 }
 
 interface FeatureInfoProps {
-  type: LabelType;
+  type: DataType;
   label: string;
```
ericboucher, line 373: This should become dataTitle, no?

frontend/src/config/types.ts
```diff
@@ -361,17 +361,17 @@ export class BoundaryLayerProps extends CommonLayerProps {
...
-export enum LabelType {
+export enum DataType {
   Date = 'date',
   Text = 'text',
   Number = 'number',
-  ValuesLabels = 'values_labels',
+  LabelMapping = 'labelMapping',
 }
 
 interface FeatureInfoProps {
```
ericboucher, line 371: Looks like we have multiple `FeatureInfoProps` interfaces defined. Is this expected?

ericboucher (approved):

frontend/src/config/cambodia/layers.json
```diff
@@ -66,6 +66,12 @@
       "mode": "forward"
     },
     "opacity": 0.9,
+    "feature_info_props": {
+      "Total_Poor_12_HHs": {
+        "type": "number",
+        "label": "Number of ID Poor (1 & 2) households"
```
ericboucher, line 72: dataTitle?

frontend/src/config/cambodia/layers.json
```diff
@@ -102,6 +108,12 @@
     "data_field": "Total_Poor_1_HHs",
     "admin_level": 3,
     "admin_code": "Adm3_text",
+    "feature_info_props": {
+      "Total_Poor_1_HHs": {
+        "type": "number",
+        "label": "Number of ID Poor 1 households"
```
ericboucher, line 114: dataTitle?

frontend/src/config/cambodia/layers.json
```diff
@@ -139,6 +151,12 @@
     "data_field": "Total_Poor_2_HHs",
     "admin_level": 3,
     "admin_code": "Adm3_text",
+    "feature_info_props": {
+      "Total_Poor_2_HHs": {
+        "type": "number",
+        "label": "Number of ID Poor 2 households"
```
ericboucher, line 157: dataTitle?

PR #727, opened 2023-01-30
circle color based on non-numeric value for PointDataLayer type COUNTRY=myanmar

### LINKED ISSUE #713: [Feature Request]: Look into options to handle non-numeric values for admin level and point data layers
### Provide a clear and concise description of what the problem is.

In some cases, data for admin level and point data layers may not be numeric. For example, ACLED provides 'event_type' as a string with values like "Explosions/Remote violence" "Battles", and "Strategic developments". 

We should handle non-numeric value options. Ideally, we can simply configure this in the legend for point and admin level data. For example:

    "legend": [
      { "label": "Explosions/Remote violence", "value": "Explosions/Remote violence", "color": "#ea5545" },
      { "label": "Battles", "value": "Battles ", "color": "#ef9b20" },
      { "label": "Strategic developments", "value": "Strategic developments", "color": "#ede15b" }
    ]

We might want to consider adding a data_type attribute as well which defaults to numeric but can accept 'string' as an option too. There may be a future case for 'date' too.

Here's a screenshot of a configured legend based on numeric text. The point data would use the symbology that was configured. This example is from Myanmar in PR https://github.com/WFP-VAM/prism-app/pull/704 

Screenshot 2023-01-18 at 12 08 09

JorgeMartinezG:
This fixes issue [enter issue number]. Closes #713. 

This Pull request includes a new type for the PointDataLayer type **DataFieldType** which is used to check the type of the data present within the dataset. Then, the mapbox layer will be colored based on the legend and data field type used. The default is set to NUMBER

How to test the feature:
- Using deployment myanmar.
- Select the **Types of armed conflict incident** layer.

Screenshot/video of feature:
image

### DIFF
--- a/frontend/src/components/MapView/Layers/styles.ts
+++ b/frontend/src/components/MapView/Layers/styles.ts
@@ -1,19 +1,40 @@
 import * as MapboxGL from 'mapbox-gl';
-import { CommonLayerProps } from '../../../config/types';
+import {
+  CommonLayerProps,
+  DataFieldType,
+  PointDataLayerProps,
+} from '../../../config/types';
 import { legendToStops } from './layer-utils';
 
 export const circleLayout: MapboxGL.CircleLayout = { visibility: 'visible' };
-export const circlePaint = (
-  { opacity, legend }: CommonLayerProps,
-  property: string = 'data',
-): MapboxGL.CirclePaint => ({
-  'circle-radius': 8,
-  'circle-opacity': opacity || 0.3,
-  'circle-color': {
-    property,
-    stops: legendToStops(legend),
-  },
-});
+
+export const circlePaint = ({
+  opacity,
+  legend,
+  dataField,
+  dataFieldType,
+}: PointDataLayerProps): MapboxGL.CirclePaint => {
+  const circleColor =
+    dataFieldType === DataFieldType.TEXT
+      ? [
+          'match',
+          ['get', dataField],
+          ...legend
+            .map(legendItem => [legendItem.label, legendItem.color])
+            .reduce((acc, item) => [...acc, ...item]),
+          '#CCC',
+        ]
+      : {
+          property: dataField,
+          stops: legendToStops(legend),
+        };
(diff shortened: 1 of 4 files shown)


frontend/src/config/myanmar/layers.json
```diff
@@ -2287,8 +2287,8 @@
       "limit": 0
     },
     "type": "point_data",
-    "data": "https://prism-api.ovio.org/acled",
-    "date_url": "https://prism-api.ovio.org/acled",
+    "data": "http://localhost/acled",
```
ericboucher, line 2290: ⚠️

frontend/src/config/myanmar/layers.json
```diff
@@ -2331,6 +2331,7 @@
     "data": "http://localhost/acled",
```
ericboucher, line 2331: this should point to the API

frontend/src/components/MapView/Layers/styles.ts
```diff
@@ -1,19 +1,40 @@
...
+  dataFieldType,
+}: PointDataLayerProps): MapboxGL.CirclePaint => {
+  const circleColor =
+    dataFieldType === DataFieldType.TEXT
+      ? [
+          'match',
+          ['get', dataField],
+          ...legend
+            .map(legendItem => [legendItem.label, legendItem.color])
+            .reduce((acc, item) => [...acc, ...item]),
```
ericboucher, line 24: Could we combine this in one step?

Could you add a quick comment with the expected format for the final `MapboxGL.Expression`?

ericboucher (approved):
Looks good overall. Please be careful with the API endpoints, config on master should not point to localhost.

wadhwamatic:
@JorgeMartinezG and @ericboucher - The API hadn't been updated yet with the ACLED feature. I deployed that and verified that this PR works as expected. CI Build is failing though so I haven't merged. Any thoughts on why this is failing?

PR #728, opened 2023-01-30
Add time dimension for static raster layer

### LINKED ISSUE #724: [Feature Request]: Add an optional time dimension to static raster layers
### Provide a clear and concise description of what the problem is.

The new layer type 'static_raster' doesn't support timeline navigation yet

I want to be able to add multiple data points for a static raster over time. For example, the static raster might capture major flood events over a rainy season, with dates on 2023-01-10, 2023-01-13, and 2023-01-17. The static raster is created by an analyst and the tileset used by PRISM would be created as needed with a new tileset for each of those 3 dates.

Look into using a similar approach as what was done for dates and admin level data (https://github.com/WFP-VAM/prism-app/pull/555).

The configuration of a static raster with dates could look like this:
```
  "flood_static": {
    "type": "static_raster",
    "base_url": " http://prism-raster-tiles.s3-website-us-east-1.amazonaws.com/khm-flood-events/{YYYY_MM_DD}/{z}/{x}/{y}.png",
    "date_url": "data/cambodia/khm-flood-event-dates.json"
}
```

The base URL would refer to these 3 folders in this example:
http://prism-raster-tiles.s3-website-us-east-1.amazonaws.com/khm-flood-events/2023_01_10/{z}/{x}/{y}.png
http://prism-raster-tiles.s3-website-us-east-1.amazonaws.com/khm-flood-events/2023_01_13/{z}/{x}/{y}.png
http://prism-raster-tiles.s3-website-us-east-1.amazonaws.com/khm-flood-events/2023_01_17/{z}/{x}/{y}.png

The static_raster layer was introduced through this PR https://github.com/WFP-VAM/prism-app/pull/684
Admin level dates were implemented in this PR: https://github.com/WFP-VAM/prism-app/pull/555

JorgeMartinezG:
This fixes issue [enter issue number]. Closes #724 

How to test the feature:
- Using cambodia country deployment select layer **Flood frequency from static raster source (with dates)**

Screenshot/video of feature:
image

**Note.** Looks like the tiles server does not contain the given tiles. Is returning 404


wadhwamatic:
This is a great start, thanks @JorgeMartinezG. I added test data for Cambodia with the same dates that were configured. 

There's now a discrepancy with how the base url is configured for static_raster layers with dates vs. without dates. Without dates we add '{z}/{x}/{y}.png' as part of the base URL. But for 'with dates', in src/components/MapView/Layers/StaticRasterLayer/index.tsx, you've assumed that the configuration will use {z}/{x}/{y}.png. We should avoid this.

I think the easiest and most flexible thing to do would be to remove the {z}/{x}/{y}.png expectation in StaticRasterLayer/index.tsx and have layers.json's config look like:
http://prism-raster-tiles.s3-website-us-east-1.amazonaws.com/khm-flood-events/{YYYY_MM_DD}/{z}/{x}/{y}.png

That should allow us to support different tiling schemes and make it part of the config process. What do you think?

wadhwamatic:
Thanks @JorgeMartinezG. This is great. I removed the Cambodia test layers and updated South Sudan (where this will be first implemented). 

I'm not sure if you intentionally set this to draft. If it's ready, please mark it so @ericboucher can review.

frontend/src/components/MapView/index.tsx
```diff
@@ -200,7 +201,7 @@ function MapView({ classes }: MapViewProps) {
   );
   const selectedLayersWithDateSupport = selectedLayers
     .filter((layer): layer is DateCompatibleLayer => {
-      if (layer.type === 'admin_level_data') {
+      if (layer.type === 'admin_level_data' || layer.type === 'static_raster') {
```
ericboucher, line 204: This works, but maybe we could change the logic a bit to make it more  clear / homogeneous?

I'm thinking we could:
- add `admin_level_data` and `static_raster` to dateSupportLayerTypes
- and then return `if dateSupportLayerTypes.includes(layer.type) -> (layer.serverLayerName in serverAvailableDates || Boolean(layer.dates))

What do you think? If it's too annoying because of type conflicts don't worry about it

JorgeMartinezG: @ericboucher @wadhwamatic  I tried the logic and modified a bit but did not work. Mostly related that serverLayerName is only a field for wms layer type

JorgeMartinezG: the way dates are collected depend on the layer type. Most of the logic per layer is in server-utils.ts file

ericboucher: ok, let's keep it like that for now then and we can revisit this later

PR #729, opened 2023-01-31
UX revamp - extra steps

echaidemenos:
This fixes the "Extra Steps" of issue https://github.com/WFP-VAM/prism-app/issues/698

- [x] 1. In the layers tab, use icons instead of the kebab menu to access: download, analysis, and visibility
- [x] 2. The chart view should scroll independently. Right now, the panel headers (tabs at the top) scroll when there are multiple charts selected
- [x] 3. Charts should always be at the top of the chart viewer. Right now, if there are < 4 charts, there's a large gap at the top
- [x] 4. There's too much spacing at the top of the result tables. I think we could bring tables almost all the way to the top, similar to what we have in the figma designs.
- [x] 5. When you switch between tabs, retain the last state. i.e., if I had some charts selected and then navigate to the layers tab and back to charts, I should see the charts I had selected before. The same applies to the analysis tab
- [x] 6. Make chart data available for download
- [x] 7. Add 'print' feature in top bar. This is the same function as the current export > image function

Sample csv export from charts: [Manica_10_day_rainfall_estimate_mm_1_month_rainfall_aggregate.csv](https://github.com/WFP-VAM/prism-app/files/10579585/Manica_10_day_rainfall_estimate_mm_1_month_rainfall_aggregate.csv)

# After Changes:
Screenshot from 2023-01-31 10-35-07
Screenshot from 2023-01-31 10-35-13
Screenshot from 2023-02-03 13-48-09
Screenshot from 2023-01-31 10-36-58
Screenshot from 2023-02-03 15-04-42
Screenshot from 2023-02-03 15-04-50
Screenshot from 2023-01-31 10-38-13
Screenshot from 2023-02-03 13-39-01
Screenshot from 2023-02-03 13-39-07

# Before changes:
Screenshot from 2023-01-31 16-33-28
Screenshot from 2023-01-31 16-34-20
Screenshot from 2023-01-31 16-34-32
Screenshot from 2023-01-31 16-35-35
Screenshot from 2023-02-03 13-44-54


wadhwamatic:
This is great, thanks @echaidemenos! 
Some feedback on what's there so far (not in the same order listed in the PR):
1. In the layer tab, can we change the 'exposure analysis' icon to either match the analysis icon used in the analysis tab, or use this one: https://mui.com/material-ui/material-icons/?selected=Troubleshoot
2. When there's only one chart loaded, it is too large. Can we reduce it by 50%?
3. When I ran an exposure analysis after I already did a 'run analysis', there's a blank pane to the right of the results (see screenshot below). 
4. Steps to reproduce the issue above: country=mozambique; select analysis tab; select any parameters in the analysis config; run analysis; view analysis results as a table; then load tropical storm layer and run exposure analysis

Everything else looks great so far. I'll keep testing tomorrow. Nice work!
Screen Shot 2023-01-31 at 22 42 33

echaidemenos:
Hello @wadhwamatic and thanks for the review.

Just a comment on 1, the icon linked is only available in Material v5 and not on v4, so i used the same with the analysis tab.

Other than that i think i resolved the rest, let me know what you think.

wadhwamatic:
Hi @echaidemenos. Sorry for the late reply, I was traveling last week. The updates look good, thanks! I made a minor change to have the top margin of the charts panel to be 0px so it's always top aligned. What do you think?

The icon next to 'print' is a bit confusing. Can we use this instead?
https://mui.com/material-ui/material-icons/?query=print&selected=Print

Once the print icon is updated, I think this should be good to go. 

@ericboucher - can you review?

frontend/src/components/MapView/LeftPanel/AnalysisPanel/index.tsx
```diff
@@ -457,6 +522,10 @@ function AnalysisPanel({
     }
   };
 
+  if (tabIndex !== tabValue) {
```
ericboucher, line 525: isn't this duplicative of what's happening above?

frontend/src/components/MapView/LeftPanel/layersPanel/MenuSwitch/SwitchItem/ExposureAnalysisOption.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const { t } = useSafeTranslation();
+  const analysisResult = useSelector(analysisResultSelector);
+  const { startDate: selectedDate } = useSelector(dateRangeSelector);
+
+  // Since the exposure analysis doesn't have predefined table in configurations
+  // and need to have a `TableKey` will use this util function to handle such case
+  // used timestamp to avoid any potential rare name collision
+  const generateUniqueTableKey = (activityName: string) => {
+    return `${activityName}_${Date.now()}`;
+  };
```
ericboucher, line 40: Since this function is duplicated, let's make it a util function to generate a unique key.

frontend/src/components/MapView/LeftPanel/layersPanel/MenuSwitch/SwitchItem/ExposureAnalysisOption.tsx
```diff
@@ -0,0 +1,97 @@
...
+    dispatch(requestAndStoreExposedPopulation(params));
+    dispatch(
+      setCurrentDataDefinition({
+        id: generateUniqueTableKey('exposure_analysis') as TableKey,
+        title: analysisResult?.getTitle(t) || '',
+        table: '',
+        legendText: t(analysisResult?.legendText || ''),
+      }),
+    );
+    // TODO: maybe we could use an enum here instead of 2
```
ericboucher, line 77: @echaidemenos, yes, we should, good call. Can you open an issue for this?

ericboucher (approved):
@echaidemenos nice work. I just left a few comments inline.

frontend/src/components/MapView/LeftPanel/AnalysisPanel/index.tsx
echaidemenos: No. We want this component's state "alive" while we keep jumping pages, in order to have memory of what we ware viewing there. This makes sure that nothing will render while we are not viewing this tab

PR #733, opened 2023-01-31
Add missing name attributes in a few admin boundary files

ericboucher:
The admin2 GeoJSON was missing a name property which was breaking the alerts request to the backend.
This PR adds the missing propery


wadhwamatic (approved):
gtg, thanks @ericboucher !

PR #748, opened 2023-03-09
fix: hot fix for GoTo boundary relations

ericboucher:


wadhwamatic (approved):
Thanks for working on this @ericboucher. Tested in a few instances. This works as expected.

PR #752, opened 2023-03-14
feat: Allow multiple point_data layers to overlap

ericboucher:
Allow multiple point_data static_raster layers to overlap

Screenshot 2023-03-14 at 5 29 07 PM

Steps to reproduce (ping me for authentication creds):
- select `exposure / field reports / incident report`
- select date `2023-01-25`
- select `flood / early warning / EWS 1294 river level data`

-> no data shows

but selecting `2023-01-25` again using the date picker reloads the data, and everything appears as expected.


HarryMytilinaios (approved):

frontend/src/components/App/index.tsx
```diff
@@ -47,31 +49,51 @@ Font.register({
...
+    return userAuth !== undefined;
+  }, [userAuth]);
+
+  // renders the Auth modal only if a user isn't already authenticated
+  const renderedAuthModal = useMemo(() => {
+    if (isUserAuthenticated) {
+      return null;
+    }
+    return <AuthModal />;
+  }, [isUserAuthenticated]);
```
ericboucher, line 65: Could this logic live in the `AuthModal` component instead?

HarryMytilinaios: of course will change it.

frontend/src/components/MapView/Layers/PointDataLayer/index.tsx
```diff
@@ -110,7 +112,7 @@ function PointDataLayer({ layer, before }: LayersProps) {
     updateHistory,
   ]);
 
-  if (!features || map?.getSource(layerId) || !queryDate) {
+  if (!features || !queryDate) {
```
ericboucher, line 115: @HarryMytilinaios this change seems to fix the disappearing layers. But this might also break something else unexpectedly. Let's investigate a bit and see if we need another safeguard.

frontend/src/components/MapView/Layers/PointDataLayer/index.tsx
```diff
@@ -142,7 +137,7 @@ function PointDataLayer({ layer, before }: LayersProps) {
   }
   return (
     <GeoJSONLayer
-      before={before}
+      // before={before}
```
ericboucher, line 140: @wadhwamatic would you like the points to appear above or below the admin boundaries?

frontend/src/context/mapStateSlice/index.ts
```diff
@@ -44,10 +44,28 @@ function keepLayer(layer: LayerType, payload: LayerType) {
...
   return (
     payload.id !== layer.id &&
-    (payload.type !== layer.type || payload.type === 'boundary')
+    (payload.type !== layer.type ||
+      TYPES_ALLOWED_TO_OVERLAP.includes(payload.type))
   );
 }
 
+// Order layers to keep boundaries and point_data on top. point_data first.
+function layerOrdering(a: LayerType, b: LayerType) {
```
ericboucher, line 53: @wadhwamatic this allows us to order layers.
Right now it puts point_data, then admin boundaries, and then it doesn't care. Can you think of other layers we would want to prioritize?

wadhwamatic: After point_data, I think the ideal state would be show whichever layer was most recently selected next.

wadhwamatic:
Thanks for pushing forward @ericboucher! Beyond the scope of this PR, but we'll need to start thinking through better UX for ordering layers. We can make some guesses based on current data, but I can see things breaking down the road.

For right now, based on the data we have, the order might need to be point, raster, admin. Or if possible - point, then most recent selection regardless of type.

ericboucher:
> For right now, based on the data we have, the order might need to be point, raster, admin. Or if possible - point, then most recent selection regardless of type.

Yeah I think that this is how it works with the current fix/ordering

wadhwamatic:
Great progress on this @ericboucher. Quite a few edge cases to think about. I do not think we need to currently support showing multiple kobo form layers simultaneously. Would it be simple to apply logic to allow EWS and one kobo layer at the same time, but not more than one kobo layer?

Here are a few issues I found (screenshots below):
1. When I toggled on EWS, then kobo flood, then kobo drought, several messages appeared and no layers were shown 
2. I'm not able to replicate this, but at some stage I had EWS and two kobo layers active (drought and incidents I think) and then when I deactivated one of the kobo layers, the app crashed
3. When I had flood extent and kobo flood active, then added drought, the boundaries disappeared

I'm hoping just allowing one kobo layer at a time might help to address some of this.

Screenshot 2023-03-19 at 12 48 30

Screenshot 2023-03-19 at 12 47 22

Screenshot 2023-03-19 at 13 00 03

wadhwamatic:
Nearly there! I found two (hopefully) small issues:

1. I'm not seeing flood report data in the popup window when I click on a commune despite data being there. This happens inconsistently (see 1st screenshot)
2. If you toggle a layer on, reduce opacity, turn it off, then back on - the layer will return to the default opacity but the UI will show the reduced capacity that was set earlier (see 2nd screenshot)
3. A horizontal scrollbar appears in the layer navigation when some layers are activated (visible on both screenshots)

Screenshot 2023-03-20 at 09 14 43

Screenshot 2023-03-20 at 09 15 25

Issue #744, opened 2023-03-07
[Bug]: Columns shift in CSV data download of charts when an indicator doesn't have a value for a date


wadhwamatic:
### What happened?

If I create a chart for multiple indicators, in some cases not all indicators will have valid data on a given date. This happens as the availability / production of indicators varies. When this happens, the values in the CSV file are shifted incorrectly.

### Which country / deployment are you running?

PR https://github.com/WFP-VAM/prism-app/pull/688; country=cambodia

### Add a screenshot (if relevant)

<img width="1225" alt="Screenshot 2023-03-07 at 13 31 34" src="https://user-images.githubusercontent.com/3343536/223558192-a02202d8-010e-428f-b84e-cdc91b723575.png">

The NDVI anomaly values are not yet available for 1-March. Instead, values for 3-month rainfall aggregate are shifted over into the NDVI anomaly columns 
<img width="1264" alt="Screenshot 2023-03-07 at 13 37 21" src="https://user-images.githubusercontent.com/3343536/223558661-024c8de2-c411-4661-b69c-bef9fb41b21b.png">

[Kampong Cham_10_day_ndvi_anomaly_modis_10_day_rainfall_estimate_mm_1_month_rainfall_aggregate_10_day_ndvi_modis_3_month_rainfall_aggregate_mm.csv](https://github.com/WFP-VAM/prism-app/files/10914126/Kampong.Cham_10_day_ndvi_anomaly_modis_10_day_rainfall_estimate_mm_1_month_rainfall_aggregate_10_day_ndvi_modis_3_month_rainfall_aggregate_mm.csv)

ericboucher:
Closed by #756

PR #756, opened 2023-03-15
Fix CSV data shift when data are missing

HarryMytilinaios:
This fixes issue [#744].

FIxes the data shift columns in CSV when data are missing.

Also this commit fixes some unique key maps in various components that throws errors on the console.

### DIFF
--- a/frontend/src/components/Common/BoundaryDropdown/index.tsx
+++ b/frontend/src/components/Common/BoundaryDropdown/index.tsx
@@ -92,6 +92,7 @@ const BoundaryDropdown = ({
 
     return relationsToRender.map(item => (
       <MenuItem
+        key={item.level}
         className={setMenuItemStyle(item.level, levels, styles)}
         value={item.bbox.join(',')}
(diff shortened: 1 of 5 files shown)


ericboucher (approved):

ericboucher:
@HarryMytilinaios looks good. Let's add a small test for it maybe?

wadhwamatic (approved):
Works as expected. Thanks @HarryMytilinaios

HarryMytilinaios:
@ericboucher. The tests were failing before in my previous PR because I have done a generalized approch changing the csv-utils file. In this PR I have made the changes component based so only the `dowloadCSV` function inside the `ChartsPanel` component is changed. Should I write snapshot test and component test of the whole component? Or just create a small unit test that includes the `downloadCSV` function?

ericboucher:
@HarryMytilinaios let's create a unitest for the `downloadCSV` function

ericboucher (approved):

Issue #757, opened 2023-03-16
[Bug]: Kobo get dates request doesn't implement filters


wadhwamatic:
### What happened?

The timeline interface shows that there's data available on a given date, but when I change to that date, I get an error message saying there's no data.

I can see that the initial request for available dates adds the filter attribute as expected, but the results do not apply the filter, so dates are being returned when they should not. See this query:
https://prism-api.ovio.org/kobo/dates?formId=aMLBShjyyhYeS5ERuRy8s3&datetimeField=Date_report&measureField=NumPeoAff&koboUrl=https://kobo.humanitarianresponse.info/api/v2/assets/&filters=status=Approved

As an example, '2023-01-16' is returned as a valid date. If I select that date in the app, the request correctly adds the filter parameter and (correctly) returns an empty dataset. Here's the request:
https://prism-api.ovio.org/kobo/forms?beginDateTime=2023-01-16&endDateTime=2023-01-16&formId=aMLBShjyyhYeS5ERuRy8s3&datetimeField=Date_report&measureField=NumPeoAff&koboUrl=https://kobo.humanitarianresponse.info/api/v2/assets/&filters=status=Approved

I can see in [api/app/kobo.py ](https://github.com/WFP-VAM/prism-app/blob/master/api/app/kobo.py)get_form_dates() doesn't make use of the filter, but get_form_responses() does. 

### Which country / deployment are you running?

Cambodia on PR https://github.com/WFP-VAM/prism-app/pull/688 

### Add a screenshot (if relevant)

https://prism-khm-new-ux.surge.sh/?hazardLayerIds=flood_report&date=2023-01-16

<img width="1520" alt="Screenshot 2023-03-15 at 22 45 06" src="https://user-images.githubusercontent.com/3343536/225529229-c47bfbe8-8065-48c0-8bfb-6bf5fc50b145.png">

Response from the date request:
<img width="355" alt="Screenshot 2023-03-15 at 23 06 23" src="https://user-images.githubusercontent.com/3343536/225529617-fa3b1d5f-f08a-421d-9c6d-efb1c53ea0fc.png">

PR #758, opened 2023-03-16
fix: Add filters to the kobo date query

ericboucher:
This fixes issue #757


wadhwamatic (approved):
thanks for addressing this @ericboucher !

Issue #741, opened 2023-03-03
[Bug]: Timeline ticks are incorrect when I'm in Asia


wadhwamatic:
### What happened?

There's a time zone issue on our dates depending on a user's location. I'm currently in Cambodia (UTC +7) and looking at 10-rainfall in PRISM. The date picker correctly shows that data is available on the 1st, 11th and 21st of the month. But the timeline indicator incorrectly shows layer availability on the 2nd, 12th, and 22nd of the month.

### Which country / deployment are you running?

This is an issue for all deployments. The screenshot is for Zimbabwe, deployed here:
https://prism-zwe-new-ux.surge.sh/

### Add a screenshot (if relevant)

<img width="1047" alt="Screenshot 2023-03-03 at 16 00 49" src="https://user-images.githubusercontent.com/3343536/222679072-fc9c6d1a-6ee4-42e0-a285-f74296164245.png">

wadhwamatic:
@sifelix @ericboucher - can you have a look? I'm not sure if this was introduced when we added the multiple timeline feature or if it's a pre-existing issue that I didn't see until now

wadhwamatic:
> I suspect this has something to do with the multiple timeline feature. Here is the corresponding PR https://github.com/WFP-VAM/prism-app/pull/667/files

I agree. Hoping it can be made a priority as it looks very confusing for users.

PR #761, opened 2023-03-18
fix: Timezone bug with timeline item positioning

ericboucher:
This fixes issue #741 

How to test the feature:

- select a layer with dates
- change the timezone in chrome - https://www.browserstack.com/guide/change-time-zone-in-chrome-for-testing
- reload the page and check that the ticks appear properly

Screenshot 2023-03-18 at 7 07 09 PM

### DIFF
--- a/frontend/src/components/MapView/DateSelector/utils.ts
+++ b/frontend/src/components/MapView/DateSelector/utils.ts
@@ -151,5 +151,5 @@ export function findDateIndex(
 }
 
 export function formatDate(date: number): string {
-  return moment(date + USER_DATE_OFFSET).format(MONTH_FIRST_DATE_FORMAT);
+  return moment(date).format(MONTH_FIRST_DATE_FORMAT);
 }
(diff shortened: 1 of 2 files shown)


wadhwamatic (approved):
This is great @ericboucher, works as expected. Thanks for the tip for testing timezones too. Please apply the fix to the new UX branch too when you merge.

frontend/src/components/MapView/DateSelector/utils.ts
```diff
@@ -151,5 +151,5 @@ export function findDateIndex(
 }
 
 export function formatDate(date: number): string {
-  return moment(date + USER_DATE_OFFSET).format(MONTH_FIRST_DATE_FORMAT);
+  return moment(date).format(MONTH_FIRST_DATE_FORMAT);
```
ericboucher, line 154: @HarryMytilinaios could you do a bit more QA?
This is the only thing that actually changed in this PR. Other changes are just cleanups.

HarryMytilinaios: I think this is good to go. I just did a little bit more QA on this and it seems that it works as expected. I added multiple layers and changed my timezone and it seems pretty solid from my side.

PR #765, opened 2023-03-20
Add national level charts; COUNTRY=cambodia

ericboucher:
This PR adds national level charts provided that `countryAdmin0Id` is set in prism.json. (Actual configuration could evolve if we have the admin0 code available by other means)

This PR also improves the csv name generation adding the country name and all admin levels selected.

### DIFF
--- a/frontend/src/components/DataViewer/index.tsx
+++ b/frontend/src/components/DataViewer/index.tsx
@@ -27,6 +27,7 @@ import { dateRangeSelector } from '../../context/mapStateSlice/selectors';
 import Chart from '../DataDrawer/Chart';
 import { ChartConfig } from '../../config/types';
 import { useSafeTranslation } from '../../i18n';
+import { appConfig } from '../../config';
 
 const isAdminBoundary = (
   params: AdminBoundaryParams | EWSParams,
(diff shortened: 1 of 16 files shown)


wadhwamatic:
This is great, thanks @ericboucher. The UI isn't intuitive though. I don't know as a user that I can skip admin selection to see national level data. 

Can we add a form field above admin 1 which says 'National' or 'Cambodia' that has no other option? Something like this?
Screenshot 2023-03-20 at 17 40 59

ericboucher:
I thought about it a bit and ended up making the selection more explicit and adding info in the menu dropdown. Let me know what you think.

Screenshot 2023-03-21 at 10 36 10 AM

wadhwamatic:
> I thought about it a bit and ended up making the selection more explicit and adding info in the menu dropdown. Let me know what you think.
> 

This works well, thanks @ericboucher. Have you seen this issue? https://github.com/WFP-VAM/prism-app/issues/764. Would it make sense to include a fix in this PR? The team in Cambodia also mentioned an issue with the Go To feature missing some provinces. I'm trying to get more details.

wadhwamatic (approved):

PR #766, opened 2023-03-20
config sri lanka with new ux incl charts; COUNTRY=srilanka

wadhwamatic:
This updates the Sri Lanka config to temporarily deal with map positioning in the new UX, and also adds charts

### DIFF
--- a/frontend/src/config/srilanka/layers.json
+++ b/frontend/src/config/srilanka/layers.json
@@ -1,7 +1,7 @@
 {
   "admin_boundaries": {
     "type": "boundary",
-    "path": "data/srilanka/lka_admin3_w_dataviz.json",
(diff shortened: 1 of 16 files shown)


frontend/src/config/srilanka/layers.json
```diff
@@ -1,7 +1,7 @@
 {
   "admin_boundaries": {
     "type": "boundary",
-    "path": "data/srilanka/lka_bnd_adm3.json",
```
ericboucher, line 4: should we replace the old file completely?

wadhwamatic: yes, done.

ericboucher (approved):

PR #767, opened 2023-03-21
modify tropical storm layer ; COUNTRY=mozambique

wadhwamatic:
This temporarily addresses tropical storm layers not available through Geoserver. It's a temporary fix


wadhwamatic:
resolved through another temporary fix https://github.com/WFP-VAM/prism-app/pull/820

PR #768, opened 2023-03-21
Shift initial map position; COUNTRY=zimbabwe

wadhwamatic:
Fix map positioning for Zimbabwe and new UX.


ericboucher (approved):

ericboucher:
Maybe you can keep the other value in the config as `latitudeNoPanel` or something like that?

wadhwamatic:
> Maybe you can keep the other value in the config as `latitudeNoPanel` or something like that?

good idea. done.

Issue #764, opened 2023-03-20
[Bug]: Chart view in new UX incorrectly uses admin 3 names in the admin 2 list


wadhwamatic:
### What happened?

In countries with admin level 3 boundaries, the chart UI uses level 3 names instead of level 2 names in the drop down list

### Which country / deployment are you running?

Cambodia; PR https://github.com/WFP-VAM/prism-app/pull/688

### Add a screenshot (if relevant)

In this example, I selected Banteay Meanchey as the admin 1, but the list of admin 2 names is actually a list of admin 3s. Boeng Beng should not appear here.

https://user-images.githubusercontent.com/3343536/226437951-0b620916-b959-4378-94f2-39dbed4c4511.mov

PR #770, opened 2023-03-21
Limit chart boundaries to admin2; COUNTRY=cambodia

ericboucher:
This fixes issue #764 

Note that dataviz_id info are needed on the admin2 level boundary file.

Screenshot 2023-03-21 at 6 32 02 PM


wadhwamatic (approved):
works well, thanks @ericboucher !

PR #771, opened 2023-03-21
Concatenate admin names in Exposure Analysis tables

HarryMytilinaios:
Tentative fixes to exposure analysis table data. 

 - Exposure Analysis Sortable Columns
 - Exposure Analysis Table Admin Names Concatenation
 
 Needs a bit more work on fixing the Pdf creation. It is still a work in progress

### DIFF
--- a/frontend/src/components/Common/MultiOptionsButton/index.tsx
+++ b/frontend/src/components/Common/MultiOptionsButton/index.tsx
@@ -89,6 +89,7 @@ function MultiOptionsButton({ mainLabel, options }: IProps) {
       >
         {options.map(option => (
           <StyledMenuItem
+            key={option.label}
             disabled={option.disabled}
             onClick={() => handleOptionClick(option.onClick)}
           >
--- a/frontend/src/components/Common/ReportDialog/ReportDocLegend.tsx
+++ b/frontend/src/components/Common/ReportDialog/ReportDocLegend.tsx
@@ -1,0 +1,70 @@
+import React, { memo, useMemo } from 'react';
+import { Theme } from '@material-ui/core';
+import { StyleSheet, Text, View } from '@react-pdf/renderer';
+import { PDFLegendDefinition } from './types';
+
+interface LegendProps {
+  theme: Theme;
+  title: string;
+  definition: PDFLegendDefinition[];
+}
+
+const makeStyles = (theme: Theme) =>
+  StyleSheet.create({
+    legend: {
+      height: 130,
+      display: 'flex',
+      flexDirection: 'column',
+      marginRight: 10,
+      padding: 10,
+    },
+    legendTittle: {
+      fontSize: theme.pdf?.fontSizes.large,
+      paddingBottom: 10,
+      fontWeight: 500,
+    },
+    legendContentsWrapper: {
+      display: 'flex',
+      flexDirection: 'column',
+      flexWrap: 'wrap',
+      height: '100%',
+    },
+    legendContent: {
+      display: 'flex',
+      flexDirection: 'row',
+      alignItems: 'center',
+      fontSize: theme.pdf?.fontSizes.large,
+      paddingTop: 5,
+      paddingRight: 5,
+    },
+    legendText: {
+      paddingLeft: 10,
+    },
+  });
+
+const ReportDocLegend = memo(({ theme, title, definition }: LegendProps) => {
+  const styles = makeStyles(theme);
+
+  // The rendered definitions
+  const renderedDefinitions = useMemo(() => {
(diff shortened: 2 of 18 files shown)


HarryMytilinaios:
**Additional Items**

- Exposure Analysis Table columns are sortable
- Exposure Analysis Table names are clickable and show map tooltip (Like every analysis table)
- Recreated PDF Report based on the new Table data 
- Component splitting for every different component in the report doc.

**Additional Todo**

Unfortunately `react-pdf/renderer` does not give us a clear loding state when the `PDFViewer` component is rendering, but only the `DowloadPDFLink` gives us this behavior which renders the pdf for dowload independently from the `PDFViewer` component. A possible fix would be to wait some seconds maybe and show a circular progress bar? @ericboucher

wadhwamatic:
Great work on this @HarryMytilinaios. Three comments from my side:
1. The Download CSV button at the bottom of the Analysis tab has disappeared
2. Can we set a default sort for the table either using sum (descending) or name (ascending)? The current default sort seems random
3. Can the report use the same table sorting the user has enabled at the time that they click 'Generate Report'? It currently uses the default. If I change the sort order and then generate report, I'd expect to see the same order in the report

HarryMytilinaios:
@ericboucher @wadhwamatic some additional notes: 

- CSV functionality is now introduced with the new datasets of the exposure analysis. (We had a problem in the CSV transformation because the admin names are concatenated with commas). I took the initiative to replace commas with `/`. Please let me know if this is the best decision in order to make the names in one column below the `Name` column.
- Currently I've set the default sorting as `Ascending order` alphabetically from the `Name` column. 
- The Report is wired to look the `sortBy` key from the exposure analysis table.

Please Nore that when someone changes the `sortBy` column or order the document needs to be recreated to reflect the changes.

ericboucher:
@HarryMytilinaios nice work. For the name you can keep them comma separated but escape them as a string so that the CSV can read it properly. We already do that for the other tables. You probably just need to update the CSV creation function slightly.

wadhwamatic:
Great work @HarryMytilinaios, thanks! Eric's suggestion for dealing with comma separators in the CSV is good. We could also consider having separate columns for each name attribute, i.e. the column headers would be admin1, admin2, admin3, sum. Whatever is simplest to implement will work for now.

HarryMytilinaios:
I escape the admin comma separated names in order to be displayed below the name column. Please let me know if this our expected behavior.

wadhwamatic:
Looks good @HarryMytilinaios, thanks! I see one minor issue which we can address separately. If I run an analysis (not an exposure analysis), the download CSV function doesn't apply the current sort order. When we have time, it would be good to have the same behavior as exposure analysis applied to general analysis.

HarryMytilinaios:
Yes @wadhwamatic you are correct. I didn't know that we wanted to have the exact same functionality in all the analysis tables, so my changes was introduced only in the exposure analysis ones, but the same logic will be applied there too.I can do the changes for the other analysis tables (sortable column by default, sorting in general applied to PDF) in this PR. Cc @ericboucher

ericboucher:
@HarryMytilinaios let's do this change for the other tables while we're at it then, thanks :)

HarryMytilinaios:
Quick update @ericboucher @wadhwamatic . The sorting functionality is now implemented in every analysis table. In the other analysis tables the integration  was a little bit more difficult because we needed to pass the `sortByKey` and `sortOrder` through redux store in order for the legend `downloadCsv` button to have access to the sorting functionality from table. Right now the sorting functionality is reflected to both `downloadCSV` in `analysisActions` section in the sidebar and `downloadCSV` in legend button also

wadhwamatic:
Excellent work @HarryMytilinaios, thanks!! Looks good from my side. 

I'm not able to test a vector-data hazard layer at the moment as the one source we have (WFP GeoServer) is unavailable at the moment. I assume there won't be any issues with those outputs. As an FYI, when we have vector data, there can be multiple columns of data sent back. @ericboucher - do you anticipate any issues once we have vector hazard layers available again?

ericboucher:
I don't anticipate any issues but we can have a look. @HarryMytilinaios can you think about the edge case mentioned by Amit and see if it could break anything? Thanks :)

ericboucher (approved):

HarryMytilinaios:
Everything will work as expected if we have multiple columns. Nothing will be rendered different. We should keep in mind though that the file will not be correct if we have null or undefined data, in which in the analysis tables does not make very much sense, in the charts section it did, because we had multiple charts with multiple different dates and data applied. I think we will be fine with what we have here.

PR #776, opened 2023-03-24
Added "Clear All" Button for clearing charts selections

HarryMytilinaios:
Created "Clear All" button for clearing the selection UX flow in Charts Panel.

@ericboucher Please let me know if we need to remove also the `adminProperties` state variable when we clear the selection flow. As I tested the functionality seems to be working ok, without removing it from the state.

Screenshot 2023-03-24 at 6 58 36 PM


wadhwamatic (approved):
Yes, this does exactly what we need. Thanks @HarryMytilinaios

ericboucher (approved):

PR #777, opened 2023-03-27
Fit map to country bounds considering screen size and side panel

ericboucher:
With this PR the prism.map config accepts a `boundingBox` option that can be used to fit the map properly, whether the panel is open or hidden. This also adapts to the user's screen size. To test, you can change the size of your window and refresh.

@wadhwamatic bounding boxes can be found here - https://github.com/sandstrom/country-bounding-boxes/blob/master/bounding-boxes.json

We could import this file and match to countries programmatically, or copy/paste them in each config file to have a bit more control. Up to you.

### DIFF
--- a/frontend/src/components/MapView/index.tsx
+++ b/frontend/src/components/MapView/index.tsx
@@ -100,6 +100,7 @@ import DataViewer from '../DataViewer';
 import LeftPanel from './LeftPanel';
 import FoldButton from './FoldButton';
 
+// Bounding boxes are adapted from https://github.com/sandstrom/country-bounding-boxes
 const {
   map: { latitude, longitude, zoom, maxBounds, minZoom, maxZoom, boundingBox },
 } = appConfig;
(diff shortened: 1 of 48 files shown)


frontend/src/components/MapView/index.tsx
```diff
@@ -227,6 +227,8 @@ function MapView({ classes }: MapViewProps) {
     | LayerData<BoundaryLayerProps>
     | undefined;
 
+  // TODO - could we simply use the country boundary extent here instead of the calculation?
```
ericboucher, line 230: @wadhwamatic this extent is used for the polygon analysis to limit the data to fetch. Currently we are calculating as the bbox of the boundary file geojson. Could we simply use the country bounds we are defining as part of this PR? Or can you think of edge cases where the admin boundaries span beyond what we would pass as the initial bounding box?

wadhwamatic: I can't imagine a scenario where using the already defined country boundaries wouldn't work. Even for upcoming regional and global versions of PRISM where we have multiple countries, this should still work from what I understand.

HarryMytilinaios (approved):

PR #779, opened 2023-03-28
Add report doc loader

HarryMytilinaios:
Added custom workaround to show loader in creating pdf reports of exposure analysis (currently waiting about 20 seconds for the report to be generated)

How to test the feature:
Select exposure analysis and click the create report button on the exposure analysis actions below in the sidebar.

### DIFF
--- a/frontend/src/components/Common/ReportDialog/index.tsx
+++ b/frontend/src/components/Common/ReportDialog/index.tsx
@@ -1,6 +1,8 @@
-import React, { memo, useCallback, useEffect, useMemo } from 'react';
+import React, { memo, useCallback, useEffect, useMemo, useState } from 'react';
 import {
+  Box,
   Button,
+  CircularProgress,
   createStyles,
   Dialog,
   DialogActions,
@@ -8,6 +10,7 @@ import {
   DialogTitle,
   IconButton,
   Theme,
+  Typography,
   useTheme,
   WithStyles,
   withStyles,
@@ -40,7 +43,8 @@ const ReportDialog = memo(
   }: ReportProps) => {
     const theme = useTheme();
     const { t } = useSafeTranslation();
-    const [mapImage, setMapImage] = React.useState<string | null>(null);
+    const [mapImage, setMapImage] = useState<string | null>(null);
+    const [documentIsLoading, setDocumentIsLoading] = useState(true);
     const selectedMap = useSelector(mapSelector);
     const analysisResult = useSelector(
       analysisResultSelector,
@@ -69,6 +73,14 @@ const ReportDialog = memo(
       },
       [selectedMap],
     );
+
+    // Manual loader wait to show that the document is loading
+    useEffect(() => {
+      const loadingTimer = setTimeout(() => {
+        setDocumentIsLoading(false);
+      }, 20000);
+      return () => clearTimeout(loadingTimer);
+    }, []);
 
     useEffect(() => {
       if (!open) {
@@ -107,6 +119,29 @@ const ReportDialog = memo(
       theme,
     ]);
 
+    const renderedPdfLoadingProgressBar = useMemo(() => {
+      if (!documentIsLoading) {
+        return null;
+      }
+      return (
+        <Box className={classes.documentLoadingContainer}>
+          <CircularProgress size={100} />
+          <Typography
+            className={classes.documentLoadingText}
+            variant="body1"
+            component="span"
+          >
+            {t('Loading document...')}
+          </Typography>
+        </Box>
+      );
+    }, [
+      classes.documentLoadingContainer,
+      classes.documentLoadingText,
+      documentIsLoading,
+      t,
+    ]);
+
     // The report type text
     const reportTypeText = useMemo(() => {
       return reportType === ReportType.Storm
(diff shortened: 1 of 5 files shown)


frontend/src/components/Common/ReportDialog/index.tsx
```diff
@@ -150,7 +192,11 @@ const ReportDialog = memo(
           </span>
           <Button className={classes.actionButton} variant="outlined">
             <PDFDownloadLink document={reportDoc} fileName={getPDFName}>
-              {({ loading }) => (loading ? 'Loading document...' : 'Download')}
+              {({ loading }) =>
+                loading || documentIsLoading
+                  ? 'Loading document...'
```
ericboucher, line 197: These texts should get translated if possible

frontend/src/components/Common/ReportDialog/index.tsx
```diff
@@ -107,6 +119,29 @@ const ReportDialog = memo(
...
+      }
+      return (
+        <Box className={classes.documentLoadingContainer}>
+          <CircularProgress size={100} />
+          <Typography
+            className={classes.documentLoadingText}
+            variant="body1"
+            component="span"
+          >
+            {t('Loading document...')}
```
ericboucher, line 134: Let's remove the `...` from the trabslation function

frontend/src/components/MapView/LeftPanel/AnalysisPanel/index.tsx
```diff
@@ -670,9 +670,21 @@ function AnalysisPanel({
     return (
       <Box className={classes.exposureAnalysisLoadingContainer}>
         <CircularProgress size={100} />
+        <Typography
+          className={classes.exposureAnalysisLoadingText}
+          variant="body1"
+          component="span"
+        >
+          {t('Loading Exposure Analysis Data...')}
```
ericboucher, line 678: let's move the `...` after the translation

wadhwamatic:
Thanks @HarryMytilinaios this is very needed. The loader itself was 'choppy' for me for the first several seconds. Are you seeing this effect?

https://user-images.githubusercontent.com/3343536/228328676-ace48499-05e4-48e6-9511-ef7b39df9f80.mov

ericboucher:
Yes @wadhwamatic we're seeing the choppiness as well. That's because the report generation is taking all the compute power. We're looking into having a simple blinking 3 dots instead to make it easier on the CPU.

HarryMytilinaios:
Added only the `Loading Document...` as a loading text. The problem however is still there. The dots loading and the overall app is laggy when the document is loading for some seconds. This is caused by `@react-pdf/renderer` component library. The problem here is that the document is really large and it is created in the frontend and especially in javascript's main thread. I am not sure what is the correct approach on this. One possible workaround would be to render only some pages (e.g. The first 5 for example), and then prompt the user to download the whole file, I think that would prevent the lagginess at some extent.

frontend/src/components/Common/ReportDialog/index.tsx
HarryMytilinaios: Updated

frontend/src/components/Common/ReportDialog/index.tsx
HarryMytilinaios: Updated

wadhwamatic:
@HarryMytilinaios and @ericboucher. The text approach definitely works better. I think this is an extreme case with the number of rows of data we have being rendered. In the future, we could also limit / filter out unimportant data. But this works for now - thanks a lot.

A server-side approach sounds like the way to go given we know we'll have users who won't have high-end devices.

frontend/src/components/Common/ReportDialog/index.tsx
```diff
@@ -107,6 +118,32 @@ const ReportDialog = memo(
...
+          <Typography
+            className={classes.documentLoaderText}
+            variant="body1"
+            component="span"
+          >
+            {t('Loading document')}
+          </Typography>
+          <span className={classes.documentLoaderDot}>.</span>
+          <span className={classes.documentLoaderDot}>.</span>
+          <span className={classes.documentLoaderDot}>.</span>
```
ericboucher, line 136: Let's wrap this in a small reusable component with a self-explanatory name please?

frontend/src/components/Common/ReportDialog/reportDoc.tsx
```diff
@@ -89,7 +89,7 @@ const ReportDoc = memo(
     columns,
   }: ReportDocProps) => {
     if (mapImage === null) {
-      return <></>;
+      return <Document />;
```
ericboucher, line 92: I think this was causing an error before. See https://github.com/WFP-VAM/prism-app/pull/624
Ask @echaidemenos for more info if needed

ericboucher (approved):

frontend/src/components/Common/ReportDialog/index.tsx
ericboucher: I meant move the blinkingDots into their own component ;) @HarryMytilinaios

frontend/src/components/Common/ReportDialog/reportDoc.tsx
HarryMytilinaios: Yes you are right @ericboucher. This was causing an error but adding the `<></>` component was causing other console errors. I've made a fix that removes the if statement at the root of the report doc component and prevents the auto downloading of the pdf when mounted and not opened. I'll make some tests in different browsers and windows firefox that we had this issue before to see that everything is going as expected

frontend/src/components/Common/ReportDialog/index.tsx
HarryMytilinaios: Updated

ericboucher: Am I missing a commit maybe?

HarryMytilinaios: Ohhh I understood to move the whole component... Sorry for the misunderstanding chaning it

PR #782, opened 2023-03-31
Fix/745 remove fallback data columns; COUNTRY=cambodia

HarryMytilinaios:
This commit removes the normalized `fallback` data from the charts CSV creation. It removes the column headers and the values of `avg` `100` of the anomaly columns.

@ericboucher this is a fix that solves our problem I think. Some additional notes are that the downloadCSV function that downloads the chart data into a CSV file has become quite unreadable. I think we need to address this in a backlog task on moving the function in a separate util file and splitting it in some smaller unit functions and test them properly. Added also individual test in order to have clear testable scenario.

Cc @wadhwamatic


wadhwamatic:
thanks @HarryMytilinaios. This functions as we need. 

From what I understand, we're deleting the column for anomaly_avg, which I think is fine. I don't anticipate anyone using that name for a chart variable but it's not entirely impossible. It would good make sure this is well documented. I don't think we have much info on the charts functionality in the readme yet. We should mention that it's a reserved name for exclusion.

frontend/src/components/MapView/LeftPanel/ChartsPanel/index.tsx
```diff
@@ -197,7 +197,12 @@ export const downloadCsv = (
         const first = value[0];
         const keys = Object.keys(first);
         const filtered = keys.filter(x => x !== dateColumn);
-        const mapped = filtered.map(x => getKeyName(x, key));
+        // created the column names and filters the anomaly_avg column names from the columns collection
+        const mapped = filtered
+          .map(x => getKeyName(x, key))
+          .filter(keyName => {
+            return !keyName.includes('anomaly') || !keyName.includes('_avg');
```
ericboucher, line 204: why test for this `anomaly` or `_avg`, vs `anomaly_avg`

HarryMytilinaios: You are right. Missed it, we could safely change that to:
```suggestion
            return !keyName.includes('anomaly_avg');
```

HarryMytilinaios: We check separately here for `anomaly` and `_avg`. Because we have a special occasion when we enable `NDVI anomaly (MODIS)` which is not following the pattern `anomaly_avg`

ericboucher (approved):

ericboucher:
@wadhwamatic could you test this latest version and let us know if this suits your needs?

wadhwamatic:
> @wadhwamatic could you test this latest version and let us know if this suits your needs?

@ericboucher and @HarryMytilinaios - looks good, thanks!

wadhwamatic (approved):

PR #783, opened 2023-03-31
Fix/759 missing translations; COUNTRY=cambodia

HarryMytilinaios:


HarryMytilinaios:
I've attached some additionals screenshots from the translations
Screenshot from 2023-03-15 12-03-50
Screenshot from 2023-03-15 12-05-17
Screenshot from 2023-04-04 18-20-43
Screenshot from 2023-04-04 18-20-58
Screenshot from 2023-04-04 18-21-03
Screenshot from 2023-04-04 18-21-12
Screenshot from 2023-04-04 18-21-32
Screenshot from 2023-04-04 18-21-57
Screenshot from 2023-04-04 18-22-10

ericboucher:
@wadhwamatic note that Harry turned this back to "draft" because we are investigating a bug with the chart admin dropdown. The way it is implemented right now it breaks when changing the language while admin levels are selected. It is actually quite tricky. One workaround for now if it's ok with you would be to reset the component when the user changes the language on this page. Would that be reasonable? We can also make it work but it will take more time as the changes would touch on multiple components.

Other than that this PR is ready for review and for you to play around / add more translations to the json files.

wadhwamatic:
Thanks both. I'll test later today but, yes - resetting the component when
a user changes language is totally fine.

wadhwamatic:
Thanks both. @ericboucher - yes, resetting the component when a user changes language is totally fine.

wadhwamatic:
I temporarily added a few missing translations and see that this is almost complete. The only things missing I see in the interface now are the analysis table results, the legend title for the results, and the legend's text description for the results. 

In the screenshot below, you can see that '10-day rainfall estimate (mm)' is correctly translated in the list of layers but not in the results table header. For the legend title, we're concatenating the layer title with the statistic (mean, median, min, max) in parentheses. I tried to add those values in translation.json with parentheses but it didn't work. In the legend's layer description, I tried to add 'above threshold' but it also didn't work.

Note, for translations in general, there's a mixture of case-sensitivity after the first word that's inconsistent. We don't have to solve this now, but it makes translations tricky.

Screenshot 2023-04-05 at 14 59 05

HarryMytilinaios:
Thank you @wadhwamatic. I'll take a look at it to resolve it.

HarryMytilinaios:
@wadhwamatic. I added the translations of the analysis table results, as well as in the legends that are created by analysis results. The problem was that the parenthesis were put by the frontend so you don't need to worry about the parenthesis in the `statistic` (e.g median, mean e.t.c). Also we had a typo in threshold. That was the reason why the translation was never took place. Instead of `Threshold` we had `Treshold`. Overall I agree we need a more solid approach on the translations as an improvement we could do something like `TABLE_TITLE` for example which will take place for all the different languages, for the fixed texts and we could work something for the dynamic ones. 

Cc @ericboucher

wadhwamatic:
> Overall I agree we need a more solid approach on the translations as an improvement we could do something like `TABLE_TITLE` for example which will take place for all the different languages, for the fixed texts and we could work something for the dynamic ones.
> 


Thanks for catching that typo. This looks good to me now. I'm waiting for some final translations from the team in Cambodia but everything looks good with the placeholders we have for now.

Agreed on the revised approach in the future. Thanks!

ericboucher:
@wadhwamatic should we merge this or do you prefer to wait for the missing Cambodian translations?

wadhwamatic:
> @wadhwamatic should we merge this or do you prefer to wait for the missing Cambodian translations?

@ericboucher - I added the missing translations. I found two more that I need to get from colleagues in Cambodia but we can go ahead and merge this and I'll make another PR later. Thanks!

wadhwamatic:
> @wadhwamatic should we merge this or do you prefer to wait for the missing Cambodian translations?

@ericboucher - I added the missing translations. I found two more that I need to get from colleagues in Cambodia but we can go ahead and merge this and I'll make another PR later. Thanks!

PR #795, opened 2023-04-07
Added maxValue and minValue in charts configuration

HarryMytilinaios:
This branch takes into account the following assumptions:
 - The `maxValue` `minValue` are added in the fields section of `layers.json` of `chart_data` collection.
 - The codebase calculates the `max` and `min` values of multiple `maxValue` and `minValue` fields, if by mistake for example we configure `maxValue` of 200 in one field and `maxValue` of 300 in another field he codebase will keep the 300, the vise versa applies also in the min values.
 - The `maxValues` and `minValues` are applied in the y-axis ticks as `suggestedMin` `suggestedMax`, this practically means that if a value is over the configured `maxValue`, or less than the configured `minValue` the ticks will be adjusted to facilitate those values.
 
 @wadhwamatic I have added some `minValue` and `maxValue` and `pointRadius` in 10 day rainfall anomaly of the `layers.json` for testing purposes, feel free to adjust and replace them with the acceptable values (or remove them, if we don't need them for the specific `chart_data`).  Following with a commit that accomplishes the `pointRadius` configuration. This was a little bit more tricky, because while `minValues` and `maxValues` are configured in the y-axis, `pointRadius` applies to each data specifically as config. 
 
 Cc @ericboucher

### DIFF
--- a/frontend/src/components/DataDrawer/Chart/index.tsx
+++ b/frontend/src/components/DataDrawer/Chart/index.tsx
@@ -1,11 +1,10 @@
-import React from 'react';
+import React, { memo, useCallback, useMemo } from 'react';
 import colormap from 'colormap';
 import { ChartOptions } from 'chart.js';
 import { Bar, Line } from 'react-chartjs-2';
-import { TFunction, TFunctionKeys } from 'i18next';
+import { TFunctionKeys } from 'i18next';
 import moment, { LocaleSpecifier } from 'moment';
-import { ChartConfig, DatasetField, ChartConfig } from '../../../config/types';
-
+import { ChartConfig, DatasetField } from '../../../config/types';
 import { TableData } from '../../../context/tableStateSlice';
 import { useSafeTranslation } from '../../../i18n';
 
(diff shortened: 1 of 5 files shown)


HarryMytilinaios:
Added `pointRadius` configuration in charts and have left the testing `pointRadius` `maxValue` `minValue` in `10-day-rainfall-anomaly`. Please let me know if this is the solution we want? @wadhwamatic @ericboucher

wadhwamatic:
@HarryMytilinaios - this is great. It does exactly what we need. Great work, thanks

frontend/src/components/DataDrawer/Chart/index.tsx
```diff
@@ -131,24 +140,44 @@ function formatChartData(data: TableData, config: ChartConfig) {
...
-    : indices.map((index, i) => ({
-        label: header[index] as string,
-        fill: config.fill || false,
-        backgroundColor: colors[i],
-        borderColor: colors[i],
-        borderWidth: 2,
-        data: tableRows.map(row => (row[index] as number) || null),
-        pointRadius: data.EWSConfig ? 0 : 1, // Disable point rendering for EWS only.
-      }));
+    ? tableRows.map((row, i) => {
```
ericboucher, line 143: Let's take this opportunity to dedup this part of the code

frontend/src/components/DataDrawer/Chart/index.tsx
```diff
@@ -27,6 +28,8 @@ function getChartConfig(
   xAxisLabel?: string,
   notMaintainAspectRatio?: boolean,
   legendAtBottom?: boolean,
+  minValue?: number,
```
ericboucher, line 31: Maybe we could pass in the complete `config` object to `getChartConfig` maybe?

ericboucher (approved):
Great work. Just two small suggestions to make the code a bit cleaner.

HarryMytilinaios:
@ericboucher decoupled a little bit the code inside the Chart component. I used smaller `useMemo` units that are a little bit more testable and are created by the specific props so this gives us a little bit of a performance optimization rather than having big blocks of code functions outside the component (This is an initiative to become the `Charts` component more of a `React` way). Ιn any case, the functions were not used anywhere else in the overall application. If in the future we need some reusability of these functions we can always address a custom reusable hook or utility.

HarryMytilinaios:
Of course @ericboucher I will try to avoid force pushing thanks for letting me know.

PR #797, opened 2023-04-11
Fixed overflow-x on tabs children

HarryMytilinaios:
Removed the horizontal scrollbar from the layers tab and I also took the initiative on fixing the scrolling in the charts section when the charts panel is opened to scroll only the panel. If we don't like this behavior please let me know to revert it back to the previous ux cc @wadhwamatic @ericboucher.


ericboucher:
Nice, could you add a few screenshots? Thanks!

wadhwamatic:
Looks great, thanks @HarryMytilinaios. And thanks for catching / fixing the charts issue. We had addressed the same problem in the analysis tab but missed it in charts. Over to @ericboucher for his final review

ericboucher (approved):

PR #798, opened 2023-04-12
Improved UI/UX of alerts popup box

HarryMytilinaios:
Also took the initiative on creating the same styles for all the buttons `GoTo` `CreateAlert` and `Legends` in order to be aligned horizontally all 3 of them. Also keep in mind that in the current PR I kept the `alertFormActive` from the `layers.json` to `true` for testing purposes. I add a screenshot below

Screenshot/video of feature:
Screenshot from 2023-04-12 15-52-42


wadhwamatic:
Looks good from my side, thanks @HarryMytilinaios and for unifying the style.

@ericboucher - down the road we should revisit the design of the go to and alert features but for the current needs, this is good.

frontend/src/components/MapView/index.tsx
```diff
@@ -596,15 +596,20 @@ function MapView({ classes }: MapViewProps) {
...
-                    isOpen={isAlertFormOpen}
-                    setOpen={setIsAlertFormOpen}
-                  />
-                ) : null}
+                {isShowingExtraFeatures && (
+                  <Grid item>
+                    <GoToBoundaryDropdown />
+                  </Grid>
+                )}
+                {appConfig.alertFormActive && isShowingExtraFeatures && (
```
ericboucher, line 604: Since both are dependent on `isShowingExtraFeatures`, you could test for `isShowingExtraFeatures` only once at the top I think.

ericboucher (approved):

PR #799, opened 2023-04-13
config updates and shift to new UX; COUNTRY=myanmar

wadhwamatic:
Migrate Myanmar with updated boundary files and removal of ACLED and tropical storm layers


ericboucher (approved):

frontend/src/config/myanmar/layers.json
```diff
@@ -1,11 +1,11 @@
 {
   "state_admin_boundaries": {
     "type": "boundary",
-    "path": "data/myanmar/mmr_admin1_boundaries.json",
```
ericboucher, line 4: Same as the other transfers, maybe delete the unused boundary files?

wadhwamatic: good catch. removed in the last commit

PR #818, opened 2023-05-03
Fix analysis for lower level admin boundaries; COUNTRY=mozambique

ericboucher:
There were two issues that this is trying to fix:
- For Sri Lanka, the matching was not happening correctly because we were comparing strings and numbers
- For Mozambique, the adminCode ADMIN1_PCODE does not appear in the level 3 boundary file that is used by default. Every admin boundary uses the same `source_id` key, which makes it hard for the matching to work properly.

@wadhwamatic  - Note that for Mozambique this will only be fixed for deployed versions - unless we change the keys - because we need to be able to send the file path to the API for calculations. Locally, it will still use admin 3 data and won't be able to match the keys.


ericboucher:
@wadhwamatic this fixes both countries. But I think we should improve the adminKeys for Mozambique and not use the same `source_id` key for each level.

wadhwamatic (approved):
gtg from my side @ericboucher. thanks!

PR #819, opened 2023-05-03
Add Analysis results as entries in the layers tab; COUNTRY=cambodia

HarryMytilinaios:
This branch adds the analysis results as entries in the `Layers` tab so the user can manipulate the analysis results also in the `Layers` tab. When a user untoggles the `Analysis Layer` the analysis is completely cleared. @wadhwamatic Right now the opacity and the download functionality is both in the `Layer` tab and in the `Legend` item (Additional functionality has been made to sync the opacity changes to both `Layer` item and `Legend` item). Do we want the handles to be in both 2 different places, or would be good to remove the opacity and download handles from the `Legend`?
cc @ericboucher 

Screenshot/video of feature:
Screenshot from 2023-05-03 17-03-34
Screenshot from 2023-05-03 17-03-48


HarryMytilinaios:
@wadhwamatic . I have filtered the exposure analysis results because I thought that they were already bounded in a specific layer, in our example Flood Extent. If we need also the exposure analysis results to be also an individual layer, I can add them of course. cc @ericboucher

HarryMytilinaios:
@ericboucher @wadhwamatic . Added Exposure analysis results as entries in the layers tab and also fixed the grouped layers that are added falsely many times when a layer had `group.activateAll` true. Also right now the implementation of the exposure analysis resuls as layer follows the functionality of the `Legend`. If the result is an Exposure analysis result the download CSV does not exist because a data transformation has to be done in order for this change to take place. This is the reason why the download CSV is missing also from the `Legend` component if we want to have also the export CSV download in both legend and layer entry. We need to change the data entries to be all in redux and from there to be accessible from `ExposureAnalysisActions` `Legend` and `LayerItem` components. Right now the data transformation has been individual taken place inside the `ExposureAnalysisActions` component

HarryMytilinaios:
@wadhwamatic @ericboucher, If this is ok I guess we could merge it for now?

wadhwamatic:
> @ericboucher @wadhwamatic . Added Exposure analysis results as entries in the layers tab and also fixed the grouped layers that are added falsely many times when a layer had `group.activateAll` true. Also right now the implementation of the exposure analysis resuls as layer follows the functionality of the `Legend`. If the result is an Exposure analysis result the download CSV does not exist because a data transformation has to be done in order for this change to take place. This is the reason why the download CSV is missing also from the `Legend` component if we want to have also the export CSV download in both legend and layer entry. We need to change the data entries to be all in redux and from there to be accessible from `ExposureAnalysisActions` `Legend` and `LayerItem` components. Right now the data transformation has been individual taken place inside the `ExposureAnalysisActions` component

@HarryMytilinaios - thanks for adding the exposure analysis as a result in the analysis section. You lost me a bit in the comments above. Are you planning to do additional work within this PR, or to create a new issue for the additional steps you've mentioned? Either approach is fine with me. What do you think @ericboucher ?

HarryMytilinaios:
@wadhwamatic it depends do we need the export CSV functionality in both `Legend` and `LayerItem` component or it is fine for now being only in the `ExposureAnalysisTabActions` ? If we need it I can do some additional work here in this branch in order to achieve it, or if we want a future enhancement we could merge this branch and add it to backlog. What do you think @ericboucher ?

ericboucher:
Let's merge this branch for now and open an issue to track the additional change.

@wadhwamatic because Exposure Analysis was built separately initially and was actually started from the legend, the behavior is still a bit different. In particular, the "Download CSV" button AND functions are only available at the bottom of the analysis tab. It is not easy to access this functionality from other places, like the layer kebab menu for example. So think it makes sense to make it more similar to other analysis types but beyond the scope of this particular PR.

ericboucher:
Let's merge this branch for now and open an issue to track the additional change.

@wadhwamatic because Exposure Analysis was built separately initially and was actually started from the legend, the behavior is still a bit different. In particular, the "Download CSV" button AND functions are only available at the bottom of the analysis tab. It is not easy to access this functionality from other places, like the layer kebab menu for example. So think it makes sense to make it more similar to other analysis types but beyond the scope of this particular PR.

ericboucher:
Let's merge this branch for now and open an issue to track the additional change.

@wadhwamatic because Exposure Analysis was built separately initially and was actually started from the legend, the behavior is still a bit different. In particular, the "Download CSV" button AND functions are only available at the bottom of the analysis tab. It is not easy to access this functionality from other places, like the layer kebab menu for example. So think it makes sense to make it more similar to other analysis types but beyond the scope of this particular PR.

ericboucher:
Let's merge this branch for now and open an issue to track the additional change.

@wadhwamatic because Exposure Analysis was built separately initially and was actually started from the legend, the behavior is still a bit different. In particular, the "Download CSV" button AND functions are only available at the bottom of the analysis tab. It is not easy to access this functionality from other places, like the layer kebab menu for example. So think it makes sense to make it more similar to other analysis types but beyond the scope of this particular PR.

wadhwamatic:
> Let's merge this branch for now and open an issue to track the additional change.
> 
> @wadhwamatic because Exposure Analysis was built separately initially and was actually started from the legend, the behavior is still a bit different. In particular, the "Download CSV" button AND functions are only available at the bottom of the analysis tab. It is not easy to access this functionality from other places, like the layer kebab menu for example. So think it makes sense to make it more similar to other analysis types but beyond the scope of this particular PR.

@ericboucher @HarryMytilinaios - a new issue was created https://github.com/WFP-VAM/prism-app/issues/826. This PR is GTG

wadhwamatic (approved):

Issue #643, opened 2022-11-22
Create a configuration option to prevent layers from appearing in the run analysis list


wadhwamatic:
When I'm configuring a WMS layer, I need an option to prevent the layer from appearing in the list of layers that can be used for an analysis. 

The reason for this is that we're modifying the way layers are organized and delivered from WFP's Humanitarian Data Cube (see PR https://github.com/WFP-VAM/prism-app/pull/642). The new configurations in that PR reflect changes to HDC, which uses a layer for a group of date intervals (10-day, 1-month, etc.) and styles to specify which date to utilize for that layer in PRISM. The new HDC approach will require changes on the PRISM side to use STAC instead of WCS requests for data used in the run analysis feature. This is captured in issue https://github.com/WFP-VAM/prism-app/issues/644.

In the meantime, we need an immediate solution to prevent misleading analyses based on a 'layer' but not a style. For example, in Cuba's[ layers.json](src/config/cuba/layers.json) in PR 642, all blended precipitation amount products use a single WMS layer 'rfb_blended_cub_dekad' but a separate style for 10-day, 1-month, 3-month, etc. Since style isn't considered in the run analysis request, the analysis for 10-day, 1-month, 3-month, etc. will be exactly the same, based on the default style which is 10-days. I want to be able to configure a WMS layer with a flag to hide it from the run analysis feature until we have a new solution in place.

wadhwamatic:
Addressed through https://github.com/WFP-VAM/prism-app/pull/829

PR #829, opened 2023-05-11
Add disableAnalysis flag to layerProps

reed1:
This fixes issue #643.

How to test the feature:

- Set "run_analysis' to false in layers.json, for example:
image

- The above will hide the layer in Analysis hazard layer selection
image

@wadhwamatic, I'm using "run_analysis" prop, but since boolean flag commonly defaults to false, show we use "hide_in_analysis" instead?


ericboucher:
Indeed, I think the flag should be reversed. Maye something like `disableAnalysis`? @wadhwamatic  what do you think?

frontend/src/components/MapView/Layers/LayerDropdown.tsx
```diff
@@ -95,12 +95,20 @@ function LayerDropdown({
...
-        layers: category.layers.filter(layer =>
-          layer.type === 'wms'
-            ? layer.type === type &&
-              [undefined, 'polygon'].includes(layer.geometry)
-            : layer.type === type,
-        ),
+        layers: category.layers.filter(layer => layer.type === type),
+      }))
+      // 4. additional filter for WMS layers
+      .map(category => ({
```
ericboucher, line 101: I would suggest creating a function `filterLayersForAnalysis` that explicitly details the steps we use to filter out layers.

This way you can call it in the `.map` above and don't need to loop through the list one more time with a no-op for most categories / layers

wadhwamatic:
> Indeed, I think the flag should be reversed. Maye something like `disableAnalysis`? @wadhwamatic what do you think?

Yes, that sounds good. Please make that change @reed1

reed1:
Thanks for the suggestion. I've updated the PR. @wadhwamatic @ericboucher

frontend/src/components/MapView/Layers/LayerDropdown.tsx
```diff
@@ -67,6 +67,16 @@ function LayerDropdown({
     tables: [],
   };
 
+  const filterLayersForAnalysis = (layer: LayerType) => {
+    if (layer.type === 'wms') {
+      return (
+        !layer.disableAnalysis &&
```
ericboucher, line 73: By design, I think layer.disableAnalysis should affect all layers, not just wms layers.

frontend/src/components/MapView/Layers/LayerDropdown.tsx
```diff
@@ -67,6 +67,16 @@ function LayerDropdown({
     tables: [],
   };
 
+  const filterLayersForAnalysis = (layer: LayerType) => {
```
ericboucher, line 70: Let's add some comments in this function please

frontend/src/components/MapView/Layers/LayerDropdown.tsx
```diff
@@ -67,6 +67,17 @@ function LayerDropdown({
     tables: [],
   };
 
+  // Filter out layers that are not supported by the analysis tool
+  const filterLayersForAnalysis = (layer: LayerType) => {
+    if (layer.disableAnalysis) {
+      return false;
+    }
+    if (layer.type === 'wms') {
+      return [undefined, 'polygon'].includes(layer.geometry);
```
ericboucher, line 76: can you explicit what type of geometry is not supported?

ericboucher: (point and linestring)

ericboucher (approved):

PR #832, opened 2023-05-12
Resolve Map flickering when switching left panel on/off

HarryMytilinaios:
This branch trying to refactor `MapView` component in order to resolve the [flickering in map layers](https://github.com/WFP-VAM/prism-app/issues/807).


ericboucher:
Nice work. This will need extensive testing @wadhwamatic around changing admin boundaries and map interactions as this touches on a lot of logic. This PR splits our MapView component into more manageable subsets.

wadhwamatic:
> Nice work @HarryMytilinaios!
> 
> This will need extensive testing @wadhwamatic around changing admin boundaries and map interactions as this touches on a lot of logic. This PR splits our MapView component into more manageable subsets.

Your comment scares me a bit! So, any tips on what I should test here?

ericboucher:
It should be ok but I just want to make sure we don't introduce any regressions.

I think testing the logic around admin boundaries and switching between different levels - admin2 data, back to admin3 and so forth. Adding an analysis for an admin2 layer...

@HarryMytilinaios any other suggestions on what to QA to make sure we don't introduce any regressions?

ericboucher:
@HarryMytilinaios it looks like the admin boundaries are staying below raster data, they should stay above.

https://prism-832.surge.sh/?hazardLayerIds=rainfall_dekad&date=2023-05-01

VS

https://prism-829.surge.sh/?hazardLayerIds=precip_blended_dekad&date=2023-05-01

HarryMytilinaios:
@wadhwamatic . No reason to be scared at all. This branch was originally created to resolve the validity infinite rendering bug, (unfortunately this isn't resolved yet). But the need to split the big `Mapview` component into smaller pieces to decouple the logic was necessary to identify this problem and possible future ones. So the `Map` component which renders the currrent Map and the layers inside of it, is different now from the `MapView` component which holds the logic from the whole app `SideBar`, `Legends` e.t.c. I currently tested the functionality to the above countries:
   - Rbd
   - Myanmar
   - Mozambique
   - Cambodia
I added all layers possible and the combinations of those and run all the analysis combinations, included the exposure analysis. From my side everything is working as expected. The problem with the admin boundaries I think is resolved @ericboucher, (I think it had to do with the deployed version that was old). So yes the tips I have is of course to test the toggle and untoggle the sidebar (To not flicker the layers in the map), and then general toggling, untoggling layers the combination between them and the analysis parts `Exposure` and not.

ericboucher (approved):

wadhwamatic:
@HarryMytilinaios and @ericboucher - I tried several scenarios, nothing broke! Nice work. Good to go from my side

wadhwamatic (approved):

PR #834, opened 2023-05-12
Fix exposure analysis for storms and earthquakes

ericboucher:
I believe https://github.com/WFP-VAM/prism-app/commit/48218fd4403e22bf9bf9634b8bf345936f09dc28 might have introduced this regression since we didn't have any way to test it for storms.

This is a temp hack


wadhwamatic:
Things are now working for tropical storm exposure analysis. Thanks a lot @ericboucher! 

I tested flood exposure analysis next and see the table results are incorrect (all have a value of 0) and the tooltip has a precision issue. The tooltip precision issue exists for tropical storms too.

Screenshot 2023-05-14 at 13 55 51

wadhwamatic:
Sorry @ericboucher - we found another issue. It seems we're dropping a number of townships during the analysis. For example, the township of Sittwe (TS_PCODE = MMR012001) in Rahkine State, Sittwe district is missing in our results. But it's a densely populated area and fully within the 120km/h wind buffer. 

When I run a local analysis I get 138,001 people exposed based on WorldPop constrained, UN-adjusted population.

Here's the result when I run zonal stats locally.
Screenshot 2023-05-14 at 15 20 50

And here's what we see in PRISM
Screenshot 2023-05-14 at 15 22 10

It looks like there's a general problem with boundaries. The list of townships is basically cutoff alphabetically after Rakhine > Sittwe. You can see this at play in the Go to feature too.
Screenshot 2023-05-14 at 15 24 33

wadhwamatic:
> Sorry @ericboucher - we found another issue. It seems we're dropping a number of townships during the analysis. For example, the township of Sittwe (TS_PCODE = MMR012001) in Rahkine State, Sittwe district is missing in our results. But it's a densely populated area and fully within the 120km/h wind buffer.

You were right @ericboucher - this was a boundary file issue. I fixed the boundaries and have this working locally and also see it working on the PR deployment https://prism-834.surge.sh/

For some reason, I am not seeing it updated on the surge deployment I've shared with others (https://prism-mmr-demo.surge.sh/). I tried clearing the cache locally on my browser and also cleared the stats API cache. I also tore down the surge project and redeployed but with no luck. If I deploy the same build to a new surge, it works as expected. Not worth spending further time on this, but something to note.

wadhwamatic:
> Things are now working for tropical storm exposure analysis. Thanks a lot @ericboucher!
> 
> I tested flood exposure analysis next and see the table results are incorrect (all have a value of 0) and the tooltip has a precision issue. The tooltip precision issue exists for tropical storms too.

@ericboucher - this is the issue we need to resolve next.

> Screenshot 2023-05-14 at 13 55 51

frontend/src/context/analysisResultStateSlice.ts
```diff
@@ -317,6 +443,40 @@ const createAPIRequestParams = (
...
+          ? mergedObject[objectKey] + obj[objectKey]
+          : obj[objectKey];
+      } else {
+        // eslint-disable-next-line fp/no-mutation
+        mergedObject[objectKey] = obj[objectKey];
+      }
+    });
+  });
+
+  // TEMPORARY LOGIC TO DEDUP POPULATION COUNTS FOR WIND BUFFERS.
```
ericboucher, line 463: @wadhwamatic should we move this logic to the backend instead of the frontend? What is your preferred behavior?

frontend/src/context/analysisResultStateSlice.ts
```diff
@@ -317,6 +443,40 @@ const createAPIRequestParams = (
   return apiRequest;
 };
 
+function mergeTableRows(tableRows: TableRow[]) {
+  const mergedObject: Record<string, any> = {};
+
+  tableRows.forEach(obj => {
```
ericboucher, line 449: we could probably simplify the function by using something like:

```
mergeWith({}, ...Object.values(objectThatNeedsToCollapse), (objValue, srcValue) =>
typeof objValue === "number" ? objValue + srcValue : undefined
)
```

HarryMytilinaios: @ericboucher I changed a little bit the code there by using `array.reduce`.

HarryMytilinaios:
@ericboucher @wadhwamatic . Fixed the problem with the exposure analysis results and thee differs between tropical storms and the rest exposure analysis  results. I think that right now, works as expected. Tested it to a bunch of countries and seemed to work solidly. Moreover added the `escapeCell` to the values of the results because some of them are not treated correctly this is now fixed. @ericboucher also I took the initiative to split the functionality a little bit to make smaller units to be more readable.

frontend/src/context/analysisResultStateSlice.ts
ericboucher: Same with the rounding of values, should this happen in the backend?

frontend/src/utils/analysis-utils.ts
```diff
@@ -696,7 +703,12 @@ export function getAnalysisTableColumns(
   ];
 
   if (analysisResult instanceof ExposedPopulationResult) {
-    return analysisTableColumns;
+    const extraCols = analysisResult?.tableColumns.map((col: string) => ({
+      id: col,
+      label: col, // invert maps from computer name to display name.
```
ericboucher, line 708: comment not needed here

frontend/src/utils/analysis-utils.ts
```diff
@@ -461,6 +462,9 @@ export async function loadFeaturesClientSide(
 }
 
 export function quoteAndEscapeCell(value: number | string) {
+  if (isUndefined(value)) {
```
ericboucher, line 465: should quoteAndEscapeCell explicitely accept undefined values then? it might help simplify the typing logic in some places

frontend/src/context/analysisResultStateSlice.ts
```diff
@@ -408,16 +530,29 @@ export const requestAndStoreExposedPopulation = createAsyncThunk<
...
       featuresWithBoundaryProps,
       adminBoundariesData,
       apiRequest.group_by,
       null,
       [], // no extra columns
       true,
+      key,
     );
 
+    if (key) {
```
ericboucher, line 549: This might not be the best test to decide if we should merge rows or not. Let's think about a few other options to make this clearer

frontend/src/context/analysisResultStateSlice.ts
```diff
@@ -317,6 +396,49 @@ const createAPIRequestParams = (
...
+  const oneHundredTwenty = Number(get(mergedObject, '120 km/h', 0));
+  const ninety =
+    Number(get(mergedObject, '90 km/h', 0)) -
+    Number(get(mergedObject, '120 km/h', 0));
+  const sixty =
+    Number(get(mergedObject, '60 km/h', 0)) -
+    Number(get(mergedObject, '90 km/h', 0));
+
+  // eslint-disable-next-line fp/no-mutation
+  mergedObject['120 km/h'] = Number(Math.ceil(oneHundredTwenty));
```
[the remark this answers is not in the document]
ericboucher: Sorry I was simply thinking about the rounding part

frontend/src/utils/analysis-utils.ts
HarryMytilinaios: Fixed

frontend/src/context/analysisResultStateSlice.ts
HarryMytilinaios: From the testing I've made to some countries, there is a pattern that when a key is present from the `exposure` property inside the `params` then we are reffering to tropical storms. However we could use a more unique key for this. However I am not realy sure what exactly though.

frontend/src/components/MapView/MapTooltip/index.tsx
```diff
@@ -7,40 +7,117 @@ import {
...
+    return popupLocationLocalName;
+  }, [i18n, popupLocationLocalName, popupLocationName]);
+
+  const popupData = useMemo(() => {
+    return popup.data;
+  }, [popup.data]);
+
+  const popupWmsGetFeatureInfoLoading = useMemo(() => {
+    return popup.wmsGetFeatureInfoLoading;
+  }, [popup.wmsGetFeatureInfoLoading]);
```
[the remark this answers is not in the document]
HarryMytilinaios: I added small useMemo pure functions to have better handling on the data inside the component and to know any time what exactly is rerendered and not everything. I can remove them of course, if we find it inefficient, I mean yes this makes the file larger and the lines of code but I thought it would be a lot cleaner and readable

wadhwamatic:
Thank you @ericboucher and @HarryMytilinaios for the big push on this. I tested across as many scenarios as I could and found no issues. We lost access to earthquake data layers on GeoServer so I couldn't test that but I hope it won't differ from the tropical storm layer analysis when that access is restored.

This is GTG from my side. Thanks again

frontend/src/context/analysisResultStateSlice.ts
ericboucher: keeping it here for now as suggested by @wadhwamatic

Issue #814, opened 2023-04-29
[Bug]: Date warnings appear twice


ericboucher:
### What happened?

When switching to a new layer. If there is a date conflict, we show a warning and inform the user of the new selected date. For some reason, this warning appears twice. We should only display the warning once.

<img width="1012" alt="Screenshot 2023-04-28 at 2 33 54 PM" src="https://user-images.githubusercontent.com/16843267/235311723-fd6f8d75-b32c-46c7-9d30-e0f68acc6a82.png">

### Which country / deployment are you running?

All

wadhwamatic:
@ericboucher - yes this is annoying and something that was fixed before but was reintroduced at some stage - I think with the new UX

PR #842, opened 2023-05-15
Fix date warnings appear twice

tboulis:
With this fix, we basically block the display of duplicate notifications by using a hash which is generated by the notification message.

This fixes issue [#814].

How to test the feature:

1. Select Cambodia
2. Enable layer X
3. Select a date that is out of range for a layer Y
4. Enable layer Y along with X
5. Expect to see a single notification saying there is no data for that date

Screenshot/video of feature:

https://github.com/WFP-VAM/prism-app/assets/24255041/a9a924c0-1bd0-47d7-831a-4c94f25af8c9

### DIFF
--- a/frontend/src/context/notificationStateSlice.ts
+++ b/frontend/src/context/notificationStateSlice.ts
@@ -17,7 +17,7 @@ type NotificationConstructor = {
 export class Notification {
   readonly message: string;
   readonly type: Color;
-  readonly key: string; // each notification needs a unique ID. Date.now() seems to do the job
+  readonly key: string; // each notification needs a unique ID. We generate a hash from the notification message
 
   constructor({ message, type }: NotificationConstructor) {
     this.type = type;
--- a/frontend/src/context/notificationStateSlice.ts
+++ b/frontend/src/context/notificationStateSlice.ts
@@ -6,6 +6,7 @@ import {
 } from '@reduxjs/toolkit';
 import { Color } from '@material-ui/lab';
 import type { AppDispatch, RootState } from './store';
+import { stringHash } from '../utils/string-utils';
 
 // to test notification reaction to various error codes, http://httpstat.us/404 can be used where 404 is the status to test.
 type NotificationConstructor = {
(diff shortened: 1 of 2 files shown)


frontend/src/utils/string-utils.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable no-bitwise, fp/no-mutation */
+
+// https://stackoverflow.com/a/7616484
+export const stringHash = (s: string): string => {
```
ericboucher, line 4: nice, except that disabling two es-lint rules is a yellow flag :)

Maybe we could use something like this one-liner?

```
const hashCode = (str) => [...str].reduce((s, c) => Math.imul(31, s) + c.charCodeAt(0) | 0, 0) 
```

See https://gist.github.com/hyamamoto/fd435505d29ebfa3d9716fd2be8d42f0

frontend/src/context/notificationStateSlice.ts
```diff
@@ -37,12 +39,19 @@ export const notificationStateSlice = createSlice({
...
-    }),
+    ) => {
+      const { notifications, ...rest } = state;
+      const notification = new Notification(payload);
+      if (notifications.find(n => n.key === notification.key)) {
+        return state;
+      }
+      return {
+        ...rest,
+        notifications: notifications.concat(notification),
```
ericboucher, line 52: could notifications be changed to a set instead? so that you don't need to do the verification above?

ericboucher:
@tboulis great find. Just a quick note, when possible, try to link your PRs to the opened issue through GitHub for easier tracking :)

ericboucher:
> > @tboulis great find. Just a quick note, when possible, try to link your PRs to the opened issue through GitHub for easier tracking :)
> 
> Of course Eric, I've already done this both in the PR description and the commit message

Yep but it's actually a bit picky in terms of formatting and I had to link it from the development pannel on the right side of the page. No big deal, just a note. In the PR body for example, "This PR fixes #834" needs to be the very first line and should not have the brackets `[]`

frontend/src/utils/string-utils.ts
ericboucher: ok

frontend/src/context/notificationStateSlice.ts
ericboucher: Either a set or a map could work. I think the notifications are not used it too many places so if this works we should be fine

ericboucher (approved):

PR #847, opened 2023-05-22
823 - Improve App Responsiveness

jbperidypathtech:
This fixes issue 823.
- LeftPanel (Layers, Chart, Analysis)
- Legend

How to test the feature:
- Test with Desktop view and Tablet view

Screenshot/video of feature:

Out of scope: mobile view

### DIFF
--- a/frontend/src/config/types.ts
+++ b/frontend/src/config/types.ts
@@ -695,7 +695,7 @@ export type UserAuth = {
 
 export enum PanelSize {
   folded = '0vw',
-  medium = '30vw',
-  large = '60vw',
-  xlarge = '80vw',
+  medium = '400px',
+  large = '800px',
+  xlarge = '900px',
 }
--- a/frontend/src/components/MapView/DateSelector/index.tsx
+++ b/frontend/src/components/MapView/DateSelector/index.tsx
@@ -347,7 +347,7 @@ const styles = (theme: Theme) =>
       position: 'absolute',
       bottom: '8%',
       width: '100%',
-      zIndex: 1,
+      zIndex: 5,
     },
 
     datePickerContainer: {
--- a/frontend/src/components/MapView/LeftPanel/AnalysisPanel/index.tsx
+++ b/frontend/src/components/MapView/LeftPanel/AnalysisPanel/index.tsx
@@ -1147,8 +1147,8 @@ const styles = (theme: Theme) =>
     },
     analysisPanel: {
       position: 'relative',
-      paddingTop: 30,
       width: PanelSize.medium,
+      height: '100%',
     },
     exposureAnalysisLoadingContainer: {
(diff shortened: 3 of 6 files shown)


ericboucher:
@jbperidypathtech thanks! Could you add a few screenshots to your PR body?

wadhwamatic:
This is a good start, thanks @jbperidypathtech! I found some issues when 4 or more charts are loaded, and also when I ran an exposure analysis using 1920 x 1080.

To reproduce the exposure analysis, under layers, select tropical storms > tropical storms. Once the layer has loaded, there's an icon for the exposure analysis in the layer section of the left panel. For the chart issue, select any 4 charts at any admin level.

Screenshot 2023-05-22 at 18 22 14

Screenshot 2023-05-22 at 18 23 28

JulienPathtech (approved):

jbperidypathtech:
@wadhwamatic thanks for your comments. It should be better now !
Capture d’écran du 2023-05-24 11-33-06
Capture d’écran du 2023-05-24 11-33-41

wadhwamatic:
Thanks @jbperidypathtech, this is looking good! I see one small issue in the charts feature. When one of the data labels is long, like 'rainfall anomaly', the second label ('normal' in this case) shows up on a second line but they are not aligned. See your 2nd screenshot, and the screenshot below. Can we left align these labels?


Screenshot 2023-05-24 at 16 27 06

jbperidypathtech:
Hi @wadhwamatic, what I can do is to always left align the label (see screenshot). If you rather like to keep center align labels when chart is larger, it should probably be possible but more complicated to do because the chart object is a lib and we don't have full ability to do what we want. What do you think about it?
Capture d’écran du 2023-05-25 09-06-59

wadhwamatic:
Thanks @jbperidypathtech. The always left aligned solution doesn't look ideal (sorry) and we don't need to make this overly complicated by making it conditional. One other thing to try - reduce the width of the color bar in the legend label and revert to centered. It could simply be a square.

jbperidypathtech:
Hi @wadhwamatic, reducing width of the color box in the legend seems to be fine.
Capture d’écran du 2023-05-30 08-56-54

ericboucher (approved):

wadhwamatic (approved):
This looks good now, thanks @jbperidypathtech !

frontend/src/config/types.ts
```diff
@@ -695,7 +695,7 @@ export type UserAuth = {
 
 export enum PanelSize {
   folded = '0vw',
-  medium = '30vw',
-  large = '60vw',
-  xlarge = '80vw',
+  medium = '400px',
```
ericboucher, line 698: These updates create a regression on desktop.

PR #849, opened 2023-05-23
Components cleanup

HarryMytilinaios:
This branch cleans up the Unused Components, which are left behind after revamping. The components are 

-  AdminLevelDataDownloadButton.tsx
-  StacRasterDownloadButton.tsx
-  exposedPopulationAnalysis.tsx
-  Dataviewer

After thorough testing and digging those components are no longer used,  not because we lost the functionalities on them, but because they were revamped into new UX.


ericboucher:
@HarryMytilinaios Thanks for the cleanup!
In your PR body, could you explain to which files each functionality has moved? I just want to make sure that we are not removing any functionality, just components.

HarryMytilinaios:
@ericboucher The functionality of the legends has moved entirely in the `index.tsx` of the `Legends` component, more specifically the `AdminLevelDataDownloadButton.tsx` and `StacRasterDownloadButton.tsx` are described as main layers so the download functionality is entirely in the panel tab `Layers` and we don't have any interaction with the `LegendItem` for the primary layers. The functionality of the `exposedPopulationAnalysis.tsx` is moved in the generic component `AnalysisDownloadButton.tsx`. The layers that are derived from an analysis result has their own manipulation in the `LegendItem` which differs from the rest `Layers` the analysisDownloadButton handles the download as csv, or geojson of the analysis result and exposure analysis result. Be careful here, the exposure analysis result does not have yet, a generic way of downloading the csv from the `LegendItem`  and the handle for the downloadCSV, is only in the exposureAnalysisResultActions below the result in the leftPanel as we have discussed. So the component `AnalysisDownloadButton` filters the result when it is from ExposedPopulation when we press the downloadCSV button. 
The thing that troubles me though,  is that the `DataViewer` component was handling a `loadDataset` action and displayed it as a Chart inside of it. I searched for the `loadDataSet` action in redux and as it happens it is not called from anywhere. But when we had the `DataViewer` component, in the `MapView` we were never rendering it. Did we missed that functionality when we were making the UX-revamp? I am not really sure, can you give me more information about the loadDataSet and what it was reffering. I am not really sure that we are using the funcionality of the extra reducers loadDataSet anymore. 
Moreover I refactored a bit the `Legends` component by splitting the `LegendItem` component and `LegendImpact` result component into seperate folders with their own tests. Lastly I found a minor bug in the Layers Tab when calculating the active layers it was using a wrong filter and it did not show the correct number of active layers I also fixed that.

frontend/src/components/MapView/Legends/LegendItem/index.tsx
```diff
@@ -0,0 +1,189 @@
...
+              step={0.01}
+              min={0}
+              max={1}
+              aria-labelledby="opacity-slider"
+              onChange={(e, newValue) =>
+                handleChangeOpacity(
+                  e,
+                  newValue as number,
+                  map,
+                  // TODO - get updated layer id in groups after switch. See issue #743
```
ericboucher, line 83: @HarryMytilinaios is this resolved?

ericboucher (approved):

frontend/src/components/MapView/Legends/LegendItem/index.tsx
HarryMytilinaios: Yes @ericboucher I think this is resolved I am removing the TODO.

PR #850, opened 2023-05-24
Fix date validity with modes backward, both and forward

HarryMytilinaios:
This branch fixes the date validity with mode `backward` or `both`. @wadhwamatic I have included the changes `backward` and `both` in the `layers.json`. If we are ok with it after your review of course please let me know to revert them. Cc @ericboucher


HarryMytilinaios:
@wadhwamatic I added the configuration that you did in your branch which is as follows:

- This layer uses the 'backwards' config option: Cadre Harmonisé > Phase Classification > Phase classification
- And this layer uses the 'both' config option: Cadre Harmonisé > Population > Population in phase 3 to 5

The changes are already took place in the `layers.json` so, after your review we can remove them.

wadhwamatic:
Thanks @HarryMytilinaios. This works now when we have just the CH data layers active. But when I then also activate a WMS layer (rainfall amount > rainfall aggregate 10-day), the app crashes (screen goes blank)

https://prism-850.surge.sh/?date=2023-06-01&baselineLayerId=ch_pop3_5&hazardLayerIds=rainfall_dekad

Note, if you load the rainfall layer first and then the CH layer, the app works as expected. Interestingly, the change in order of the URL parameters also handles this correctly
https://prism-850.surge.sh/?date=2023-05-21&hazardLayerIds=rainfall_dekad&baselineLayerId=ch_pop3_5

HarryMytilinaios:
@wadhwamatic @ericboucher . After thorough testing with different validity values in various different layers. I found the problem. So the problem had 2 different subproblems. The first was that when we calculated the `backward` and `both` the values were not calculated correctly. This is now fixed. The other and most effective problem was with the forward validity the problem was that if the default dates of a layer were too close together and we added `forward`, `backward` or `both` validities to them the were creating duplicate items inside of the `availableDates` that was causing a lot of different problems and the most important one was the rerendering. Because the `DateSelector` and `MapView` were trying to updateHistory both of them the one rendered the other infinitely. I resolved the duplication problem by creating unique arrays with all the necessary available dates, based on the validity. So when dates were overlapping it will only keep one date. The other problem was to refactor a bit both the `Mapview` component and `DateSelector` component. So the updateHistory on the app initialization will only been done in the `Mapview` component and not both, regardless of whether the `DateSelector` component updates its own state to facilitate the changes from the layer selection. However the `DateSelector` component will update the url and the history ONLY if the user has interaction with it. So for example if he clicks on an available date or changes something from datepicker, or drags the caret to a selected date. So this branch also fixes all of the validity problems hopefully. 
@wadhwamatic I have added also a validity forward of 10 days in rainfall aggregate (for testing purposes) in the config in this PR, when you review all the different changes we can easily revert it to normal I will link also the forward validity bug in this PR.

ericboucher:
Nice work @HarryMytilinaios! Let's QA this some more to be sure but it looks like we're on the right track :)

frontend/src/utils/server-utils.ts
```diff
@@ -207,44 +207,45 @@ const updateLayerDatesWithValidity = (layer: ValidityLayer): DateItem[] => {
...
-      return mergedDateItems;
+      return [...filteredDateItems, ...dateItemsToAdd];
     },
     [],
   );
 
-  const dateItems = [...dateItemsDefault, ...dateItemsWithValidity];
-
-  return sortBy(dateItems, 'displayDate');
+  return sortedUniqBy(
```
ericboucher, line 245: @HarryMytilinaios @wadhwamatic why might want to think a bit harder about how we select dates here.

The underlying "tricky" question is: which data do we select if have multiple overlapping options after adding the validity periods.

- Do we select the closest real value?
- Do we select the closest past value?

frontend/src/utils/server-utils.ts
```diff
@@ -299,22 +300,17 @@ export async function getLayersAvailableDates(): Promise<AvailableDates> {
       };
     });
 
-  const mergedLayersWithUpdatedDates = Object.entries(mergedLayers).reduce(
-    (acc, [layerKey, dates]) => {
-      const layerWithValidity = layersWithValidity.find(
-        validityLayer => validityLayer.name === layerKey,
-      );
+  return Object.entries(mergedLayers).reduce((acc, [layerKey, dates]) => {
```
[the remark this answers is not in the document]
HarryMytilinaios: of course

frontend/src/utils/server-utils.ts
ericboucher: Yes I agree, but we still need to decide how to dedup them.

wadhwamatic:
This is working well now, thanks a lot @HarryMytilinaios. I didn't do a specific comparison, but it feels much quicker than before when loading dates across layers. Nice work!

I'd like to test in the context of Cambodia next where we have a few other layers with varying dates. I'll try to get to that later today and report back.

frontend/src/utils/server-utils.ts
ericboucher: @wadhwamatic we'll need your input here :)

frontend/src/context/mapStateSlice/index.ts
```diff
@@ -41,35 +42,51 @@ const initialState: MapState = {
...
 };
 
-// Order layers to keep boundaries and point_data on top. boundaries first.
-export function layerOrdering(a: LayerType, b: LayerType) {
-  if (a.type === 'boundary') {
-    return -1;
-  }
-  if (b.type === 'boundary') {
-    return 1;
+const getTypeOrder = (layer: LayerType) => {
```
ericboucher, line 45: This seems a bit duplicative of the existing function `getLayerGeometry`. You might want to combine these two?

frontend/src/utils/server-utils.ts
wadhwamatic: I think I'm following what we're trying to decide here... 

So let's say the first layer has validity range of 10-days forwards, and has 'real dates' on 1-May, 11-May, and 21-May. The latest is loaded which is valid from 21-May to 31-May. 

Then a second layer is loaded which is valid on 11-May and 24-May with three days backwards. Even though 11-May is an 'exact' match for the reference date, we should use 24-May. 

Basically any valid date, whether it's the 'real date' or an interpreted date should be treated the same. We're really trying to show that the date of observation is not necessarily the only correct date for a measurement.

I hope I understood this right and what I'm saying is clear!

frontend/src/context/mapStateSlice/index.ts
HarryMytilinaios: The `getLayerGeometry` is slightly different function I am not sure that those 2 can be combined together. The `getLayerGeometry` function is only used to get the `geometryIconSrc` and nothing more. As I can see in the files. The `getTypeOrder` function is a static utility function that is not exported at all and just tells if a type of a layer is a polygon, also  the `getLayerGeometry ` returns `'point' | 'polygon' | 'raster' | 'unknown'` which are all the known geometry types of a layer. I don't see how those 2 can be combined together.

frontend/src/utils/server-utils.ts
HarryMytilinaios: I don't really follow here what are you saying I am gonna give you an example. We have 2 dates  for 2 different layers `11-May` and `21-May` and we have `forward` 20 days at the first and 10 days `both` at the second. The funcionality above takes every single default date in our case `11-May` and `21-May` and tries and adds or subtracts or both the number of days each specific date has a validity. After those calculations in our particular case we would end up with an array with duplicated dates firstly from the `forward` of the first date and secondly of the `both` of the second date. Those dates will be duplicates, the above code takes a unique of those. What exactly do we care if we will keep the layer's date overlap or second's? Am I missing something here?

HarryMytilinaios: I am sorry but I don't follow here. Every layer has its own date array. The duplicates were being on the same layer's date array. We can have duplicate entries in the date array of different layers. This is how the intersected dates occur. We cannot have duplicate dates on the same layer's date array. Is there something I am missing here?

HarryMytilinaios: So @ericboucher explained to me what we are reffering here, and right now this is  exactly what this functionality does. The `sortUniqBy` after the array of dates is sorted takes the most recent actual date. so all good. I didn't understand at the first place what you guys were reffering.

HarryMytilinaios:
@ericboucher @wadhwamatic. If this is ok let me know to revert the changes I've done in `layers.json` about the validities and give it a merge

Issue #725, opened 2023-01-26
[Feature Request]: Update PRISM header with the option to modify text, add a logo and add a subtitle


wadhwamatic:
### Provide a clear and concise description of what the problem is.

Some deployments want to customize the 'PRISM' text header on the top left of the navigation. They want to:
1. Be able to write title text besides PRISM
2. Add a logo instead of the text PRISM
3. Add a subtitle that would appear in smaller font below the current PRISM text

### Provide a clear and concise description of what you want to happen.

In prism.json, add new attributes called 'title', 'subtitle', and 'logo'. Title and subtitle should be text fields which can be translated. Logo should point to a local image which is scaled to fit within the header section.

PR #851, opened 2023-05-29
Custom header (title, subtitle, and logo)

hav142:
This fixes issue #725.

How to test the feature:
- Add new config to prism.json for each country, i.e:
```
{
  "country": "Mongolia",
  "header": {
     "title": "PRISM v.3",
     "subtitle": "test subtitle",
     "logo": "images/wfp_logo.png"
   },
  "icons": {
...
```
- The logo will dynamically adjust its size based on the height, allowing it to accommodate logos of different size.

before (or without config):
image

after:
image
image


HarryMytilinaios:
Sorry to intervene @hav142 @ericboucher @wadhwamatic in this PR I know I am not part of the reviewers, but because this PR I think it has similarities with an issue assigned to me https://github.com/WFP-VAM/prism-app/issues/629 . I'd like to ask on how are we supposed to translate those keys (if we wanted to, of course), that are part of the `prism.json`. Because as far as I know, the translations are derived from the `tranlsations.json` file of every country.

frontend/src/components/NavBar/index.tsx
```diff
@@ -48,19 +51,31 @@ function NavBar({ classes }: NavBarProps) {
     </Grid>
   ));
 
+  const title = get(appConfig, 'title', 'PRISM');
```
ericboucher, line 54: I think you should create a `header` config object within the config instead of having everything at the top level

frontend/src/components/NavBar/index.tsx
```diff
@@ -48,19 +51,31 @@ function NavBar({ classes }: NavBarProps) {
...
-            </Typography>
+            {logo && <img className={classes.logo} src={logo} alt="logo" />}
+            <Box display="flex" flexDirection="column">
+              <Typography
+                variant="h6"
+                className={classes.title}
+                component={Link}
+                to="/"
+              >
+                {t(title)}
```
ericboucher, line 71: @HarryMytilinaios the English version can be passed as the key directly as it is done here

wadhwamatic:
This is a good start @hav142, thanks! When I add an empty 'title' to the config, we just see the logo, which is good. I tried extending over muliple lines and clearly there's a limit which is fine as I don't see a need to have multiline titles / subtitles. However, please do look at https://github.com/WFP-VAM/prism-app/pull/847 and see how this will work adaptively 

And as @HarryMytilinaios mentioned, let's make sure we have a means to handle translations. I didn't try that yet but can once you've confirmed that is also ready for testing.

Screenshot 2023-05-31 at 00 02 02

after reducing the page width:
Screenshot 2023-05-31 at 00 02 19

frontend/src/components/NavBar/index.tsx
hav142: Thanks @ericboucher  for the correction. So, all country will have default config where the title is set to 'PRISM' like `'header': { title: 'PRISM' }`?

ericboucher: So we would do something like that:

{title, subtitle, logo} = get(appConfig, 'header', {title: 'PRISM'});

ericboucher:
Just confirming that we are ok to move the default title all the way to the left:

BEFORE:
Screenshot 2023-06-02 at 6 24 59 PM

AFTER:
Screenshot 2023-06-02 at 6 24 46 PM

hav142:
@ericboucher I have updated the grid layout to enhance the responsiveness of the title, particularly for longer titles that were previously displayed incorrectly on tablets or smaller screens. To address this issue, I have moved the default title to the left, which has the same result as the example provided by Amit above, where the title and subtitle were also aligned to the left when they were longer.

@wadhwamatic the translation for the title and subtitle should already be functioning correctly.

wadhwamatic:
@hav142 @ericboucher - the previous alignment is preferable. I can do some testing but not for a while. Please try on your side, including with translations.

ericboucher (approved):

ericboucher:
@hav142 you could simply use a breakpoint `.up('md')` to keep the alignment we had for desktop but remove it for smaller screens. What do you think?

https://mui.com/material-ui/customization/breakpoints/

hav142:
@ericboucher if we want to maintain the previous alignment, we can put back `justifyContent: 'center'` (without breakpoint). Instead, we can set the breakpoint in the Grid component, like `xs={6} md={9}`. This way the logo will have more space on smaller screen, ensuring consistency with larger screen. What do you think? 

image

Issue #791, opened 2023-04-07
[Feature Request]: Modify WFP chart requests 


wadhwamatic:
### Provide a clear and concise description of what the problem is.

WFP's HDC API is now available via an organizational API Gateway. We have to modify our requests to make use of this new access method. In addition, the request parameters have changed slightly.

### Provide a clear and concise description of what you want to happen.

1. A token will be required to access the API through WFP's Gateway. We have tokens available for PRISM which I can share for testing and production when ready.
2. Change request parameters (see instructions below)
    - admin_id becomes id_code
    - new vam code for blended products = rfb
    - change get request URL to https://api.wfp.org/hdc/1.0.0/stats/admin

Instructions for the HDC API:
 
> Get data extractions globally averaged over administrative boundary levels 0, 1 and 2. The extractions are performed for three coverages that can be selected using the 'coverage' argument: full (all pixels), crop (only cropland pixels), past (only pasture pixels).
> 
> The endpoint returns a dictionary with entries: - data (containing the extracted data and added indicators) - valids (containing the number of valid pixels for the admin used in the extraction) - date (the corresponding dates of the observations)
> 
> Args: id_code: Identifier for administrative region level: Administrative level (0, 1, 2) vam: Three letter vam code for variable to be retrieved (rfh, vim, rfb) coverage: Coverage version of extractions (either full, crop or past) start: Filter by start date (format YYYY-MM-DD) end: Filter by end date (format YYYY-MM-DD)

Example request:
curl -X GET "https://api.wfp.org/hdc/1.0.0/stats/admin?id_code=44&level=0&coverage=full&vam=rfh&env=prod" -H "accept: application/json" -H "Authorization: Bearer *insert-token*"

JulienPathtech:
Hello @ericboucher @wadhwamatic 
If I understand correctly we need to change a public ressource ( which is at the moment something like : https://api.earthobservation.vam.wfp.org/stats/admin/fetch... ) with the new protected ressource directly in the frontend app. 
It also means that the token will be exposed in the app and hence compromised ? Or am I missing something ?

ericboucher:
@JulienPathtech - @reed1 will be working on this. But most likely we will do a simple re-routing through the backend API to authenticate the call, as we do for KOBO and ACLED

reed1:
[...]

Issue #839, opened 2023-05-13
[Bug]: ACLED or KOBO should not prevent the rest of the app from loading


ericboucher:
### What happened?

- 

### Which country / deployment are you running?

-

wadhwamatic:
@ericboucher - is this the same issue, or something else? https://github.com/WFP-VAM/prism-app/issues/593

HarryMytilinaios:
@ericboucher @wadhwamatic those 3 issues are linked and I think they have the same solution. https://github.com/WFP-VAM/prism-app/issues/824 https://github.com/WFP-VAM/prism-app/issues/593. I think we can merge them into one issue.

PR #860, opened 2023-06-03
Add an HDC api endpoint to handle chart requests

reed1:
This fixes issue #791 

How to test the feature:

- Set HDC_TOKEN env in fastapi
- Point the chart_data url to "/hdc" of fastapi server instead of api.earthobservation.vam.wfp.org (changes already included in this PR)

Only warning (not error) will show if the HDC_TOKEN is not set. I saw bug #839, so I thought that should apply to backend as well


ericboucher:
@wadhwamatic would you like us to deploy the backend part of this PR for testing?

wadhwamatic:
Yes please!

frontend/src/config/cambodia/layers.json
```diff
@@ -227,7 +227,7 @@
       "styles": "rfh_16_0_300"
     },
     "chart_data": {
-      "url": "https://api.earthobservation.vam.wfp.org/stats/admin/fetch",
+      "url": "https://prism-api.ovio.org/hdc",
```
ericboucher, line 230: Now that the data source is going to be the same across countries, I think it might be worth extracting into a config variable as "chart_api" or similar and make it a default.

This way you can remove "chart_data.url" from all the layers.json. But keep it available in case we need to access charts from other sources.

frontend/src/context/datasetStateSlice.ts
```diff
@@ -247,7 +247,7 @@ export const loadAdminBoundaryDataset = async (
 
   const hdcRequestParams = {
     level,
-    admin_id: adminCode,
+    id_code: adminCode,
```
ericboucher, line 250: Is this change necessary? Do we know what the most common standard is?
admin_id vs id_code?

Linked to my comment on URLs, ideally the frontend stays indifferent as to where the data is coming from.

reed1:
@ericboucher I updated the branch based on your suggestions. I put chartApiUrl below serversUrls in prism.json. Also restored client request with admin_id.

frontend/src/config/afghanistan/prism.json
```diff
@@ -13,6 +13,7 @@
   "serversUrls": {
     "wms": ["https://api.earthobservation.vam.wfp.org/ows/wms"]
   },
+  "chartApiUrl": "https://prism-api.ovio.org/hdc",
```
ericboucher, line 16: If this is going to be the same for every country, this could even live in a constants.ts, similarly to what's done in the PR linked above

api/app/hdc.py
```diff
@@ -0,0 +1,33 @@
+"""Get data from HDC"""
```
ericboucher, line 1: Add a comment explaining what HDC is

ericboucher:
The API part looks good, I'll deploy now so that we can test.

api/app/main.py
```diff
@@ -176,6 +177,11 @@ def get_acled_incidents(request: Request):
     return Response(content=response.content, media_type="application/json")
 
 
+@app.get("/hdc")
+def wrap_get_hdc_stats(request: Request):
```
ericboucher, line 181: you need to better type this request

reed1:
@ericboucher if possible, can you forward me error log for 

https://prism-api.ovio.org/hdc?level=0&admin_id=170&coverage=full&vam=rfh&start=2022-06-07&end=2023-06-07

If it's requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://api.wfp.org/hdc/1.0.0/stats/admin?level=0&id_code=170&coverage=full&vam=rfh&start=2022-06-07&end=2023-06-07
ERROR:uvicorn.error:Exception in ASGI application

Then the HDC_TOKEN env is invalid or not set. I've sent you the token in Slack, let me know if that works, thank you

ericboucher:
Some of the charts are not working it seems. I get the following logs:

```
api        | requests.exceptions.HTTPError: 422 Client Error: UNPROCESSABLE ENTITY for url: https://api.wfp.org/hdc/1.0.0/stats/admin?level=0&id_code=170&coverage=full&vam=rfb&start=2022-06-07&end=2023-06-07
```

Screenshot 2023-06-07 at 10 11 16 AM

reed1:
Confirmed, the rfb doesn't work,

```
  // HDC API expects a parameter which depends on the layer
  // rainfall = rfh; ndvi = vim; blended = rfb
  const getVamParam = () => {
    switch (true) {
      case serverLayerName.includes('vim'):
      case serverLayerName.includes('viq'):
        return 'vim';
      case serverLayerName.includes('blended'):
        return 'rfb';
      default:
        return 'rfh';
    }
  };
```

vim and rfh work, but rfb doesn't. @wadhwamatic perhaps you can notify the API owner on HDC.

curl -H "Authorization: Bearer TOKEN" 'https://api.wfp.org/hdc/1.0.0/stats/admin?level=0&id_code=170&coverage=full&vam=rfh&start=2022-06-07&end=2023-06-07'

does not works, it gives:
```
{"detail":[{"ctx":{"enum_values":["rfh","vim"]},"loc":["query","vam"],"msg":"value is not a valid enumeration member; permitted: 'rfh', 'vim'","type":"type_error.enum"}]}
```

wadhwamatic:
@reed1 - I sent a message to Valentin copying you. We should hear back soon.

wadhwamatic:
@reed1 - sorry I wasn't following this very closely. I see an issue with how we've implemented this. It seems we're expecting that there can only be 1 source for chart data for a deployment. I think we need to keep the option to configure charts / layers which can get data from additional sources, not just the one set in constants.ts.

Right now, we're seeing the issue that 'rfb' is not yet available through the WFP Gateway but still exists through the previous HDC API. If we can allow for a default config which covers all charts but rfb, but still be able to configure rfb using the old API, that would be ideal. That should also allow us to have more than 1 chart data source in the future.

Let me know what you think, also @ericboucher

ericboucher:
@wadhwamatic this is how it's done yes, each layer can override the source to use a different URL

frontend/src/utils/admin-utils.ts
```diff
@@ -50,7 +51,10 @@ export const getChartAdminBoundaryParams = (
 ): AdminBoundaryParams => {
   const { serverLayerName, chartData } = layer;
 
-  const { levels, url, fields: datasetFields } = chartData!;
+  const { levels, url: chartUrl, fields: datasetFields } = chartData!;
+
+  // Take in chartUrl if provided, otherwise use default CHART_API_URL
+  const url = chartUrl || CHART_API_URL;
```
ericboucher, line 57: cc @wadhwamatic this is where we select the URL from layers.json, or default to CHART_API_URL otherwise

ericboucher:
@reed1 alsmot there :)
Please add typing for the API endpoint and as Amit suggested, revert to the old URL for`rfb` datasets for now.

reed1:
@ericboucher Or maybe route rfb in the backend to 'api.earthobservation.vam.wfp.org' ? That way if the HDC is ready then only backend needs to be updated, no need to rebuild frontend instances. The only custom URL for now is in Ukraine rainfall_dekad: https://dev.earthobservation.vam.wfp.org/prism/ukr.

ericboucher:
@reed1 I'd rather make this change in the frontend but I'll let @wadhwamatic decide if it's too cumbersome.

wadhwamatic:
> @reed1 I'd rather make this change in the frontend but I'll let @wadhwamatic decide if it's too cumbersome.

@ericboucher @reed1 - yes, I'd also prefer we handle this as changes to the frontend

reed1:
Sure, configs changed for blended charts:

rfb_blended_moz_dekad
rfq_blended_moz_dekad
rfb_blended_zwe_dekad
rfq_blended_zwe_dekad

api/app/main.py
ericboucher: @reed1 can you take care of this so that we can merge? Thanks.

ericboucher: You need to type the request, like the one for kobo/dates. This will allow the API to parse the request and let the user know what parameters are missing, if any. While you're at it, maybe you can type the ACLED request as well? Let me know if you run into any issues

```
@app.get("/kobo/dates")
def get_kobo_form_dates(
    koboUrl: HttpUrl,
    formId: str,
    datetimeField: str,
    filters: Optional[str] = None,
):
```

reed1: @ericboucher , updated, let me know if that works for you. I tried to use Depends to avoid duplicating function signature. It works on POST, but not GET, only on recent version of fastapi it will work on GET.

ericboucher (approved):

ericboucher:
I am redeploying the latest version of the API so we can test

api/app/main.py
```diff
@@ -161,19 +163,49 @@ def stats(stats_model: StatsModel) -> list[dict[str, Any]]:
...
     response = get(acled_url, params=params.dict())
     response.raise_for_status()
 
-    return Response(content=response.content, media_type="application/json")
+    return JSONResponse(content=response.content)
+
+
+@app.get("/hdc")
+def wrap_get_hdc_stats(
+    level: str, admin_id: str, coverage: str, vam: str, start: str, end: str
```
ericboucher, line 197: non of the params are optional?

api/app/main.py
```diff
@@ -161,19 +163,49 @@ def stats(stats_model: StatsModel) -> list[dict[str, Any]]:
...
+    iso: int,
+    limit: int,
+    fields: Optional[str],
+    email: str,
+    key: str,
+    event_date: Optional[date],
+):
     acled_url = "https://api.acleddata.com/acled/read"
 
     try:
```
ericboucher, line 176: I don't think you need this try/except anymore since the typing you introduced should take care of this now.

wadhwamatic (approved):
looks good from my side too - thanks @reed1 !

PR #861, opened 2023-06-06
Improve overall error handling and add request timeout; COUNTRY=cambodia

HarryMytilinaios:
- Generic error handling
- Error handling and dispatching errors
- Added try catch boundaries to every fetch async function
- Created fetchWithTimeout encapsulated function

This branch  fixes the overall error handling of all the requests we are doing in the frontend. Points that we need to observe here are the following:

I added a generic encapsulated method `fetchWithTimeout` and replaced every fetch in the overall app with this function. This function encapsulates the browser's `fetch` function but aborts it in specific timeout.

Right now I have defaulted the timeout inside the function in 8 seconds. Of course I made this function to pass custom timeout in specific function calls and of course we can change the timeout in an `env` variable or a `config` variable the 8 seconds its just for testing purposes.

The other key note in this branch is the generic error handing which is located in the `error-utils.ts` file. This generic error handling function which is called `catchErrorAndDispatchNotification` does exactly that catch error and dispatches a notification with the passed error message and abort message.

If the request has timed out the abort message will be shown else the error message will be shown. By default, this function utilizes warning notifications but we can  easily pass the variant whenever we want if we decide that we need in some places error (red) or warning (yellow).

Also the functions are using the dispatch method of redux which is passed by the async thunks inside the individual async functions. By accepting this error handling here we may need to rethink of removing the generic error handling middleware in redux.

Cc @wadhwamatic @ericboucher I understand that there are a lot of changes. I tried throwing custom errors in every fetch that we are doing to test and it seemed pretty solid. I understand that we need some time to test a good amount of scenarios, but i think we are on a good track here. I will link the potential issues that this branch may solve. Also we need to think from where exactly we will pass the timeout. Also are there requests that they may need different timeout than others? `Cambodia` is just for testing purposes

I added an error in charts by default to demonstrate the behavior.


frontend/src/context/datasetStateSlice.ts
```diff
@@ -188,35 +194,57 @@ const fetchHDC = async (
...
+            ? responseJson.data[field.key][index]
+            : field.fallback,
+        }),
+        {},
+      );
+
+      return { date, values };
+    });
+  } catch {
+    const error = new Error('Impossible to get HDC data.');
```
ericboucher, line 221: Impossible to get chart data from HDC.

frontend/src/components/MapView/Layers/raster-utils.ts
```diff
@@ -218,40 +217,34 @@ export async function downloadGeotiff(
...
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+      },
+      // body data type must match "Content-Type" header
+      body: JSON.stringify(body),
+    });
+    if (response.status !== 200) {
+      throw new Error(
+        `The raster layer ${collection} could not be generated. Please try your download again later.`,
```
ericboucher, line 232: I feel like this could all happen within `fetchWithTimeout` by adding a new parameter:
`errorMessage`

frontend/src/components/MapView/AlertForm/index.tsx
```diff
@@ -166,7 +166,7 @@ function AlertForm({ classes, isOpen, setOpen }: AlertFormProps) {
     };
 
     setAlertWaiting(true);
-    const response = await fetchApiData(ALERT_API_URL, request);
+    const response = await fetchApiData(ALERT_API_URL, request, dispatch);
```
jbperidypathtech, line 169: I found not friendly to pass the "dispatch" in all functions that need to call fetchWithTimeout function.

Do you have considered something like this?
> const response = await withTimeout(fetchApiData(ALERT_API_URL, request), dispatch, errorMessage);

In this cas withTimeout is a wrapper that throw an error when timeout is out and dispatch the error message. You don't have to modify fetchApiData function.

HarryMytilinaios:
Added some additional error handling and fixed the `fetchWithTimeout` function as follows:

- We have 2 different timeouts the `DEFAULT_REQUEST_TIMEOUT=15000 as 15 seconds` which is for all the available requests we are doing in the frontent and the `ANALYSIS_REQUEST_TIMEOUT = 60000 as 60 seconds or 1 minute` which is for the analysis endpoint `fetchApiData`. Those values are used in the `fetchWithTimeout` function and are configurable as we can pass custom timeouts for specific requests if we needed.

- `fetchWithTimeout` now handles all the `network` errors and `abort` errors by dispatching the necessary error messages (I also added the path in the error messages if it isn't user friendly we can definitely remove it). Moreover we can pass an optional errorMessage for the network error from the function that is calling `fetchWithTimeout`.

- Some try catch blocks have been added to some higher order functions in order to catch higher order throws, which are happened before or after the fetch requests.

- To handle those errors a custom `LocalError` class has been created which extends the generic `Error` class. The `LocalError` class is the one that is thrown throughout the higher order functions, in order to understand the `instanceof` the Error and dispatch the correct messages. I think its good if we throw the `LocalError` class from now on if an error is not a `Network` or `Abort` error.

- `ChartsSection` component has been refactored a bit. In order to facilitate the loading states and the error states, which was not handling before. Also linked to this issue https://github.com/WFP-VAM/prism-app/issues/863. Better check in `myanmar`, because the charts there are failing if we enable only admin1 or admin1 and admin2. 

Cc @ericboucher @wadhwamatic @jbperidypathtech @JulienPathtech 

I would appreciate your thoughts on this and fixes also of course.

frontend/src/context/datasetStateSlice.ts
```diff
@@ -214,9 +221,22 @@ const fetchHDC = async (
   });
 };
 
+// HDC API expects a parameter which depends on the layer
+// rainfall = rfh; ndvi = vim; blended = rfb
+const getVamParam = (serverLayerName: string): string => {
```
ericboucher, line 226: curious, why change from a switch to this?

ericboucher: you could simply have done the switch on `serverLayerName`

frontend/src/utils/fetch-with-timeout.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const DEFAULT_REQUEST_TIMEOUT = 15000;
+
+export const fetchWithTimeout = async (
+  resource: RequestInfo,
+  dispatch: Dispatch,
+  options?: FetchWithTimeoutOptions,
+  fetchErrorMessage?: string,
+): Promise<Response> => {
+  // The abort controller
```
ericboucher, line 18: unnecessary

frontend/src/components/MapView/AlertForm/index.tsx
ericboucher: I agree, is there anything we could do to make this a bit more dev-friendly?

frontend/src/context/datasetStateSlice.ts
HarryMytilinaios: The `getVamParam` function was there before but was initializing inside `loadAdminBoundaryDataset` function so I moved it outside and also the switch case was initializing with `switch(true)`, see below function. I found it for this specific occasion to be more clear using if statements. If we are not ok with it I can convert it back to where it was.

HarryMytilinaios:
The problem  with `dispatch` is that we can take it from 2 different places. The first place is through the component with the `useDispatch` hook, and the other place is through an asyncThunk. The problem is that there are like 1 or 2 levels of functions with data manipulation to arrive at the actual asynchronous fetch. If we want abstraction and a generic error hanndling for the `Network` and `Abort` errors I don't see how can we achieve it. Moreover there are some fuctions that are returning initial or empty data in the catch block sometimes. The way I see it, if we want to now pass the dispatch in every function, we can throw the errors everywhere  and let the middleware of the redux handle it, for the requests that are inside async thunks and for the other requests that are inside of a component directly to manually try catch dispatching the error there. With this approach we don't need to pass `dispatch` function anywhere else than the actual `asyncthunk` or in the component, which we already have access through the `useDispatch` hook. How does that sound @ericboucher @jbperidypathtech . The only problem is that I am not sure how to handle the initial or empty data of some functions that are returning them on catch block.

frontend/src/utils/fetch-with-timeout.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const ANALYSIS_REQUEST_TIMEOUT = 60000;
+
+const DEFAULT_REQUEST_TIMEOUT = 15000;
+
+export const fetchWithTimeout = async (
+  resource: RequestInfo,
+  dispatch: Dispatch,
+  options?: FetchWithTimeoutOptions,
+  fetchErrorMessage?: string,
```
JulienPathtech, line 16: Why covering the real error message with a custom one ?
I would have just propagate the real error message of the HTTP response.

ericboucher: The underlying error is not always very telling to non-developers / users

HarryMytilinaios: Because the HTTP error message most of the times is not very informative friendly so what I did is to give the option to pass a better errorMessage than the HTTP's one. Which will involve the url that we are hitting and is going to give some more infformation about the actual problem. Also if we have a generic error handler there for network and abort messages, will give us a better handling to handle all other local errors in  higher order functions, when we throw them.

ericboucher: @HarryMytilinaios maybe we could still log the actual error in the console? wdyt?

HarryMytilinaios: of course

PR #866, opened 2023-06-09
Changed the comparison of dates in layerDateSelector and queryDate for different timezones; COUNTRY=srilanka

HarryMytilinaios:
This change hopefully will fix the https://github.com/WFP-VAM/prism-app/issues/865. Unfortunately this bug was not reproducable from localhost.

### DIFF
--- a/frontend/src/context/mapStateSlice/selectors.ts
+++ b/frontend/src/context/mapStateSlice/selectors.ts
@@ -24,7 +24,11 @@ export const layerDataSelector = (id: LayerKey, date?: number) => (
 ): LayerDataTypes | undefined =>
   state.mapState.layersData.find(
     ({ layer, date: dataDate }) =>
-      layer.id === id && (!date || date === dataDate),
+      layer.id === id &&
+      (!date ||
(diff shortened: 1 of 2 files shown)


wadhwamatic:
Thanks a lot for the quick turnaround on this @HarryMytilinaios and @ericboucher. This has addressed the issue as far as I can tell. I'll get more feedback from the team in Sri Lanka, but I think we should be good based on the initial response I heard from them and what I was able to test locally

ericboucher:
@HarryMytilinaios a few things to do:
- let's find a cleaner way to do this if there is one
- then move it to a utility function
- make sure we don't need it in other places

In addition, but don't spend too much time on this, let's investigate quickly why a 30min difference appears in our dates when we are on the IST timezone...

HarryMytilinaios:
Created a utility function which is comparing the equality  of dates without considering the time, and replaced it whenever we do date comparison. Also I took the initiative on removing the console.logs @ericboucher if you think we need them I can put them back.

ericboucher (approved):

JulienPathtech (approved):

Issue #794, opened 2023-04-07
[Feature Request]: Make the 'tooltip' more configurable to include more text in an easier to read format


wadhwamatic:
### Provide a clear and concise description of what the problem is.

The tooltip feature allows multiple datapoints and text to be added to it through layer configuration, but the format of the popup is limiting

### Provide a clear and concise description of what you want to happen.

We need design options that will allow for more text to be displayed in an easy to read format. 

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

For example, it would be useful to see much more in-depth information from a Kobo form submission when a user clicks on the map. Some users have also requested narrative text to be visible in the popup. I will link to some examples of text we'd want to include.

wadhwamatic:
Hi @JulienPathtech. 

One current example is from [ACLED](https://acleddata.com/) reports which have been integrated in Myanmar. The report includes a couple of sentences of text that a user has to scroll within a small window to see as of now. See this layer as an example:
https://prism-mmr-demo.surge.sh/?hazardLayerIds=conflict_fatalities

<img width="481" alt="Screenshot 2023-05-24 at 16 01 18" src="https://github.com/WFP-VAM/prism-app/assets/3343536/d9345c7e-2855-4949-956c-ffa92b19cf89">

Text:
> Number of fatalities in an armed conflict incident: 2
> Date: 2023-05-19
> Event type: Explosions/Remote violence
> Sub event type: Air/drone strike
> Source: Myanmar Pressphoto Agency
> Details: On 19 May 2023, in In Pin village (Pakokku township, Pakokku district, Magway region), the Earthquake People's Defense Force attacked military soldiers stationed in the village with Energa bombs attached to a drone. Two military soldiers were killed.

A second example is for our regional deployment in Dakar (RBD) where the Cadre Harmonise data layers show multiple statistics in the tooltip.

https://prism-rbd.surge.sh/?baselineLayerId=ch_phase

<img width="520" alt="Screenshot 2023-05-24 at 15 57 43" src="https://github.com/WFP-VAM/prism-app/assets/3343536/e303ffea-355a-4acf-8ab5-550dfb01dd3e">

Text: 
> Population in phase 1: 62683
> Population in phase 2: 125063
> Population in phase 3: 102655
> Population in phase 4: 81256
> Population in phase 5: 1817
> Population in phase 3 to 5: 185728
> Reference period start: 2022-10-01
> Reference period end: 2023-01-31

[...]

PR #867, opened 2023-06-12
Add a table for phase data in tooltips; COUNTRY=cameroon

ericboucher:
This fixes issue #794 

This PR:
- updated the size of the tooltip (increased max height) and changed fonts
- allowed it to be below the point if it fits better on the screen
- allowed users to hide the data title
- creates a table view for Phase Classification data

Screenshot 2023-06-12 at 6 15 20 PM

Screenshot 2023-07-20 at 5 55 22 PM


wadhwamatic:
@ericboucher - would Marc have some time to give us some design options? The two current use cases are mentioned in https://github.com/WFP-VAM/prism-app/issues/794. There could be a use case for the decision maker view too.

ericboucher:
Screenshot 2023-07-14 at 4 07 58 PM

jbperidypathtech (approved):
I approve because it is ok for me after last comment accounted.

frontend/src/components/MapView/MapTooltip/index.tsx
```diff
@@ -6,15 +6,86 @@ import {
...
+import { PopupData, tooltipSelector } from 'context/tooltipStateSlice';
 import { isEnglishLanguageSelected, useSafeTranslation } from 'i18n';
+import { TFunction } from 'utils/data-utils';
+import { ClassNameMap } from '@material-ui/styles';
+
+// This function prepares phasePopulationTable for rendering and is specific
+// to the data structure of the phase classification layer.
+// Note - this is a bit hacky for now and will likely need to be revamped if we
+// encounter other complex needs for tooltips.
+const generatePhasePopulationTable = (
```
JulienPathtech, line 21: We could add `JSX.Element` as return value for this method ? We could also make a separate component file but it depends on the revamp you have in mind

PR #869, opened 2023-06-14
Adding fill patterns to admin level data layers; COUNTRY=rbd

HarryMytilinaios:


frontend/src/context/layers/admin_level_data.ts
```diff
@@ -58,14 +58,18 @@ export async function getAdminLevelDataLayerData({
...
+  let adminBoundariesLayer = layerDataSelector(adminBoundaryLayer.id)(
     getState(),
   ) as LayerData<BoundaryLayerProps> | undefined;
   // TEMP - add a 15s wait time to load admin boundaries which are very large
   // WARNING - This is a hack and should be replaced by a better handling of admin boundaries.
   // TODO - make sure we only run this once.
   if (!adminBoundariesLayer || !adminBoundariesLayer.data) {
     await new Promise(resolve => setTimeout(resolve, 15000));
+    // eslint-disable-next-line no-param-reassign
+    adminBoundariesLayer = layerDataSelector(adminBoundaryLayer.id)(
```
ericboucher, line 70: @HarryMytilinaios I put this quick hack together to avoid the "boundary file not found" error but maybe there is a better way to do this? With a useEffect maybe or another way to re-trigger the layerDataSelector above?

ericboucher:
@HarryMytilinaios great work! Let's silence the console error for re-upload of the svgs and then I think we should be good to go

frontend/src/context/layers/admin_level_data.ts
HarryMytilinaios: @ericboucher can you provide me some information when that error was happening, so that maybe we can find a better approach on that?

HarryMytilinaios:
Adding some additional information here for testing and how the whole `fill-patterns` feature actually works. This is tested in `rbd` but can be easily added to other countries as well Cc @wadhwamatic. 

- We added a `fill_pattern` optional attribute in the specific admin level data layers that we want to observe together. The properties that the `fill_pattern` accepts is `left` or `right`. Those special values are displaying diagonal lines svg patterns based on their direction `left` displays left diagonal lines (from left to right) and `right` displays right diagonal lines (from right to left).

- The color of the patterns applied is based on the legend colors.

- Because Mapbox did not accept svg's as patterns we need to create base64 encoded png's,  from the svg's with their color based on the legend. Additional utility functions have been created to make that convertion. Right now we are exposing 2 utility functions for this current implementation and possible future implementations. The first is the `convertSvgToPngBase64Image` which converts an svg to a base64 encoded png image and the `convertSvgToPngBaseBlobImage` which converts an svg to a blob image (maybe for future implementations).

- Right now we accept multiple admin level data layers to overlap only if they have different `fill_pattern` attibute in their `layers.json`. So for example we may have a layer that does not have at all `fill_pattern` and one that has `right` or `left` those 2 are going to overal together. We may have 3 layers that one does not have at all `fill_pattern` on has `right` and one has `left`, then all those 3 are going to overlap together. Of course this logic can be tweaked for out needs.

- For our testing purposes we added in some admin level data a `fill_pattern` already of course this can change according to the requirements.

frontend/src/context/layers/admin_level_data.ts
ericboucher: the error was happening because when we first call the layerDataSelector, `adminBoundariesLayer` might be undefined because the boundary takes times to load.

Then we wait a bit for the boundary layer to load. But since we didn't reassign the variable, the code still thought that `adminBoundariesLayer` was undefined.

So I think what I did works ok, but maybe a useEffect might make more sense to have the component reload from the top and re-get the value for `adminBoundariesLayer`. 🤷

frontend/src/utils/image-utils.ts
```diff
@@ -0,0 +1,190 @@
...
+    default:
```
ericboucher, line 186: You don't need to duplicate the left shape. simply have case "right" and default to left.

frontend/src/utils/image-utils.ts
```diff
@@ -0,0 +1,190 @@
...
+    case 'right':
+      // Right Diagonal lines svg (comes to left)
```
ericboucher, line 184: you can use `/////` and `\\\\\` which is more telling than `comes to right` or `comes to left`

ericboucher (approved):

ericboucher (approved):

wadhwamatic:
Great work, thanks @HarryMytilinaios. This works as expected and looks good. One request - can we add the fill style to the legend as well?

HarryMytilinaios:
@wadhwamatic I am not exactly sure I am following here. So we want to show the svgs in the legend also when a layer with fill pattern is applied? Also another thing that I'd like to ask. When an AdminLevelDataLayer is not overlapping with another one and has fillPattern should we show the fillPattern every time, or should we show the pattern only if it is overlapping with another one or else to show the default coloring behavior?

ericboucher:
@HarryMytilinaios yes, either show the svg in the legend or add a note in the legend saying which pattern is being used.
I'm in favor of keeping the layer consistent and always use the pattern, otherwise it might get a bit confusing. But I'll let @wadhwamatic decide.

wadhwamatic:
> @HarryMytilinaios yes, either show the svg in the legend or add a note in the legend saying which pattern is being used. I'm in favor of keeping the layer consistent and always use the pattern, otherwise it might get a bit confusing. But I'll let @wadhwamatic decide.

@HarryMytilinaios - I'm in full agreement with Eric on both of his inputs. Thanks!

HarryMytilinaios:
@ericboucher @wadhwamatic I tried adding  the images of `fill-patterns` in the legend component, but unfortunately, they are not looking so good, because the size is too small a user's naked eye cannot identify that those images are indeed diagonal lines, see image below:
Screenshot from 2023-06-26 11-12-59
I tried to make it clearer but then the icons are getting too big in order to identify the pattern. See image below:
Screenshot from 2023-06-26 11-13-32
So instead I added a notification below the legend items `Fill pattern is used`, if we need to change the text or anything please let me know. See image below:
Screenshot from 2023-06-26 11-20-40

ericboucher:
@HarryMytilinaios maybe we could use a different svg with bigger but fewer lines?
Otherwise text is ok, let's indicate "right" or "left" though.

HarryMytilinaios:
@ericboucher the lines in the svg are continued in a patternand changed with the scaling, therefore it is not so easy to manipulate them at the current state, it would require additional creation of another svg that will take place in the Legends.

wadhwamatic:
> @ericboucher the lines in the svg are continued in a patternand changed with the scaling, therefore it is not so easy to manipulate them at the current state, it would require additional creation of another svg that will take place in the Legends.

@HarryMytilinaios - we should definitely aim to display a simplified version of the fill pattern in the legend. This is what you'd see in cartography as a standard practice. I'd suggest we go ahead and create an additional svg for the legend to accomplish that. Hope it's relatively straightforward.

HarryMytilinaios:
@wadhwamatic @ericboucher I added the fill pattern in legend items and seems to be working pretty well, I had to increase a little bit the width and height of the legend colorIndicators in order for it to be clear in a naked eye, but I think this is pretty solid, wdyt?

ericboucher:
CSS does the trick, nice work. I resized to the smaller version as I think it's enough and otherwise it makes the legend too bulky.

frontend/src/components/MapView/Legends/ColorIndicator/index.tsx
```diff
@@ -12,22 +18,34 @@ const useStyles = makeStyles(() =>
...
-        color: string;
-        opacity: number;
-      }) => fade(color, opacity),
+      backgroundColor: ({ color, opacity }: StylesProps) =>
+        fade(color, opacity),
+    },
+    fillPatternIndicator: {
+      height: 10,
+      width: 10,
+      marginRight: 4,
```
ericboucher, line 27: since these are shared maybe this could be extracted into a common indicator class?
and the `fillPatternIndicator` overrides just what's necessary?

HarryMytilinaios: Done

Issue #751, opened 2023-03-10
[Feature Request]: Add start / end dates to admin level data layers


wadhwamatic:
### Provide a clear and concise description of what the problem is.

We can add validity days to a data layer, which works well when data rounds are fixed by number of days. But in some cases, the data rounds do not follow this pattern.

### Provide a clear and concise description of what you want to happen.

Allow layer configuration to have a start date and end date attribute. It should read the value from the data file directly. For example, layers.json, I want to be able to add something like:
> "data_field": "ch_phase_class",
> "start_date_field": "start_date",
> "end_date_field": "end_date"

With this approach, each dataset would need to have start and end dates in every round of data.

Or we could do something similar to the approach used with multiple dates:
> "path": "data/rbd/ch_round_{x}.json",  
> "period_dates": {
>     ["1", "2022-01-01", "2022-02-28"],
>     ["2", "2022-03-01", "2022-05-31"],
>     ["3", "2022-05-31", "2022-08-31"],
>     ["4", "2022-09-01", "2022-08-31]"
> ]

In this case, the dataset wouldn't need dates but we'd need to update layers.json each time new data is added.

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

In terms of functionality, this should work the same way as validity periods. In the timeline, any date between the start and end date should show valid data for that layer.

wadhwamatic:
@ericboucher - the datasets used in the RBD CH files have start and end date. For example, see https://github.com/WFP-VAM/prism-app/blob/master/frontend/public/data/rbd/ch/admin1_2023_06_01.json. The fields 'start_date' and 'end_date' are available.

If we can add start/end date as configuration options in layers.json, this would be a good approach from my perspective

JulienPathtech:
@wadhwamatic I have some questions to check that I'm going in the right direction :
- Does it means that indicating a start and end date in a layer imply that there are datas available on every days in between ? 
- What happens if `dates` or `validity` properties are specified ? What properties should I use first ? (Or maybe we should not handle that case)

JulienPathtech:
You can ignore my previous comment. I had a clarification call with eric.

PR #877, opened 2023-06-16
Get start and end date for validity period from admin level data; COUNTRY=rbd

JulienPathtech:
This fixes issue #751.

How to test the feature:

1. Use RBD country with phase layer.
2. Timeline should load with a direction for each start and end date for every periode range 

Notes:
- every `admin_level_data` layer type having a `path` also have `start_date` and `end_date` fields in there corresponding data files. 
- it's always  `start_date` and `end_date`, hence it does not need to be configurable

**TODO**
- [x] POC on phase classification for rbd 
- [x] Ensure working only when a path is defined for the layer and start_date end_date is defined
- [x] Remove unused validity object for layer that now use start_date end_date feature 
- [x] Safe http request on data ressources (~~timeout~~, error, ...)
- [x] Draft another ticket for perf improvements => https://github.com/WFP-VAM/prism-app/issues/885
- [x] Draft another ticket to generate DateItem only for visible layers and on the fly instead of all layers at application first load => https://github.com/WFP-VAM/prism-app/issues/888

Screenshot/video of feature:

Before (RBD, Population in phase 3 to 5)
image

Now :
image

With rainfall
image

Test URL :  https://prism-877.surge.sh/?baselineLayerId=ch_pop3_5&date=2023-06-01&hazardLayerIds=rainfall_dekad


frontend/src/utils/server-utils.ts
```diff
@@ -229,14 +239,49 @@ const createDefaultDateItem = (date: number, validity?: Validity): DateItem => {
...
+  layerPathTemplate: string,
+) {
+  const ranges: StartEndDate[] = await Promise.all(
+    layerDates.map(async r => {
+      const path = layerPathTemplate.replace(/{.*?}/g, match => {
+        const format = match.slice(1, -1);
+        return moment(r).format(format);
+      });
+
+      const res = await fetch(path);
```
HarryMytilinaios, line 256: Wouldn't it be better to use the `fetchWithTimeout` here?

JulienPathtech: I wanted to and I also wanted 2 distincts error messages depending on the error.

frontend/src/utils/date-utils.test.ts
```diff
@@ -0,0 +1,49 @@
+import { generateDateItemsBetweenForRanges } from './date-utils';
+
+describe('Test buildDateItemsFromStartEndDates', () => {
+  test('should return empty', () => {
+    // Act
```
ericboucher, line 5: I'm not sure these Act / Arrange comments are very telling. Let's find more descriptive comments or get rid of them.

frontend/src/utils/date-utils.ts
```diff
@@ -1,6 +1,53 @@
...
+  startDate: Moment,
+  endDate: Moment,
+): number[] => {
+  return Array.from(
+    { length: endDate.diff(startDate, 'days') + 1 },
+    (_, index) => startDate.clone().add(index, 'days').valueOf(),
+  );
+};
+
+export const generateDateItemsBetweenForRanges = (
```
ericboucher, line 26: How about renaming these two simply:

`generateDatesRange`

`generateDateItemsRange`

ericboucher, line 26: How about renaming these two simply:

`generateDatesRange`

`generateDateItemsRange`

frontend/src/utils/server-utils.ts
```diff
@@ -411,17 +458,61 @@ export async function getLayersAvailableDates(
...
-      : dates.map((d: number) => createDefaultDateItem(d));
+        if (matchingValidityLayer) {
+          return {
+            [layerDatesEnntry[0]]: generateIntermediateDateItemFromValidity(
+              matchingValidityLayer,
+            ),
+          };
+        }
+
+        // Generate date for layers with path
```
ericboucher, line 491: // Generate date`S`?

frontend/README.md
```diff
@@ -0,0 +1,7 @@
+# PRISM React App
+
+## Run a single test 
+
+```
+npm run test -- -t 'Test name'
```
ericboucher, line 6: I am not sure this is needed. But if you think it's useful, we use yarn in this repo so let's use that as the pm. :)

JulienPathtech:
Hello @wadhwamatic .
There is a known issue on this PR with Nigeria as its the only country that have data with validity period and without a `start_date` and `end_date` proprety.
It has a `reference_date`

Should we keep using `validity` object or use `reference_date` to generate available dates ?

frontend/src/utils/date-utils.test.ts
JulienPathtech: It's just a habit of separating unit test in 3 steps.
// Arrange // Act // Assert.
I always found it usefull to know each part at a glance.
But as you wish

frontend/README.md
JulienPathtech: I will remove it

wadhwamatic:
> Hello @wadhwamatic . There is a known issue on this PR with Nigeria as it's the only country that have a validity period and without a `start_date` and `end_date` data proprety. It has a `reference_date`
> 
> Should we keep using `validity` object or use `reference_date` to generate available dates ?

@JulienPathtech - can you point me to the specific data file which has the issue? There must have been an error in generating the start/end dates that I'll fix.

JulienPathtech:
Hello @wadhwamatic, every data files inside` /public/data/nigeria/ch` actually.

wadhwamatic:
@JulienPathtech - I fixed the configuration and data files for Nigeria. But, I would just focus on RBD as Nigeria was just a demo version and not being used.

For both RBD and Nigeria, the CH data doesn't load for me locally. I also get an error on the deployed version.
Screenshot 2023-07-05 at 13 25 21

ericboucher:
@JulienPathtech i pushed a fix for timezone issues. Let me know if it looks ok for you. It seemed to work for Amit but let's QA a bit more.

HarryMytilinaios:
merged master and added the functionality on removing the layers without available dates that was already in master. Quick addition I found in the console the below errors, I don't know if this is something we already know just wanted to point it out. 

Screenshot from 2023-07-07 13-37-30

frontend/src/utils/date-utils.test.ts
```diff
@@ -0,0 +1,49 @@
+import { generateDateItemsRange } from './date-utils';
+
+describe('Test buildDateItemsFromStartEndDates', () => {
+  test('should return empty', () => {
+    // Act
+    expect(generateDateItemsRange([])).toEqual([]);
+  });
+
+  test('should return and array of DateItems for one range of 5 days', () => {
```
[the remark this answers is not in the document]
HarryMytilinaios: Done

JulienPathtech:
Thanks @ericboucher for fixing the bug. It seems ok for me as I don't reproduce Amit's bug.

@HarryMytilinaios I added this console error to point either a layer missconfiguration or missing data files.
For instance, we specified "2021-02-01" in the layer "Urban areas phase classification" configuration but the data file is not present.

It also means that we currently says we have data for a layer in the timeline that we actually don't have which lead likely error for users.

JulienPathtech:
> Should we remove the dates from the available dates at this time then?

Yes, maybe I should leave it to Amit ?
So he can add dataset when missing and remove configuration when needed ?

@wadhwamatic I added a validityPeriod configuration that contains start and end date field.
I added it whenever I found a layer with bot a `dates` and `path` properties but it would be great if you double check on that.

wadhwamatic:
Thanks @JulienPathtech for the changes on this. I see the new config for validityPeriod start_date_field and end_date_field. That's perfect. In that case, I believe you're asking if we can then remove the date array. Yes, it should be removed from the config so as to not cause confusion. I'll make note of this in the readme file as part of this feature.

JulienPathtech:
> I was thinking that we would log an error in the console like you're doing but then remove the dates programatically so as to avoid confusion for the users. Not a priority but nice to have if it's easy to implement

I agree with you, unavailable dates should be removed from the date array. I can remove missing dates from the array myself but I wanted Amit to do it so he can check.

I think there will be no confusion for the end user since date range will not be generated for missing dates. It's more for us to have a clean configuration that reflects actual data files.

@wadhwamatic I'm not sure to understand. `dates` array is still used to find data files. (by replacing template `path` property for every date of the dates array). So we cannot just simply remove it from every layer :thinking: 

It's more about this scenario :

> @HarryMytilinaios I added this console error to point either a layer missconfiguration or missing data files.
> For instance, we specified "2021-02-01" in the layer "Urban areas phase classification" configuration but the data file is not present.

ericboucher (approved):

PR #880, opened 2023-06-19
476 analysis comparison; COUNTRY=myanmar

jbperidypathtech:
This fixes issue 476.

How to test the feature:

- Use Myanmar
- Choose Hazard layer "Flood extend", you should be able to:
- Select Exposure option in statistic
- When selected, two new inputs appear to fill the operator and the associated value

Notice: I tried to improve the responsivity of the app when we change the statistic (to max to min for example) because values are available in the front. But there is a lot of thing to change and I am not sure of the impacts. So I think it should be great to create new ticket to globally improve the left menu UX to auto-recalculate statistic table and map when any configuration value in the left menu change.

Screenshot/video of feature:
Capture d’écran du 2023-06-23 10-28-07


frontend/src/utils/analysis-utils.ts
```diff
@@ -111,6 +111,7 @@ const operations = {
     const ceil = sortedValues.length / 2;
     return (sortedValues[floor] + sortedValues[ceil]) / 2;
   },
+  intersect_percentage: (data: number[]) => mean(data),
```
jbperidypathtech, line 114: I am not sure about calculation required for intersect_precentage. Let me know if we need to change it.

ericboucher: Hmm I think we need to error our in this case or at least raise a warning. With just the percentage we can't give an accurate number here from the frontend. But this shouldn't be triggered atm as the analysis is run by the backend, no?

jbperidypathtech: I think I don't really understand what is the usage of this operator. I can see it is used by the loadFeaturesClientSide function that is called by the fetchImpactLayerData. I was thinking it was used to display the layer with the selected statistic but it seems not. If I change with "intersect_percentage: () => null" it continues working fine. I change this line because of typescript but I don't know the usage/impact and what to do with it.

wadhwamatic:
I see this is WIP, but wanted to mention there's an error in the calculation where an entire area is listed as flooded (100%) which is clearly not the case.  

Screenshot 2023-06-27 at 10 47 22

jbperidypathtech:
Hi @wadhwamatic,
Thanks for the test. I just checked and the problem seems to come from the API.

When we query the stats with: intersect_comparison | ">=3"

We get in response for Thegon TS_PCODE: MMR008005
```
    {
        "TS_PCODE": "MMR008005",
        "stats_min": 3.0,
        "stats_max": 3.0,
        "stats_mean": 3.0,
        "stats_sum": 33.0,
        "stats_std": 0.0,
        "stats_median": 3.0,
        "stats_intersect_percentage": 1.0
    },
```

ericboucher:
Yes this is an edge case with the API.

That particular grid is mostly empty and the API returns the percentage of pixels with the correct value. But it cannot count empty pixels. So what it means is that 100% of the grid pixels available in this area have a value of 3.

The API should work well with normal rasters already. I'll look into a way to count/add empty pixels as well.

frontend/src/utils/analysis-utils.ts
ericboucher: So if I recall correctly this is used to make the analysis calculations locally, if you don't have access to a stats API.
It is less relevant now but at the beginning, some deployments wanted to be "independent" from the API.

jbperidypathtech: ok understood ! I change to raise an error if trying to calculate intersect_percentage from client side then

ericboucher:
The API edge case should be resolved by #893

wadhwamatic:
@ericboucher - I tried to test this locally for Cambodia and through the surge PR deployment for Myanmar. I'm still seeing unexpected results.

Running an analysis through the analysis tab shows every admin area in the same classification as we're not properly handling percentages (everything is 0)
Screenshot 2023-07-12 at 10 45 54

If I run an exposure analysis and sort the resulting table by sum descending, I see the two highest areas have just a few pixels shown as flooded. I don't think it's possible that those pixels equate to 5000+ people
Screenshot 2023-07-12 at 10 48 45

jbperidypathtech:
Hi @wadhwamatic @ericboucher,

This is what I could do:
Capture d’écran du 2023-07-13 11-14-56

Several questions:
- I am not sure about expected words, please let me know what to change
- I am not sure about the units
  - intersect_percentage: seems to be between 0 and 1 then I transformed in percent in the frontend
  - intersect_area: values from API are > 100 is it normal? Then I didn't make any transformation of the API value in the frontend

As the frontend was not thinking to manage several statistics, I thought the code not very clean :(.

ericboucher:
@wadhwamatic I believe you might not have been looking at the latest version of the PR.
Here is what I get for Cambodia:

Screenshot 2023-07-13 at 4 15 58 PM

Also, note that this PR is not touching the `"exposure analysis"` at all. Most likely we have a scaling issue with world pop data here as well, similar to what we were seeing for Myanmar.

wadhwamatic:
> I am not sure about the units
> intersect_percentage: seems to be between 0 and 1 then I transformed in percent in the frontend
> intersect_area: values from API are > 100 is it normal? Then I didn't make any transformation of the API value in the frontend

@jbperidypathtech - yes, intersect_percentage is correct after transforming it, thanks.  The unit for intersect_area should be square kilometers

Let's use these table headers: 
- Percent of area exposed
- Area exposed in sq km

wadhwamatic:
> @wadhwamatic I believe you might not have been looking at the latest version of the PR.

@ericboucher - Yes, it seems I didn't have the latest. I see the same results as you now.

> Also, note that this PR is not touching the `"exposure analysis"` at all. Most likely we have a scaling issue with world pop data here as well, similar to what we were seeing for Myanmar.

Ok, I opened an issue to track this. https://github.com/WFP-VAM/prism-app/issues/917

frontend/src/components/MapView/Layers/AnalysisLayer/index.tsx
```diff
@@ -102,18 +103,35 @@ function AnalysisLayer({ before }: { before?: string }) {
...
+          );
           dispatch(
             addPopupData({
               [analysisData.getStatTitle(t)]: {
-                data: getRoundedData(
-                  get(evt.features[0], ['properties', statisticKey]),
+                data: `${getRoundedData(
+                  formattedProperties[statisticKey],
                   t,
                   precision,
```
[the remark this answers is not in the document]
jbperidypathtech: Right I changed and fix issue I added (the x100 conversion was applied also for others statistics :disappointed:)

wadhwamatic:
@jbperidypathtech - I just sync'd with @ericboucher on this and agreed that a couple of things should be added here:
1. We should filter out areas where 0% was exposed. They shouldn't be colored in the map, and don't need to appear in the table. Eric suggested we can handle this in the backend
3. When a user chooses 'area exposed' as the statistic in the analysis tab, it's not clear what the threshold parameter refers to. Let's add % to the right of the above input field when the statistic is 'area exposed'. See screenshot below
4. The user input on the threshold value is not scaled. So, right now I have to enter 0.01 to set the threshold to 1%. We should handle the scaling to avoid this

Screenshot 2023-07-19 at 11 16 57

api/app/zonal_stats.py
```diff
@@ -368,6 +376,48 @@ def intersect_percentage(masked) -> float:
...
+                )
+
+            clean_stats_properties = {
+                **clean_stats_properties,
+                f"{safe_prefix}intersect_percentage": intersect_percentage,
+            }
+
+        # merge properties back at the proper level
+        # and filter out properties that have "no" intersection
+        # by setting a limit at 0.005 (0.5%).
```
ericboucher, line 411: @wadhwamatic As it happened there were actually very few admin boundaries with `zero` value, but very small percentages. I set an arbitrary limit of 0.5%. Let me know what you think.

ericboucher (approved):

PR #881, opened 2023-06-20
Fix/layer ordering; COUNTRY=rbd

HarryMytilinaios:
This PR trying to fix layer paint ordering in map.


HarryMytilinaios:
@ericboucher Looks good overall I think we can merge this into [test-paterns-branch](https://github.com/WFP-VAM/prism-app/pull/869) wdyt?

Issue #787, opened 2023-04-06
[Feature Request]: Visualize start and end dates for layers with validity ranges


wadhwamatic:
### Provide a clear and concise description of what the problem is.

If a layer has a range of valid dates, the timeline UI shows a continuous line between the start and end dates. When there are consecutive date ranges available, it's impossible to know when a new data point exists.

### Provide a clear and concise description of what you want to happen.

Add markers to the timeline to show when a validity period starts and ends for any layer with date ranges. Since validity ranges can go 'forward' or 'backward', it would be useful to also show direction in the UI. 

In addition, the date picker (calendar) would ideally also give some indication of new ranges. We could use bold text on start/end dates. 

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

Here's a mockup for discussion. 

<img width="1183" alt="Screenshot 2023-04-06 at 10 53 30" src="https://user-images.githubusercontent.com/3343536/230458322-a6d7f44e-0412-4e11-a230-33f88952c284.png">

JulienPathtech:
Hello @wadhwamatic 
- I don't see any backward going validity dates ATM, is it a real use case ? 
- Would it be better to just pick a date range (i.e "2022-02-01 to 2022-06-01" ) instead of a particular date in the timeline ? (Most of the time it's a 100+ days forward)

wadhwamatic:
@JulienPathtech - Eric covered it well. I'd add that we haven't really implemented the ideal state for validity periods because of the issues he referenced. Once those are resolved, the next order of business is to have a clear indication to users of the range of dates that an indicator references. Let me know if you need more clarifications!

JulienPathtech:
I had a call with eric and I understood better, thanks @wadhwamatic  ! 
Is it ok for you if I just add an arrow on the start **or** end date with the direction of the validity ? (In other words, the arrow would display both direction and start or end date information)

wadhwamatic:
@JulienPathtech - great, glad you were able to connect with eric and discuss. Yes, we can use an arrow on just start or end date. I would propose we use start date and experiment a bit to see how clear it is for a user.

ericboucher:
@JulienPathtech it might be worth looking at https://github.com/WFP-VAM/prism-app/issues/751 at the same time or at least make sure what you're implementing here is compatible.

PR #882, opened 2023-06-20
Revamp styling of the timeline; COUNTRY=rbd

JulienPathtech:
This fixes issue #787.

This complete the PR https://github.com/WFP-VAM/prism-app/pull/852 by updating style according to sketch

How to test the feature:

- [x] Use rbd with 2 layers => https://prism-882.surge.sh/?baselineLayerId=ch_phase&date=2023-05-27&hazardLayerIds=rainfall_dekad
- [x] Use southsudan with 3 layers => http://localhost:3000/?hazardLayerIds=flood_events%2Crainfall_dekad%2Cconflict_fatalities&date=2022-12-11
- [x] Move cursor of the timeline
- [x] Select date from the date picker or the timeline

Screenshot/video of feature:
image

Figma designs:
image


JulienPathtech:
There is a known bug when we try to select any date on the Timeline. I'm on it :)

wadhwamatic:
> Hi @wadhwamatic, I kept the grey layer that represent available or selectable dates in the calendar. Should we keep that layer or not ?

@JulienPathtech - I don't see value in the grey layer anymore since the intersections are clearly illustrated by the individual layer's bars. Selectable dates in the calendar should still appear though and that is based on the intersection of selected layers. Hope this is clear!

wadhwamatic:
@JulienPathtech - great start on this! I took a quick look for now and noticed a bug when you toggle between English and French (click 'FR') in the top right

Minor graphical issue: I expected the rectangle in the timeline and the handle for the slider to move together. I see that the rectangle waits until the layers are loaded. Instead, perhaps the rectangle should disappear until the layers are loaded for the new date.

JulienPathtech:
**Update**
- Fix when selected a point in the Timeline (https://github.com/WFP-VAM/prism-app/pull/882#issuecomment-1599108150)
- Removed grey / intersection layer
- Make the whole cursor moving when selected (There is a drawback to this -> tooltip is not available arround the cursor : /. If I use a lower z-index on the cursor, it breaks its display). We can keep it or revert to previous version

I'm waiting your feedbacks

ericboucher:
Looks great. I think it's ok this way @JulienPathtech, it only blocks the tooltip and clicks for the day before and the day after, which I think is fine since there are two other ways to move to these dates already.

frontend/src/components/MapView/DateSelector/index.tsx
```diff
@@ -335,10 +325,11 @@ const DateSelector = memo(
           justify="center"
           className={classes.datePickerContainer}
         >
+          {/* Mobile */}
           <Grid item xs={12} sm={1} className={classes.datePickerGrid}>
             <Hidden smUp>
               <Button onClick={decrementDate}>
-                <ChevronLeft />
+                <ChevronLeft style={{ color: '#101010' }} />
```
ericboucher, line 332: It might be worth extracting some of these colors into the theme?

frontend/src/types/custom.d.ts
```diff
@@ -3,6 +3,8 @@
 // Allow svg imports in .tsx
 // https://stackoverflow.com/questions/44717164/unable-to-import-svg-files-in-typescript
 declare module '*.svg' {
+  import React = require('react');
+  export const ReactComponent: React.FC<React.SVGProps<SVGSVGElement>>;
```
ericboucher, line 7: are these needed?

ericboucher (approved):
Great work @JulienPathtech!
Let's take this opportunity to add a bit more comments around the timeline items maybe?

HarryMytilinaios:
@JulienPathtech great work this feels really nice I agree with the comments of @ericboucher regarding the colors in the theme, if that is possible, also another addition seems to me that the dateSelector component is a little bit higher, would be nice if it was a little bit lower to give more space in the map because with the legends enabled this seems a little bit too crowded wdyt @ericboucher @wadhwamatic ?

wadhwamatic:
@JulienPathtech - this is looking good but I still see an issue when I toggle to French - the cursor moves to 1-January. 

English:
Screenshot 2023-06-22 at 22 14 02

French:
Screenshot 2023-06-22 at 22 14 20

ericboucher:
Looks like this bug was introduced with the UX revamp. @JulienPathtech can you have a look? My guess would be that this has something to do with the fact that French dates and English dates have month and days inverted

ericboucher:
Update, I found the issue, which was actually due to the fact that Frenchies don't capitalize months! There might be a better way to do it than what I did @JulienPathtech. Since you've been head down in this component for a few days, let me know if you can think of a better approach.

frontend/src/types/custom.d.ts
JulienPathtech: Yep, to enable importing SVG element at compilation time :)

frontend/src/components/MapView/DateSelector/index.tsx
JulienPathtech: How do I do that without possibly breaking other component current color ? 
I could add a `color.secondary` item with that color but it would not reflec actual application theme

JulienPathtech:
Update : 
- @ericboucher I use the `datesAreEqualWithoutTime` helper. 
- Added comments 

For your point about the Timeline @HarryMytilinaios , I agree. Here are my problems : 
- If I lower / center the Timeline the SVG desapear at the bottom
- I reworked the SVG cursor so it perfectly match the Timeline in size but it's height and width are fixed

So I don't find any good solution to lower the Timeline that ATM.

frontend/src/components/MapView/DateSelector/index.tsx
```diff
@@ -149,16 +149,10 @@ const DateSelector = memo(
...
-          date.label.toLocaleLowerCase() ===
-          moment(stateStartDate)
-            .locale(locale)
-            .format(MONTH_FIRST_DATE_FORMAT)
-            .toLocaleLowerCase()
-        );
-      });
-    }, [locale, range, stateStartDate]);
+      return findIndex(range, date =>
+        datesAreEqualWithoutTime(date.value, stateStartDate || 0),
```
ericboucher, line 153: how about `stateStartDate && datesAreEqualWithoutTime(date.value, stateStartDate)` instead? Same result but maybe a bit more straightforward?

frontend/src/components/MapView/DateSelector/TimelineItems/index.tsx
```diff
@@ -111,8 +105,9 @@ const TimelineItems = memo(
           );
         },
       );
-    }, [selectedLayerDateItems, timelineStartDate]);
+    }, [selectedLayers, timelineStartDate]);
 
+    // Draw a colomn for each date of the Timeline that start at the begining of the year
```
ericboucher, line 110: column?

PR #883, opened 2023-06-21
Restore DataViewer to display charts from map tooltip (cambodia)

HarryMytilinaios:
As it happened the DataViewer component was actually rendering the charts from the Map Tooltip actions, hence there was the error with the charts. In the process of an unused components cleanup I mistakenly removed the DataDrawer component. I restored it and seems that everything is working as expected.


ericboucher (approved):

frontend/src/components/MapView/index.tsx
```diff
@@ -517,6 +518,7 @@ const MapView = memo(({ classes }: MapViewProps) => {
             </Grid>
             {renderedGridItemAlertForm}
           </Grid>
+          <DataViewer />
```
ericboucher, line 521: Let's add a comment noting that at the moment, this is only used for Cambodia. In the future, we could also investigate moving this into the chart drawer.

PR #893, opened 2023-06-28
Fix intersect_percentage API

ericboucher:
- Fix the intersect_percentage calculation by considering nodata pixels as valid for the denominator.
- Add intersect_area statistic

### DIFF
--- a/api/app/main.py
+++ b/api/app/main.py
@@ -18,7 +18,7 @@ from app.kobo import get_form_dates, get_form_responses, parse_datetime_params
 from app.models import AcledRequest, RasterGeotiffModel
 from app.timer import timed
 from app.validation import validate_intersect_parameter
-from app.zonal_stats import GroupBy, calculate_stats, get_wfs_response
+from app.zonal_stats import DEFAULT_STATS, GroupBy, calculate_stats, get_wfs_response
 from fastapi import Depends, FastAPI, HTTPException, Path, Query, Response
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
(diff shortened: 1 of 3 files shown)


api/app/main.py
```diff
@@ -356,7 +356,7 @@ def stats_demo(
     features = _calculate_stats(
         zones_filepath,
         geotiff,
-        stats=" ".join(["min", "max", "mean", "median", "sum", "std"]),
+        stats=" ".join(DEFAULT_STATS),
```
laurentS, line 359: You've dropped "sum" and "std" in doing this. did you mean to?

api/app/zonal_stats.py
```diff
@@ -335,19 +334,24 @@ def calculate_stats(
...
             total = masked.count()
-            # Avoid dividing by 0
             if total == 0:
-                return 0
-            percentage = (intersect_operator(masked, intersect_baseline)).sum()
-            return percentage / total
+                return 0.0
+            return float((intersect_operator(masked, intersect_baseline)).sum())
+
+        def intersect_area(masked) -> float:
```
laurentS, line 348: You have a lot of `float` values which are not of the same type. You would probably make it easier to follow the code (and avoid bugs) by doing something like:
```python
AreaInSqKmPerPixel = NewType("AreaInSqKmPerPixel", float)

def intersect_area(masked) -> AreaInSqKmPerPixel:
    ...
    return AreaInSqKmPerPixel(myvalue)
```

api/app/main.py
ericboucher: nice catch! fixed

api/app/zonal_stats.py
ericboucher: it's just SqKm, but I see your point. Adding SqKm and Percentage

PR #894, opened 2023-07-04
Deactivate layers after failing to load dates

HarryMytilinaios:
This branch deactivates the layers when no available dates are loaded with dispatching a message to info the user. Also this branch fixes a miscalculation of the active layers with the special case of Tropical Storms (Myanmar). Currently this applies to every country. The way to test this is to thow custom exceptions to the different functions that populate the dates to the `getLayersAvailableDates`, such us `localWMSGetLayerDates`, `localFetchCoverageLayerDays`, `getPointDataCoverage`. Those async functions are populating the dates remotely so if one of them throws an error we will not get the possible dates for some layers and we wil be able to test the feature of deactivating those layers that do not have available dates.

### DIFF
--- a/frontend/src/components/MapView/DateSelector/TimelineItems/TimelineLabel/__snapshots__/index.test.tsx.snap
+++ b/frontend/src/components/MapView/DateSelector/TimelineItems/TimelineLabel/__snapshots__/index.test.tsx.snap
@@ -2,11 +2,8 @@
 
 exports[`TimelineLabel renders as expected 1`] = `
 <div>
-  <mock-typography
-    classname="TimelineLabel-dateItemLabel-1"
-    variant="body2"
-  >
-    juillet 2023
-  </mock-typography>
+  <div
+    class="TimelineLabel-dayItem-2"
+  />
 </div>
 `;
--- a/frontend/src/components/MapView/DateSelector/TimelineItems/TimelineLabel/__snapshots__/index.test.tsx.snap
+++ b/frontend/src/components/MapView/DateSelector/TimelineItems/TimelineLabel/__snapshots__/index.test.tsx.snap
@@ -6,7 +6,7 @@ exports[`TimelineLabel renders as expected 1`] = `
     classname="TimelineLabel-dateItemLabel-1"
     variant="body2"
   >
-    juin 2023
+    juillet 2023
   </mock-typography>
 </div>
 `;
--- a/frontend/src/components/MapView/LeftPanel/layersPanel/MenuSwitch/SwitchItem/index.tsx
+++ b/frontend/src/components/MapView/LeftPanel/layersPanel/MenuSwitch/SwitchItem/index.tsx
@@ -43,6 +43,9 @@ import { handleChangeOpacity } from '../../../../Legends/handleChangeOpacity';
 import { Extent } from '../../../../Layers/raster-utils';
 import LayerDownloadOptions from './LayerDownloadOptions';
 import ExposureAnalysisOption from './ExposureAnalysisOption';
+import { availableDatesSelector } from '../../../../../../context/serverStateSlice';
+import { addNotification } from '../../../../../../context/notificationStateSlice';
+import { filterActiveLayers } from '../../../../utils';
 
 const SwitchItem = memo(({ classes, layer, extent }: SwitchItemProps) => {
   const {
@@ -54,6 +57,7 @@ const SwitchItem = memo(({ classes, layer, extent }: SwitchItemProps) => {
   } = layer;
   const { t } = useSafeTranslation();
   const selectedLayers = useSelector(layersSelector);
+  const serverAvailableDates = useSelector(availableDatesSelector);
   const map = useSelector(mapSelector);
   const [isOpacitySelected, setIsOpacitySelected] = useState(false);
   const [opacity, setOpacityValue] = useState<number>(initialOpacity || 0);
@@ -79,16 +83,9 @@ const SwitchItem = memo(({ classes, layer, extent }: SwitchItemProps) => {
 
   const selectedActiveLayer = useMemo(() => {
     return selected
-      ? selectedLayers.filter(sl => {
-          return (
-            (group?.activateAll &&
-              group?.layers.find(l => l.id === sl.id && l.main)) ||
-            (!group?.activateAll &&
-              group?.layers.map(l => l.id).includes(sl.id))
-          );
-        })
+      ? selectedLayers.filter(sl => filterActiveLayers(sl, sl))
       : [];
-  }, [group, selected, selectedLayers]);
+  }, [selected, selectedLayers]);
 
   const initialActiveLayer = useMemo(() => {
     return selectedActiveLayer.length > 0 ? selectedActiveLayer[0].id : null;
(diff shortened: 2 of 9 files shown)


frontend/src/components/MapView/DateSelector/TimelineItems/TimelineLabel/__snapshots__/index.test.tsx.snap
```diff
@@ -6,7 +6,7 @@ exports[`TimelineLabel renders as expected 1`] = `
     classname="TimelineLabel-dateItemLabel-1"
     variant="body2"
   >
-    juin 2023
+    juillet 2023
```
ericboucher, line 9: this means that one of our tests doesn't have a properly mocked date @HarryMytilinaios

frontend/src/components/MapView/LeftPanel/layersPanel/MenuSwitch/SwitchItem/index.tsx
```diff
@@ -79,16 +83,9 @@ const SwitchItem = memo(({ classes, layer, extent }: SwitchItemProps) => {
...
     return selected
-      ? selectedLayers.filter(sl => {
-          return (
-            (group?.activateAll &&
-              group?.layers.find(l => l.id === sl.id && l.main)) ||
-            (!group?.activateAll &&
-              group?.layers.map(l => l.id).includes(sl.id))
-          );
-        })
+      ? selectedLayers.filter(sl => filterActiveLayers(sl, sl))
```
ericboucher, line 86: Maybe add a comment? It feels weird to have a function that takes twice the exact same argument

frontend/src/components/MapView/DateSelector/TimelineItems/TimelineLabel/__snapshots__/index.test.tsx.snap
HarryMytilinaios: Yes I think so this test was failing for the TimelineLabels before I push my code. I think when this test was created we had june and the props we are passing set a new Date() with the appropriate locale this is the reason for this. Do you want me to change it?

frontend/src/components/MapView/index.tsx
```diff
@@ -215,9 +215,26 @@ const MapView = memo(({ classes }: MapViewProps) => {
   const addMissingLayers = useCallback(() => {
     missingLayers.forEach(layerId => {
       const layer = LayerDefinitions[layerId as LayerKey];
+      const { serverLayerName } = layer as any;
+      if (
+        // The layer does not have date support
+        serverAvailableDates[serverLayerName] !== undefined &&
+        // The layer does have date support but no additional available dates are loaded
+        serverAvailableDates[serverLayerName].length === 0
```
ericboucher, line 223: The top comment is a bit confusing imho.

Could this be simply: `serverAvailableDates[serverLayerName]?.length || -1) === 0`

frontend/src/components/MapView/LeftPanel/layersPanel/MenuSwitch/SwitchItem/index.tsx
HarryMytilinaios: Of course

frontend/src/components/MapView/index.tsx
HarryMytilinaios: Yes the way I did this is because when a layer does not exist at all in the serverAvailableDates, which will have `undefined` value it means that the layer does not have date support but it can be enabled, only if the layer is included in the serverAvailableDates and has a length of 0 means that it does not have available dates.

ericboucher: What you are actually checking is that the layer has date support BUT not dates were returned by the date endpoint.

ericboucher: Can't you simply check `serverAvailableDates[serverLayerName]?.length === 0`

frontend/src/components/MapView/utils.ts
```diff
@@ -153,6 +153,13 @@ export const filterActiveLayers = (
   selectedLayer: LayerType,
   categoryLayer: LayerType,
 ): boolean | undefined => {
+  // Special case for tropical storms `adamts_buffers`
+  if (
+    selectedLayer.id.includes('adamts') &&
+    selectedLayer.id.includes('adamts')
+  ) {
```
ericboucher, line 160: You mean categoryLayer for the second one?
Can you explain why this is a special case? Ideally, we want to avoid named exceptions like this.

frontend/src/components/MapView/DateSelector/TimelineItems/TimelineLabel/__snapshots__/index.test.tsx.snap
ericboucher: yes, let's find where this prop is getting set and fix the date in jest if necessary

ericboucher: You can see in `/DateSelector/index.test.tsx` how we "fix" the date

frontend/src/components/MapView/utils.ts
HarryMytilinaios: Yes because the Tropical Storms is not one layer with group, but it is splitted into 3 different layers the default first one has a group and the other 2 have not, and when those are going to be compared they have different ids so they are going to be compared with the function `filterActiveGroupedLayers`, but because the 2 tropical storms layers will not have a group will be listed as different layers so the summary of the functions will going to give us 3 selected layers instead of 1

frontend/src/components/MapView/index.tsx
HarryMytilinaios: Yes you are right I am changing it.

frontend/src/components/MapView/index.tsx
```diff
@@ -215,9 +215,21 @@ const MapView = memo(({ classes }: MapViewProps) => {
   const addMissingLayers = useCallback(() => {
     missingLayers.forEach(layerId => {
       const layer = LayerDefinitions[layerId as LayerKey];
+      const { serverLayerName } = layer as any;
```
ericboucher, line 218: Why do we need the same logic in two places?

frontend/src/components/MapView/utils.ts
ericboucher: @HarryMytilinaios I also meant that right now this is testing twice for the same thing `selectedLayer.id.includes('adamts')`

frontend/src/components/MapView/LeftPanel/layersPanel/MenuSwitch/SwitchItem/index.tsx
ericboucher: @HarryMytilinaios if the change was useful you can keep it, I would just recommend adding a comment to explicit the behavior

HarryMytilinaios: I think this was not the case here and the change was not correct this is why I removed it.

frontend/src/components/MapView/utils.ts
HarryMytilinaios: You are right I will remove it and investigate a little bit more.

frontend/src/components/MapView/index.tsx
ericboucher: ok, maybe create a util function?

HarryMytilinaios: Done

frontend/src/components/MapView/utils.ts
HarryMytilinaios: Done

ericboucher (approved):

PR #897, opened 2023-07-06
Removes relative paths from imports

HarryMytilinaios:
This PR removes the relative paths from appplication's imports, throughout all the files.

Adds: 

```
"compilerOptions": {
    ...
    "baseUrl": "./src"
  }
```
to tsconfig.json to enable us to import directly from src instead of using relative paths
`(eg. '../../context/dataStateSlice'` => `'context/dataStateSlice')`

### DIFF
--- a/frontend/src/components/MapView/AlertForm/__snapshots__/index.test.tsx.snap
+++ b/frontend/src/components/MapView/AlertForm/__snapshots__/index.test.tsx.snap
@@ -44,11 +44,594 @@ exports[`renders as expected 1`] = `
           >
             Hazard Layer
           </mock-typography>
-          <mock-layer-dropdown
-            classname="AlertForm-selector-10"
-            placeholder="Choose hazard layer"
-            type="wms"
-          />
+          <div
+            class="MuiFormControl-root AlertForm-selector-10"
+          >
+            <mock-textfield
+              classes="[object Object]"
+              inputlabelprops="[object Object]"
+              inputprops="[object Object]"
+              select="true"
+              variant="outlined"
+            >
+              <mock-menuitem
+                disabled="true"
+                value="placeholder"
+              >
+                Choose hazard layer
+              </mock-menuitem>
+              <li
+                class="MuiListSubheader-root MuiListSubheader-sticky MuiListSubheader-gutters"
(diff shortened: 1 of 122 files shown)


ericboucher:
Nice. Now let’s add the commit hashes to `.git-blame-ignore-revs` please

frontend/src/components/MapView/AlertForm/__snapshots__/index.test.tsx.snap
```diff
@@ -44,11 +44,594 @@ exports[`renders as expected 1`] = `
           >
             Hazard Layer
           </mock-typography>
-          <mock-layer-dropdown
```
ericboucher, line 47: I'm surprised that this test changed so much. Maybe some modules are not mocked properly anymore? Could you investigate @HarryMytilinaios ?

HarryMytilinaios: of course

frontend/src/components/MapView/AlertForm/index.test.tsx
```diff
@@ -3,10 +3,8 @@ import { Provider } from 'react-redux';
 import { render } from '@testing-library/react';
 import { ThemeProvider } from '@material-ui/styles';
 import { createMuiTheme } from '@material-ui/core/styles';
+import { store } from 'context/store';
 import AlertForm from '.';
-import { store } from '../../../context/store';
-
-jest.mock('../Layers/LayerDropdown', () => 'mock-Layer-Dropdown');
```
ericboucher, line 9: @HarryMytilinaios that's the reason for the snapshot changes

HarryMytilinaios: Yes I re-added it. I was trying to find a way for jest to identify the absolute paths, but everything I did in a `jest.config.js` was not correct, and I wasn't sure how to make jest to identify the mock modules through absolute and not relative paths, this is the reason that this was mistakenly deleted. Right now absolute paths are working for every file module but the `jest.mock` is only working with relative paths

ericboucher (approved):

Issue #792, opened 2023-04-07
[Feature Request]: Modify report feature implementation to a server-side task


wadhwamatic:
### Provide a clear and concise description of what the problem is.

The report feature can be extremely slow as we can have 1000+ rows of data to render into a table. Even on relatively fast machines, this can take quite a bit of time and slows down the browser significantly 

### Provide a clear and concise description of what you want to happen.

Move elements of the report generation functionally to a server side process. 

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

@ericboucher already has some thoughts and suggestions on this

PR #900, opened 2023-07-10
Generate pdf report using playwright; COUNTRY=cambodia;

JulienPathtech:
This fixes issue #792.

This PR implements report generation in the backend with a 48h cache.

How to test the feature:

1. https://prism-900.surge.sh/?hazardLayerIds=flood_extent&date=2023-07-07
2. Flood extent -> Exposure Analysis -> Create Report 

screenshots: 

image

We give a shortcut with "Download report" to the user to avoid the preview and have a direct download. That way we can measure if users really need a preview. It also a mean to avoid triggering the slow react-pdf renderer for the user

Questions for @wadhwamatic:
- Currently, Preview Report still uses the frontend generation and is quite slow. Now that download report is faster, should we remove the preview button?
- Is 48h ok for the cache?


JulienPathtech:
@ericboucher 
Despite, it's working, I'm getting less and less confortable with this feature : 
- The fact that we open a headless chrome from python is some complexity and it might be hard to debug or fix issue in the future.
- If you want to keep the preview as of today, client browser we still need to use the existing slow pdf-rendere lib to do so. (i.e we don't make the end user win time in the process) 
- We need to re-compute layer in the headless browser wich is some double work

frontend/src/components/MapView/LeftPanel/AnalysisPanel/ExposureAnalysisActions/index.tsx
```diff
@@ -87,6 +87,8 @@ function ExposureAnalysisActions({
     };
   };
 
+  const API_URL = 'http://localhost:80/report/';
```
JulienPathtech, line 90: Change to preview API if any or production URL otherwise

api/app/playwright_renderer.py
```diff
@@ -0,0 +1,106 @@
...
+            "./report-"
+            + layerIdParam
+            + "-"
+            + language
+            + "-"
+            + str(random_uuid)
+            + ".pdf"
+        )
+
+        await download.save_as(report_file_path)
```
JulienPathtech, line 70: You should know reports are going to pile up in the production container.
- I didn't implemented a cleanup since each new API deployment should remove those files
- In the long run we might implement a cache strategy for them

ericboucher: I don't think new API deployments actually remove any files. But it's a reasonable idea

api/app/playwright_renderer.py
```diff
@@ -0,0 +1,112 @@
...
+# Timeouts
+PAGE_TIMEOUT: Final[int] = 25000
+PAGE_LANGUAGE_CHANGE_TIMEOUT: Final[int] = 5000
+
+
+async def playwright_download_report(url: str, language: Optional[str]) -> str:
+    language = "en" if language is None else language
+    layerIdParam = extract_query_param(url, "hazardLayerIds")
+    dateParam = extract_query_param(url, "date")
+    report_file_path = (
```
ericboucher, line 24: you should probably use os.path.join(

api/app/playwright_renderer.py
```diff
@@ -0,0 +1,112 @@
...
+        await page.wait_for_selector(LAYER_ACCORDION_SELECTOR, state="visible")
+
+        # Toggle level one dropdowns
+        await toggle_every_visible_dropdown(page)
+
+        # Toggle level two dropdowns
+        await toggle_every_visible_dropdown(page)
+
+        # Click on exposure analysis toggle button
+        await click_target_exposure_analysis(page, layerIdParam)
```
ericboucher, line 55: this would be broken if multiple layerIds are in the list

api/app/playwright_renderer.py
```diff
@@ -0,0 +1,112 @@
...
+
+        # Wait for page to be loaded on exposure analysis
+        await page.wait_for_selector(CREATE_REPORT_BUTTON_SELECTOR, state="visible")
+
+        # Change language if not english
+        await change_language_if_not_default(language, page)
```
ericboucher, line 61: Some functions have page as the first argument while others have page as the second. Let's try to be consistent to avoid confusions

api/app/playwright_renderer.py
ericboucher: You could also type your functions when possible

JulienPathtech: I made my comment before I implemented the logic that put reports in the `/cache` volume. So yes now they won't be removed

api/app/playwright_renderer.py
ericboucher: Maybe loop through layerIds until successful?

frontend/src/components/MapView/LeftPanel/AnalysisPanel/ExposureAnalysisActions/index.tsx
```diff
@@ -81,6 +85,43 @@ function ExposureAnalysisActions({
     [analysisDefinition, exposureAnalysisCsvData],
   );
 
+  const handleDownloadReport = async () => {
```
ericboucher, line 88: let's add a bit of context here

frontend/src/components/MapView/LeftPanel/AnalysisPanel/ExposureAnalysisActions/index.tsx
```diff
@@ -97,8 +138,20 @@ function ExposureAnalysisActions({
...
       )}
-      <Button className={bottomButton} onClick={handleToggleReport(true)}>
-        <Typography variant="body2">{t('Create Report')}</Typography>
+      <Button
+        id="create-report"
+        className={bottomButton}
+        onClick={handleToggleReport(true)}
+      >
+        <Typography variant="body2">{t('Preview Report')}</Typography>
+      </Button>
```
ericboucher, line 147: This is using the old method, right? So very slow. Maybe we should add a warning?

`Preview Report (slow)` - wdyt?

JulienPathtech: Yes it's the old method. I'm OK with adding (slow) label.
I will also try to add missing translations for that key

api/app/playwright_renderer.py
JulienPathtech: Looping seems good, can you give a case where I have 2 hazardLayerIds in query param ?

frontend/src/components/MapView/LeftPanel/AnalysisPanel/ExposureAnalysisActions/index.tsx
JulienPathtech: I added some comment, don't hesitate if any suggestion or want some more details

ericboucher (approved):

wadhwamatic:
> Questions for @wadhwamatic:
> 
> * Currently, Preview Report still uses the frontend generation and is quite slow. Now that download report is faster, should we remove the preview button?
> * Is 48h ok for the cache?

@JulienPathtech - thanks for working on this. For your questions above:
1. We don't need to have a 'preview', just go directly to download
2. This depends on what you've implemented. I assume that if a new flood extent data layer is available, we'll have a new report generated which then gets cached. Most users would go straight to the latest available report, which as of now, would get updated every day.

I wasn't able to test this as I got an error when I tried to download the report. I'm on Mac OS 13.2 running Chrome 114.0.5735.198 (Official Build) (arm64). I get the same error on Safari 

Screenshot 2023-07-28 at 09 00 31

Issue #793, opened 2023-04-07
[Feature Request]: Provide design options to access layer functionality while the layer is not visible in the left panel


wadhwamatic:
### Provide a clear and concise description of what the problem is.

Transparency, downloading layer data, removing a layer, and running an exposure analysis require the layer navigation to be expanded with the layer of interest in view. This means you can't access any of those features if the navigation is collapsed

### Provide a clear and concise description of what you want to happen.

We need some UX / UI options that allow users to access these functions even if the navigation is collapsed or the left panel is hidden completely. We could consider adding a kebab menu or icons for each layer in the legend section.

hav142:
How about adding the icons for transparency control, data download, and layer removal into the legend section, as shown in below picture? It can provide convenient access to the users. However, we should also consider issue #780 which involves obtaining the HTML element of the legend to be drawn onto the canvas.

![legend_mockup](https://github.com/WFP-VAM/prism-app/assets/25983260/a7f88302-07a0-4f20-8266-2cb62f64513e)

wadhwamatic:
Thanks @hav142 - I like the icon approach. @ericboucher - what do you think?

ericboucher:
yes, that seems fine. If we want to have more space, in case the title is longer for example, we could have the icons as a new row at the bottom of the legend

wadhwamatic:
> yes, that seems fine. If we want to have more space, in case the title is longer for example, we could have the icons as a new row at the bottom of the legend

That's a good suggestion as the title can easily break the formatting. @hav142 - please consider this as you start working on this again

PR #919, opened 2023-07-14
Add layer config access in legend

hav142:
This PR is still drafted to fixes issue #793 

How to test the feature:

- activate any layer
- check the icons on the legend


wadhwamatic:
Nice work on this @hav142! Some minor inputs from me which are just for consideration:
1. I think the icons can now be left aligned since they no longer are in-line with the layer name. What do you think?
2. A little more spacing between the legend detail text and the icons would be good to give some more room
3. I think we can space out the icons a bit too

hav142:
@wadhwamatic Thanks for the suggestions. Is this good enough?

image

wadhwamatic:
@hav142 - yes, that looks good. as you move forward with this, please have a look at https://github.com/WFP-VAM/prism-app/pull/920 to make sure that the CSV download feature for exposure analysis works with your revised UI

ericboucher:
@hav142 can you have a look at the linting issues? The demo endpoint will only get deployed if all the tests and linters are passing. Thanks!

frontend/yarn.lock
```diff
@@ -2690,14 +2690,28 @@
   dependencies:
     "@types/react" "*"
 
-"@types/react@*", "@types/react@^16.9.0":
+"@types/react@*":
   version "16.9.23"
```
ericboucher, line 2694: Should we update the types too?

wadhwamatic:
Thanks for picking this up again @echaidemenos. This works well except when I try to download a raster (WMS) layer. The download link works in the layer navigation, but not in the legend.

Screenshot 2023-11-08 at 13 33 35

@ericboucher - this sounds related to the issue we saw earlier today.

ericboucher (approved):

wadhwamatic (approved):
Looks good, thanks @echaidemenos !

PR #924, opened 2023-07-25
Silence "converting a masked element to nan" warning

ericboucher:
Silence "converting a masked element to nan" logs warnings after the first occurrence to avoid spamming the logs


api/app/utils.py
```diff
@@ -14,3 +14,18 @@ def forward_http_error(resp: requests.Response, excluded_codes: list[int]) -> No
         obj = resp.json()
         detail = obj.get("detail", "Unknown error")
         raise HTTPException(status_code=resp.status_code, detail=detail)
+
+
+# Define a flag to indicate if the warning has been recorded already
+warning_recorded = False
+
+
+def custom_warning_handler(message, category, filename, lineno, file=None, line=None):
```
JulienPathtech, line 23: I don't know if it's an issue for you, but https://docs.python.org/3/library/warnings.html#warnings.showwarning write to a file which is sys.stderr by default. 

We could call the default `_showwarnmsg_impl`  for messages we don't want to filter to be clother to existings behavior ?

JulienPathtech (approved):

api/app/utils.py
ericboucher: This is an intenal function and looks like it has different signature based on the pyhton version. I'll use a standard logger instead, wdyt?

PR #925, opened 2023-07-28
Pdf report - rename backend functions; COUNTRY=cambodia

ericboucher:
- Rename functions
- Add white blinking dots


wadhwamatic:
Looking good so far from my side @ericboucher. I need to figure out why we don't have new flood extent coming in though.

PR #929, opened 2023-10-04
Update dev tools config

### LINKED ISSUE #731: [Bug]: conflicting linter configs between docker-compose and poetry
### What happened?

The black configuration is not the same if we run through docker (make api-lint), or through poetry (like the CI does).
We need to decide on one way to run the linter, or make sure they use the same config.

This might also apply to `isort`

N/A

_No response_

laurentS:
This PR fixes #731 and updates a few dev tools to latest versions.

## docker compose "breaking" change

Docker compose v1 reached end of life in July.
The new command format should be used now (`docker compose` instead of `docker-compose`).
See https://docs.docker.com/compose/migrate/ for details.


ericboucher:
TODO - fix python test init / "user_info" error

Issue #930, opened 2023-10-05
[Feature Request]: Compare a chart indicator over two time periods


wadhwamatic:
### Provide a clear and concise description of what the problem is.

The chart feature currently allows a user to visualize and access time-series hazard data for one location and one period (date range) at a time. Users want to be able to compare hazard trends between two places, and for the same place over two time periods. Currently, this requires multiple steps to download data locally and then create charts in another tool for comparison.

### Provide a clear and concise description of what you want to happen.

The PRISM charts UI needs to be updated so that users can either compare two time periods for the same location, or compare two locations for the same time period.

This will require a few changes:
1. The chart UI needs two new elements to either compare periods or compare locations (not both at the same time)
2. If comparing periods, two date pickers are needed
3. If comparing locations, two location drop downs are needed. The level (national, admin 1, admin 2) should be flexible so users can compare across levels if they choose
4. Only one indicator can be used when comparing periods or comparing locations

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

**Mockups**

Compare periods
![01-compare-periods](https://github.com/WFP-VAM/prism-app/assets/3343536/3c76ae25-bd2e-4074-8671-d2749d39b30c)

Compare locations
![02-compare-locations](https://github.com/WFP-VAM/prism-app/assets/3343536/8d0fc7f1-7c98-49d2-add6-3979c7eaa9c9)

**User stories**
1. As an analyst, I need to compare a hazard layer between two time periods, so that I can understand how severe it is compared to a historical event
2. As an analyst, I need to compare a hazard layer between two locations, so that I can determine where to prioritize any response

laurentS:
What would the "Download CSV" button do when comparing locations/periods? Should 2 buttons be shown? no button at all?

wadhwamatic:
Hi @laurentS. The download CSV button should remain in place. We should make each location/period a separate CSV file. The download could either be a zip with both files, or serve two separate CSV files in a single click.

Issue #931, opened 2023-10-05
[Feature Request]: The date range for chart data should be flexible


wadhwamatic:
### Provide a clear and concise description of what the problem is.

The chart features let's a user see a time-series trend for a 1-year period. That is fine for default behavior, but users often have a specific start / end dates they want to analyze. Currently, this requires them to make multiple chart requests with different end dates to get to the data they want.

### Provide a clear and concise description of what you want to happen.

The following changes are needed in the charts feature:
1. Include separate date pickers for the start and end date. The default values should be an end date of today, and start date of 1 year ago.
3. Do not allow the end date to come before the start date. Alert users if try to select this
4. Do not limit the length of the date range. It can be of any duration
5. Scale the chart based on the selected range
6. Continue to allow multiple indicators to be selected
7. Continue to allow CSV downloads of the data selected 

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

Nice to have: add a slider in the UI to adjust the date range on the chart axis

ericboucher:
@wadhwamatic would you like to open another issue for the slider if you think it can be useful?
`add a slider in the UI to adjust the date range on the chart axis`

PR #932, opened 2023-10-13
Compare charts over location/time periods; COUNTRY=cambodia

laurentS:
This fixes issue #930 and #931 

How to test the feature:

- go to the charts tab
- click the "compare locations" switch
- select locations 1 & 2
- select time period and chart layers
- see charts show up

## TODO
- [x] Provide 2 CSV files when comparing 
- [x] Limit number of selectable layers to 2 (is this really needed?)
- [x] Adjust styling
- [x] Add time period comparison

Screenshot/video of feature:
Screenshot 2023-10-23 at 5 18 38 PM


laurentS:
This is more or less ready to review. There's an open point in the todo list above. I think it makes sense to keep the code as is, allowing multiple layers to be compared, but maybe I'm missing something on the "business" logic side.

Also CSV filenames are a bit chaotic, not sure how to make them nicer. Ideas welcome!

wadhwamatic:
This is looking great, thanks @laurentS! I would like to test a bit more locally tomorrow with other regions but the surge deployment mostly works as expected. I did see strange behavior with the CSV filenames. It seems an admin 2 name is being added to the filename even if I am only requesting to admin 1. See screenshots below.

Screenshot 2023-10-23 at 18 48 34
Screenshot 2023-10-23 at 18 51 14
Screenshot 2023-10-23 at 18 45 49
Screenshot 2023-10-23 at 18 45 41

I'm also thinking about how we can improve the UX for time period selection. More on this tomorrow!

wadhwamatic:
Comparing locations when we have admin0 (country-level) data crashes the app. Screenshot below.

Steps to reproduce:
1. COUNTRY=rbd
2. Go to charts choose a single location (country, Benin for example)
3. Keep default dates
4. Choose any chart
5. Next, toggle 'compare locations'. The app will crash with the message below. This seems to only happen if you choose a chart for a single country and then try to compare locations afterwards 

Screenshot 2023-10-24 at 17 25 40

wadhwamatic:
When comparing locations for the RBD deployment, we don't need the text 'Regional Bureau Dakar' in the location header. It incorrectly replaces the country name. See screenshots below.

Screenshot 2023-10-24 at 17 49 55

Screenshot 2023-10-24 at 17 50 33

laurentS:
> Comparing locations when we have admin0 (country-level) data crashes the app. Screenshot below.

Fixed in 88c70e5172d70f140c53e091929ac8a75f8195be

> When comparing locations for the RBD deployment, we don't need the text 'Regional Bureau Dakar' in the location header.

Fixed in ca3b242a8df7b0b2cb63f5c73d3b022a3b0fcc66

> It seems an admin 2 name is being added to the filename even if I am only requesting to admin 1.

Fixed in 8445cd2695cec00567ed95e9b8ddb8a40932d19f

I think I got all your comments above.

As a side note, loading the charts page is extremely slow on my machine (1 year old laptop, reasonably powerful) on a 1Gb/s fiber internet connection. It seems that with country=rbd, there is something like 160MB of country boundary data that is loaded. Most of it is cached, but the page still takes over 12 seconds to be responsive, which seems like too much to me. Not sure if we can do something about it?

frontend/src/components/MapView/LeftPanel/ChartsPanel/index.tsx
```diff
@@ -264,157 +261,143 @@ const ChartsPanel = memo(
...
-        })?.value;
-        if (data) {
-          setAdminProperties(getProperties(data, admin2Id));
-        }
-        setAdmin2Key(event.target.value);
-        setAdminLevel(2);
-      },
-      [selectedAdmin1Area, data],
-    );
+    const generateSecondCSVFilename = useCallback(() => {
```
ericboucher, line 342: Can we find a better name than"SecondCSVFilename"?

wadhwamatic (approved):
Thanks for the fixes @laurentS. All looks good from my side now.

Regarding the performance issues, I don't see it so badly on my side but this has been a known issue. See https://github.com/WFP-VAM/prism-app/issues/467. cc @ericboucher

PR #936, opened 2023-10-18
Testing deployment script on new machine

lowriech:
This introduces a deployment script for working with secrets.  

Notes:
- Not sure about the process for deploying `alerts` at the moment, since it's not running on Docker on the existing machine I'm replicating.
- Also not sure about the process for deploying to a new hostname.  Didn't want to push that and accidentally replace the existing API, @ericboucher if you could advise.

### DIFF
--- a/alerting/docker-compose.yml
+++ b/alerting/docker-compose.yml
@@ -26,5 +26,5 @@ services:
       - POSTGRES_DB=postgres
       - POSTGRES_PASSWORD=!ChangeMe!
       - DATABASE_URL
-      - EMAIL_PASSWORD
-      - EMAIL_USER
+      - EMAIL_PASSWORD=${ALERTING_EMAIL_PASSWORD} # Set from secrets in api/set_envs.sh
+      - EMAIL_USER=${ALERTING_EMAIL_USER}
--- a/api/README.md
+++ b/api/README.md
@@ -106,13 +106,13 @@ Generate a geotiff for any wfp raster using the stac API and saves it in S3. It 
 To run the api locally, run:
 
 ```
-make api
+source set_envs.sh && make api
 ```
 
 To run flask api together with database within same network, run:
 
 ```
-docker-compose -f ./docker-compose.develop.yml -f ../alerting/docker-compose.yml up
+source set_envs.sh && docker-compose -f ./docker-compose.develop.yml -f ../alerting/docker-compose.yml up
 ```
 
 ### Tests
@@ -120,7 +120,7 @@ docker-compose -f ./docker-compose.develop.yml -f ../alerting/docker-compose.yml
 To run linting and tests, run:
 
 ```
-make test
+source set_envs.sh && make test
 ```
 
 ## Deployments
@@ -132,7 +132,7 @@ Specifically, update `info@ovio.org` with a domain admin email and `prism-api.ov
 To deploy, run:
 
 ```
-make deploy
+source set_envs.sh && make deploy
 ```
 
 There are a few known issues happening from time to time
--- a/api/docker-compose.deploy.yml
+++ b/api/docker-compose.deploy.yml
@@ -46,12 +46,12 @@ services:
     environment:
       - "RUN=uvicorn app.main:app --host 0.0.0.0 --port 80"
       # make sure that the alert password is properly set
-      - DATABASE_URL=${DATABASE_URL:?"Provide a URL for the alerts database."}
-      - KOBO_PW=${KOBO_PW:?"Add a password to access KOBO data."}
+      - DATABASE_URL=${PRISM_ALERTS_DATABASE_URL:?"Provide a URL for the alerts database."}
+      - KOBO_PW=${KOBO_PASSWORD:?"Add a password to access KOBO data."}
       - KOBO_USERNAME=${KOBO_USERNAME:-kobo_user}
       - HDC_TOKEN=${HDC_TOKEN:?"Provide a token to access HDC chart data."}
-      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:?"Provide the AWS access key for the stac api."}
-      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY:?"Provide the AWS secret key for the stac api."}
+      - AWS_ACCESS_KEY_ID=${STAC_AWS_ACCESS_KEY_ID:?"Provide the AWS access key for the stac api."}
+      - AWS_SECRET_ACCESS_KEY=${STAC_AWS_SECRET_ACCESS_KEY:?"Provide the AWS secret key for the stac api."}
     command: uvicorn app.main:app --host 0.0.0.0 --port 80
     restart: always
     # adding a DNS to allow self lookup of the api
--- a/api/docker-compose.deploy.yml
+++ b/api/docker-compose.deploy.yml
@@ -40,7 +40,7 @@ services:
       - "80:80"
     labels:
       - "traefik.enable=true"
-      - "traefik.http.routers.whoami.rule=Host(`prism-api.ovio.org`)"
+      - "traefik.http.routers.whoami.rule=Host(${HOSTNAME:?'Must provide hostname'})"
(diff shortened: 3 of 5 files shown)


ericboucher:
I think pushing the hostname is fine, nothing should happen unless we change it on the DNS side as well

.github/workflows/api.yml
```diff
@@ -50,7 +50,15 @@ jobs:
...
+          role-to-assume: arn:aws:iam::613752250312:role/github_production_deploy
+          aws-region: us-east-2
+        run: cd api && source set_envs.sh && make build
       - name: Run tests
-        run: cd api && make api-test
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::613752250312:role/github_production_deploy
+          aws-region: us-east-2
+        run: cd api && source set_envs.sh && make api-test
```
ericboucher, line 63: Are these changes needed?

api/README.md
```diff
@@ -105,21 +105,21 @@ Generate a geotiff for any wfp raster using the stac API and saves it in S3. It
 To run the api locally, run:
 
 ```
-make api
+source set_envs.sh && make api
```
ericboucher, line 108: I don't think we want to rely on AWS for everything. Not all the developers working on PRISM have access to the AWS console.

api/docker-compose.develop.yml
```diff
@@ -14,8 +14,10 @@ services:
       - POSTGRES_DB=postgres
       - POSTGRES_PORT=54321
       - POSTGRES_PASSWORD=!ChangeMe!
-      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
-      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
+      - KOBO_PASSWORD=${KOBO_PASSWORD:?"Add a password to access KOBO data."}
+      - KOBO_USERNAME=${KOBO_USERNAME:?"Add a user to access KOBO data."}
```
ericboucher, line 18: If they are declared here you don't need to redeclare them in .deploy if you are not changing the error handling

api/README.md
lowriech: Makes sense.  It's not clear to me what should be able to be tested at what stage of development, given that we're sometimes passing in inactive API keys.

.github/workflows/api.yml
lowriech: Only if we want to test with active API keys to see that functionality for ie ACLED, STAC, etc works.  For now I've removed since it requires modifying permissions and seems like we weren't doing that before.

api/docker-compose.develop.yml
lowriech: Is that true?  `.deploy` doesn't pick up on `.develop` in the Makefile.

lowriech:
Here's where we're at:
- We can build and deploy using AWS Secrets running in EC2, without having to manually pass around keys
- The instance is up and running and the deploy command has been run

Some open questions/issues:
1. There's an issue running code coverages tests, that I've filed [here](https://github.com/WFP-VAM/prism-app/issues/938).  Seems to be present in master as well for the code coverage, so not addressing as part of this ticket.  
2. It's not clear what should be able to be tested and when.  Seems like we generally align to not testing the use of external services until we deploy.  Which is outside of the scope of this ticket, but does make it difficult to know if I'm surfacing new or existing issues while developing.  I've kept the secrets to only be required when deploying, not when developing, although that means some functionality won't be available without us passing around ENV vars.
3. What if anything about the alerts DB changes with a new deployment?
4. And the DNS needs to be changed to point at a new EC2 instance.

For now, I've run the deployment commands and it is up-and-running at `3.134.94.21`.  

@ericboucher could you advise on points 3 and 4?

api/docker-compose.develop.yml
ericboucher: Sorry I wasn't clear :)
If most of the files use the same behavior, you could define it in the base docker-compose.yml directly and only override it when needed (eg. for tests)

alerting/docker-compose.yml
```diff
@@ -25,6 +25,6 @@ services:
       - POSTGRES_USER=postgres
       - POSTGRES_DB=postgres
       - POSTGRES_PASSWORD=!ChangeMe!
-      - DATABASE_URL
-      - EMAIL_PASSWORD
-      - EMAIL_USER
+      - PRISM_ALERTS_DATABASE_URL
+      - ALERTING_EMAIL_PASSWORD=${ALERTING_EMAIL_PASSWORD} # Set from secrets in api/set_envs.sh
+      - ALERTING_EMAIL_USER=${ALERTING_EMAIL_USER}
```
laurentS, line 30: If these 3 are related, maybe rename them to use a common prefix to make this clearer? `PRISM_ALERTS_EMAIL_PASSWORD` and `PRISM_ALERTS_EMAIL_USER` ?

api/README.md
```diff
@@ -147,7 +147,7 @@ Specifically, update `info@ovio.org` with a domain admin email and `prism-api.ov
 To deploy, run:
 
 ```
-make deploy
+source set_envs.sh && make deploy
```
laurentS, line 150: Could the `source...` part be moved to the `Makefile` instead? I find it nice when all the useful commands are available in there with tab completion. And it's slightly less error-prone.

laurentS: Also noting that the text above (line 145) is outdated as the variable definition moved elsewhere.

api/docker-compose.deploy.yml
```diff
@@ -2,7 +2,7 @@ version: "3"
 
 networks:
   traefik_default:
-    external: true
+    external: false
```
ericboucher, line 5: why this change?

api/docker-compose.deploy.yml
```diff
@@ -40,18 +40,18 @@ services:
       - "80:80"
     labels:
       - "traefik.enable=true"
-      - "traefik.http.routers.whoami.rule=Host(`prism-api.ovio.org`)"
+      - "traefik.http.routers.whoami.rule=Host(${HOSTNAME:?'Must provide hostname'})"
```
ericboucher, line 43: Is this properly getting set on the new instance? Anyway to confirm?

ericboucher: I think this is missing the ` ` no?

api/docker-compose.deploy.yml
lowriech: The current redeploy didn't work because it requires a network being created beforehand.  This is just to create the network, rather than expecting an existing network.

I could also just put the network creation command in CLI somewhere, do you think this is throwing of HTTPS?

ericboucher: I think that's fine as is actually

api/docker-compose.deploy.yml
lowriech: Fixed ✅

api/README.md
lowriech: Fixed ✅

alerting/docker-compose.yml
lowriech: Fixed ✅

api/docker-compose.deploy.yml
```diff
@@ -18,14 +18,16 @@ services:
...
+      - "--certificatesresolvers.myresolver.acme.email=$INFO_EMAIL"
       - "--certificatesresolvers.myresolver.acme.storage=/letsencrypt/acme.json"
       - traefik.docker.network=traefik_default
     networks:
       - traefik_default
     ports:
       - "443:443"
       - "8080:8080"
+    environment:
+      - INFO_EMAIL=${INFO_EMAIL:?'Must provide info email'}
```
ericboucher, line 30: is this needed?

alerting/docker-compose.yml
ericboucher: This does not appear to be fixed?

ericboucher (approved):
A few minor comments left to address but otherwise this is gtg

api/docker-compose.deploy.yml
lowriech: Good catch, not needed

Issue #938, opened 2023-10-23
[Bug]: API Coverage


lowriech:
### What happened?

When running `make api-coverage`, we're encountering the following issue:
```
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --cov=app --cov-config --cov-fail-under=30 --cov-report term-missing
  inifile: None
  rootdir: /app
```

This has happening in master at the moment, not clear when it was introduced.

### Which country / deployment are you running?

Testing master.

PR #948, opened 2023-10-31
Fix pytest-cov

ericboucher:
This fixes issue #938

### DIFF
--- a/api/docker-compose.yml
+++ b/api/docker-compose.yml
@@ -6,6 +6,9 @@ services:
     volumes:
       - "./cache:/cache:rw"
       - ./app:/usr/local/lib/python3.9/site-packages/app:ro
+      # Use these to run the app in development mode and update depedencies
+      # - ./pyproject.toml:/pyproject.toml
+      # - ./poetry.lock:/poetry.lock
     environment:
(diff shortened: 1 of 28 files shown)


api/docker-compose.yml
```diff
@@ -6,6 +6,9 @@ services:
     volumes:
       - "./cache:/cache:rw"
       - ./app:/usr/local/lib/python3.9/site-packages/app:ro
+      # Use these to run the app in development mode and update depedencies
+      # - ./pyproject.toml:/pyproject.toml
+      # - ./poetry.lock:/poetry.lock
```
laurentS, line 11: Could these be in `docker-compose.develop.yml` instead? Also, I think you'd need to mount the venv folder in the container for this to actually have any effect, no? Maybe we'd have to amend the venv location to live inside the project directory using the [right config option](https://python-poetry.org/docs/configuration/#virtualenvsin-project)?

laurentS (approved):

api/docker-compose.yml
ericboucher: It seemed to have worked when I used it. But good call, I'll move this to .develop

Issue #946, opened 2023-10-31
[Feature Request]: Add a direct link to PRISM DMP from the field report layer


ericboucher:
### Provide a clear and concise description of what you want to happen.

When a user selects a field report layer (kobo layer) in Cambodia, admin level data is visualized in the map. When a user clicks on the map, they see a few attributes of the field report which have been configured in layers.json. I want to also provide a direct link to the report directly in the DMP platform (temporarily hosted at https://dmp.ovio.org/). 

As an example, the screenshot / mockup in PRISM below shows a flood report on 2023-10-27. When I click on the area, I see details in the tooltip. 
1. I want to add a link here that says 'Report details' with a new window icon. 
2. Both the link and icon should open a new window that brings me to the form on DMP (https://dmp.ovio.org/form/FLOOD/496999294).

![dmp-link](https://github.com/WFP-VAM/prism-app/assets/3343536/e72ba059-1744-4089-8eb1-8f5f03e210db)

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

I'm not sure if we have a form submission identifier accessible through the PRISM API. If we can add it, we can use that to link to the DMP platform (https://dmp.ovio.org/form/FLOOD/496999294). 

If there isn't access to a unique ID, we could instead bring the user to the DMP form search page with the same parameters as the PRISM UI. In this example, we'd pass the admin IDs, report type, and date so you land on https://dmp.ovio.org/forms/search with these fields submitted:

<img width="1494" alt="Screenshot 2023-11-01 at 09 44 13" src="https://github.com/WFP-VAM/prism-app/assets/3343536/0c0a00e7-187d-4f68-8e88-7373ba452588">

PR #954, opened 2023-11-03
946 add redirect to DMP

jbperidypathtech:
This fixes issue #946.

How to test the feature:

- Get this picture
image

Screenshot/video of feature:
image
image

This PR also fixes a few edge cases for incident reports with no DisDate and only "start"
image

### DIFF
--- a/frontend/src/components/MapView/utils.ts
+++ b/frontend/src/components/MapView/utils.ts
@@ -6,6 +6,7 @@ import { LayerDefinitions } from 'config/utils';
 import { formatFeatureInfo } from 'utils/server-utils';
 import {
   AvailableDates,
+  DataType,
   FeatureInfoObject,
   FeatureInfoType,
   LayerType,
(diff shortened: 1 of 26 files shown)


frontend/src/components/MapView/utils.ts
```diff
@@ -87,29 +88,40 @@ export const downloadToFile = (
...
+              featureInfoProps[item].labelMap,
+            ),
+            coordinates,
+          },
+        };
+      },
+      {
+        dmpDisTyp: properties.DisTyp,
+        dmpSubmissionId: properties.submission_id,
+      },
```
ericboucher, line 123: This piece of code feels a bit too specific. Maybe wrap it around a util to handle metadata?
Remember that DMP and Kobo stuff are only used for Cambodia atm

jbperidypathtech: Indeed I just change it

frontend/src/config/cambodia/layers.json
```diff
@@ -1713,6 +1713,14 @@
       "status": {
         "type": "text",
         "dataTitle": "Form status"
+      },
+      "submission_id": {
```
ericboucher, line 1717: Maybe metadata should just keep their initial name and be a bit more lenient in the netadata type? Wdyt?
No strong feelings, open to suggestions.

cc @laurentS interested in your input as well

wadhwamatic:
This works well for flood and drought reports. After Eric's fix, the incident reports load without an error, but the hyperlink in the tooltip doesn't appear.

Screenshot 2023-11-03 at 11 51 22

frontend/src/config/cambodia/layers.json
jbperidypathtech: I transform submission_id in dmpSubmissionId to use camel case in the typescript code, and add also dmp prefix to be more specific about what is submission id. But if you rather like to use submission_id directly it is ok for me.

"be a bit more lenient in the metadata type" do you suggest to add a type to control and raise an error if it is a number rather than a string for instance? It is possible but it appears a little overkill for me. If required I can add this control.

jbperidypathtech:
For incident reports it was an edge case when DisTyp is both data (to display in popup) and metadata (to not display in popup) I changed a little bit the logic to cover this case.

frontend/src/config/cambodia/layers.json
ericboucher: No I actually meant "do not create a type for metadata and return all additional info that's not "data". This way people can simply decide to use the extra data in their logic, or not. :)
I am ok with the current implementation as well

ericboucher: So thinking about it a bit more we could actually revamp this a bit to also take into account the "do_not_display" behavior by making the dataTitle optional.

Then:
Option 1 - keys in `feature_info_props` would become data and all additional info would become metadata
Option 2 - keys in `feature_info_props` accept a "metadata" flag to split info between data and metadata like you're doing now.

Wdyt?

jbperidypathtech: I think option 2 is better to covert the case when we want both to display the data in the popup and use the data in the metadata while remaining generic. But I bring a small simplification in the layer.json to simplify metadata object.

ericboucher (approved):

frontend/src/config/types.ts
```diff
@@ -372,6 +372,7 @@ interface FeatureInfoProps {
   type: DataType;
   dataTitle: string;
   labelMap?: { [key: string]: string };
+  metadata?: string;
```
ericboucher, line 375: could this be types with the keys of `PopupMetaData`? This will make it easier to understand what to change in the future

Issue #952, opened 2023-11-01
[Bug]: Charts should remain top-aligned when we compare locations and periods


wadhwamatic:
### What happened?

Charts should always be top-aligned. This works as expected if you don't compare indicators / time periods.

### Which country / deployment are you running?

All

### Add a screenshot (if relevant)

Working correctly when using two indicators for the same location / period:
<img width="1404" alt="Screenshot 2023-11-01 at 10 14 25" src="https://github.com/WFP-VAM/prism-app/assets/3343536/81104087-097c-478c-8d69-484ff5b9be8c">

Not top-aligned when comparing indicators:
<img width="1499" alt="Screenshot 2023-11-01 at 10 15 30" src="https://github.com/WFP-VAM/prism-app/assets/3343536/48d559e1-a8ca-4673-ae89-927ecd8ae96a">

wadhwamatic:
Sorry @laurentS - I didn't catch this before

wadhwamatic:
This works as expected for more than one chart, but I just noticed that we're now reducing the height when we have a single chart. Here's what it looks like in Mozambique now:

<img width="1650" alt="Screenshot 2023-11-08 at 13 18 40" src="https://github.com/WFP-VAM/prism-app/assets/3343536/1a6173b6-14c4-4c39-b138-bb51c92036fa">

Whereas before it looked like this:
<img width="1649" alt="Screenshot 2023-11-08 at 13 30 19" src="https://github.com/WFP-VAM/prism-app/assets/3343536/6287f81c-f095-4f2d-bd3e-08af52744295">

Sorry for not catching this before @laurentS !

wadhwamatic:
Closing this issue and created a new issue for the single chart bug: https://github.com/WFP-VAM/prism-app/issues/989

PR #955, opened 2023-11-03
Align charts to top when comparing

laurentS:
This fixes issue #952.

How to test the feature:

- [ ] In the charts panel, activate comparison (location or period)
- [ ] Select a layer to chart
- [ ] Charts should now be top aligned

Screenshot/video of feature:
image


laurentS:
I've also added a commit to fix another small bug I just noticed: it prevents more than a single layer from being charted when comparing locations/periods. If more than a layer is selected when activating a comparison, only the first layer will remain selected, any other one will be deselected.

laurentS:
Added one more small change upon @ericboucher's suggestion to show region name above charts when comparing time periods:
image

ericboucher (approved):

wadhwamatic (approved):
All of these changes look good, thanks @laurentS!

Issue #941, opened 2023-10-26
[Feature Request]: Display time-series chart data directly from map context


wadhwamatic:
### Provide a clear and concise description of what the problem is.

Charts have been separated from the map context. However, it's useful for users to also be able to see trends when they are viewing layers in the main map view. This helps to show trend data in context.

### Provide a clear and concise description of what you want to happen.

If a user has selected a layer with a chart configuration, they should see an option to view a chart when they click on an admin area. There might be more than one admin area selectable (admin 1 or admin 2). This should be clear in the UI through direct links to the available areas for the selected layer.

See mockups below.

![Chart via map layer-01](https://github.com/WFP-VAM/prism-app/assets/3343536/056e94e5-40c5-4898-9dbf-7252c19b371f)

![Chart via map layer-02](https://github.com/WFP-VAM/prism-app/assets/3343536/a4cea17d-a0d7-4c2d-8a75-aab05e661f71)

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

We can limit this feature to 12 months of data. We might consider a direct link to the full featured chart module

Question:
- how easy would it be to add a link to to the chart module?

Issue #778, opened 2023-03-28
[Feature Request]: Improvement and merging overall CSV creation and data manipulation


HarryMytilinaios:
### Provide a clear and concise description of what the problem is.

Right now thoughout the application, we have different downloadCsv functions (for different parts of the project e.g `ChartsPanel`, `ExposureAnalysis`, `Analysis` .e.t.c), that have similar functionalities, and the only thing that is different is the data manipulation that every function does in order to download a CSV file with the correct data. Moreover it is missing from some of those the handling of the edge cases (e.g `undefined`, `null` data).

### Provide a clear and concise description of what you want to happen.

We need to have a generic abstracted function that will handle all the possible, different data provided and edge cases. We should consider adding a library for that in order to help us with the abstraction and the handling of the edge cases

PR #958, opened 2023-11-07
941 Display time-series chart data directly from map context; COUNTRY=cambodia

jbperidypathtech:
This fixes issue #941.

How to test the feature:

- On layers tab select a rainfall analysis
- Click on map to open tooltip, you should get list of available charts
- Click on a chart, it should display
- On the left menu if you select an other analysis, chart should update
- Into the popup when you click the download button, file should download

Screenshot/video of feature:
image
image
image


wadhwamatic:
Hi @jbperidypathtech. I tried to test this on the surge deployment (https://prism-958.surge.sh/) and locally but I don't see the option to view the charts for Mozambique in the tooltip. But if I test with Cambodia (like in your screenshots), it works as expected. But, when I tried locally for Sri Lanka, the app crashes when I click on the link in the tooltip:

Screenshot 2023-11-08 at 13 13 04

Screenshot 2023-11-08 at 13 13 12

Based on the working local version for Cambodia, we should remove the original location from the tooltip header when we view the chart as it's a bit misleading.

For example, in this screenshot, we should only see 'Kampong Thom' in the tooltip header since that is what is being shown in the chart.
Screenshot 2023-11-08 at 13 17 23

And in this screenshot, we should only see 'Kampong Thom, Santuk'.
Screenshot 2023-11-08 at 13 17 37

After closing the chart, we can revert to the full location name in the tooltip.
Screenshot 2023-11-08 at 13 23 15

wadhwamatic:
Thanks for the fixes @jbperidypathtech. This is looking good now. Over to @ericboucher for final review

ericboucher:
@jbperidypathtech nice work. I see that you merge parts of the CSV manipulation. Do you think it's enough to close https://github.com/WFP-VAM/prism-app/issues/778 ? Or are there a few more items worth exploring?

jperidy:
Hi @ericboucher,
I just add a small refacto to DownloadCsvButton (I didn't test it, I am in a train and connection is not good enough).
I think, we can still merge "downloadToFile" in components/MapView/utils.ts with DownloadCsvButton to end #778 task.
If I have time I can try this week elsewhere next week.

frontend/src/components/MapView/LeftPanel/ChartsPanel/index.tsx
```diff
@@ -887,12 +751,33 @@ const ChartsPanel = memo(
...
-          ])}
+        <DownloadCsvButton
+          filesData={[
+            {
+              fileName: buildCsvFileName([
+                multiCountry ? admin0Key : country,
+                selectedAdmin1Area ?? '',
+                selectedAdmin2Area ?? '',
+                ...(selectedLayerTitles as string[]),
+                comparePeriods ? 'first_period' : '',
```
ericboucher, line 762: Let's do that before hand and keep telling names such as `firstCSVFilename`  `secondCSVFilename`

ericboucher: alternativelt, you could create a utility function to generate the filename from params, but this bloc of logic (x2) is not great

ericboucher:
@jbperidypathtech great work. Sorry, I did not notice this before but some of this code seems duplicative of the `DataViewer` component used for Early Warning in Cambodia. Can you have a quick look? Let's see if it's a good idea to use your new component and leverage the DownloadCSV functionality.

frontend/src/components/MapView/LeftPanel/ChartsPanel/index.tsx
ericboucher: True haha, I meant that outside of a function called `generateCSVFilename` it feels a bit confusing.

Maybe we could simply use named arguments in buildCSVFilename and handle the `?? ''` in there.

wadhwamatic:
@jbperidypathtech - I just noticed that we have the text 'No chart available' appear for all types of layers, even if there isn't a chart configuration associated with it. Can we remove this text please? I don't think we need it in any case, simply don't show the chart link if no chart data is available.

Here's an example of it displaying out of context for Cambodia:
Screenshot 2023-11-14 at 17 52 34

ericboucher:
@wadhwamatic done

wadhwamatic (approved):
GTG, thanks!

wadhwamatic:
@jbperidypathtech @ericboucher - For RBD, the admin areas available in the tooltip are limited at admin 0 and admin 1. We should also be able to access admin 2. 

In this example, 'Tarka' chart data should also be accessible:
Screenshot 2023-11-16 at 09 32 18

frontend/src/components/MapView/MapTooltip/PopupCharts.tsx
```diff
@@ -0,0 +1,246 @@
...
+    | LayerData<BoundaryLayerProps>
+    | undefined;
+  const { data } = boundaryLayerData || {};
+  const mapState = useSelector(layersSelector);
+  const dataForCsv = useRef<{ [key: string]: any[] }>({});
+  const [adminLevel, setAdminLevel] = useState<AdminLevel | undefined>(
+    undefined,
+  );
+
+  const adminLevelsNames = useCallback(() => {
```
ericboucher, line 120: The whole logic around admin level names and admin level selection is quite confusing and error prone.

- It took me a while to fix the code and make it work well for multi-country deployments without breaking other deploys.

Don't we have a better way than splitting the locationName strings to get the admin names?

frontend/src/components/MapView/MapTooltip/PopupCharts.tsx
```diff
@@ -0,0 +1,246 @@
...
+    const adminLevelLimit =
+      adminLevel === undefined
+        ? MAX_ADMIN_LEVEL
+        : adminLevel + (multiCountry ? 1 : 0);
+    // If adminLevel is undefined, return the whole array
+    // eslint-disable-next-line fp/no-mutating-methods
+    return splitNames.splice(0, adminLevelLimit);
+  }, [adminLevel, popup.locationLocalName, popup.locationName]);
+
+  const startDate1 = new Date().getTime() - oneYearInMs;
```
ericboucher, line 135: no need for the `1` suffix

frontend/src/components/MapView/MapTooltip/PopupCharts.tsx
```diff
@@ -0,0 +1,246 @@
...
+    }
+  }, [adminLevel, adminLevelsNames, setPopupTitle]);
+
+  if (filteredChartLayers.length === 0) {
+    return null;
+  }
+
+  return (
+    <>
+      {adminLevel === undefined ? (
```
ericboucher, line 162: Using the same component with a ternary feels a bit odd. This means that when you click on the chart, the other data sets that are loaded are still displayed:

Screenshot 2023-11-16 at 8 26 38 PM

Screenshot 2023-11-16 at 8 26 33 PM

@wadhwamatic is that an ok behavior or should other data be hidden?

frontend/src/components/MapView/MapTooltip/PopupCharts.tsx
```diff
@@ -0,0 +1,237 @@
...
+              <Button
+                key={level}
+                variant="text"
+                size="small"
+                className={classes.selectLevelButton}
+                onClick={() => setAdminLevel((index + 1) as 0 | 1 | 2)}
+              >
+                <div className={classes.selectLevelButtonValue}>
+                  <div className={classes.selectLevelButtonText}>
+                    View {level} {layer.title} chart
```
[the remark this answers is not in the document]
wadhwamatic: Yes, translation is not rendered even though it exists in translation.json:
Screenshot 2023-11-16 at 17 47 29

wadhwamatic (changes requested):
Please fix missing translation (see https://github.com/WFP-VAM/prism-app/pull/958#discussion_r1396579333)

frontend/src/components/MapView/MapTooltip/PopupCharts.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+import ChartSection from '../LeftPanel/ChartsPanel/ChartSection';
+import { oneYearInMs } from '../LeftPanel/utils';
+import DownloadCsvButton from '../DownloadCsvButton';
+import { buildCsvFileName } from '../utils';
+
+const chartLayers = getWMSLayersWithChart();
+const { countryAdmin0Id, country, multiCountry } = appConfig;
+const MAX_ADMIN_LEVEL = appConfig.multiCountry ? 3 : 2;
+const admin0Offset = multiCountry ? 0 : 1;
```
ericboucher, line 35: It might make more sense to use something like this instead of these two variables? wdyt @jbperidypathtech ?
`const availableAdminLevels = multiCountry ? [0, 1, 2] : [1, 2];`

frontend/src/components/MapView/MapTooltip/PopupCharts.tsx
```diff
@@ -0,0 +1,246 @@
...
+const boundaryLayer = getBoundaryLayersByAdminLevel(MAX_ADMIN_LEVEL);
```
ericboucher, line 36: You probably don't need to use the MAX_ADMIN_LEVEL here. Just using it without arg should get the admin boundary with the most admin levels

frontend/src/components/MapView/MapTooltip/PopupCharts.tsx
laurentS: Based on what [I did in #974](https://github.com/WFP-VAM/prism-app/pull/974/files#diff-cb442affc6c2fb89fbf76cbc5bc40b49d94d3d6e9978824ac6454edb3ce5b2a0R182) (which is itself copy/pasted from the previous iteration of code), I think you could just do:
```javascript
const splitNames = isEnglishLanguageSelected(i18n)
      ? boundaryLayer.locationName
      : boundaryLayer.locationLocalName;
```

We seem to have code like this copy/pasted across a number of components.

frontend/src/components/MapView/MapTooltip/PopupCharts.tsx
```diff
@@ -0,0 +1,246 @@
...
+      width: '400px',
+      flexGrow: 1,
+    },
+  });
+
+interface PopupChartProps extends WithStyles<typeof styles> {
+  popup: MapTooltipState;
+  setPopupTitle: React.Dispatch<React.SetStateAction<string>>;
+}
+type AdminLevel = 0 | 1 | 2;
```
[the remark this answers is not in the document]
ericboucher: with the caveat that it should be limited to 2

frontend/src/components/MapView/LeftPanel/ChartsPanel/index.tsx
jbperidypathtech: I think we can simply remove the `?? ''` because we already filter null, undefined, 0, empty string data in buildCSVFilename. And ok to create two constants firstCSVFilename and secondCSVFilename before calling the component. It is more readable. I do that.

jbperidypathtech: I think we can simply remove the `?? ''` because we already filter null, undefined, 0, empty string data in buildCSVFilename. And ok to create two constants firstCSVFilename and secondCSVFilename before calling the component. It is more readable. I do that.

frontend/src/components/MapView/MapTooltip/PopupCharts.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+const PopupChart = ({ popup, setPopupTitle, classes }: PopupChartProps) => {
```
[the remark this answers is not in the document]
jbperidypathtech: Indeed. I rather like to rename the component, because in this component logic you can have several charts.

ericboucher (approved):

ericboucher (approved):

PR #966, opened 2023-11-10
Add background color on tick marks in the timeline UI

echaidemenos:
This fixes issue https://github.com/WFP-VAM/prism-app/issues/890.

How to test the feature:

- check timeline

Screenshot/video of feature:
let me know which of the following we would prefer:

image

image


ericboucher:
Thanks @echaidemenos! 
I think I prefer option 2 with just the ticks but option 1 looks great too.

wadhwamatic:
I also vote for option 2. Thanks @echaidemenos !

ericboucher (approved):

PR #972, opened 2023-11-13
Add controls to map "print" functionality

wadhwamatic:
This fixes issue https://github.com/WFP-VAM/prism-app/issues/780.

How to test the feature:
- Select a layer
- Click on the print icon
- See legend and footer text as map options

Screenshot/video of feature:
Screenshot from 2023-11-17 11-45-57
Screenshot from 2023-11-21 16-28-32
Screenshot from 2023-11-21 16-28-41


ericboucher:
@echaidemenos nice work, a few small comments:
- the date doesnt seem to update in the print view, it shows `--date--``
- the scale seems to be moving around. Sometimes it disappears a bit and sometimes it is "floating" above the bottom

echaidemenos:
> @echaidemenos nice work, a few small comments:
> 
>     * the date doesnt seem to update in the print view, it shows `--date--``
> 
>     * the scale seems to be moving around. Sometimes it disappears a bit and sometimes it is "floating" above the bottom
> 
> 
> Screenshot 2023-11-17 at 6 11 45 PM

@ericboucher I fixed the issues above. Let me know if anything else comes up

ericboucher (approved):

wadhwamatic:
Nice work. It's nearly there but I have a few small requests:
1. Please change the order of the map elements toL Legend, Full layer description, Footer text
3. In the footer text, "Layers represent data on " is a bit misleading as the text refers to the map publication data, not the layer date. We should change it to say "Map produced on " and we should also have this text editable so a user can reword it / translate it as needed
4. The list of sources in the footer is incorrect. Let's remove this for now. In the future, we should have a source attribute in the layer config which could then be shown here dynamically based on the layers in view

wadhwamatic:
@ericboucher - I see one small issue. If I edit the footer text and then click download, I won't see the changes as I have to click 'refresh image' for any changes to be saved. That's not intuitive. Can we re-render the map after a user clicks 'ok' if the footer text was changed?

frontend/src/components/NavBar/PrintImage/__snapshots__/index.test.tsx.snap
```diff
@@ -200,11 +200,16 @@ exports[`renders as expected 1`] = `
         inputprops="[object Object]"
         maxrows="4"
         multiline="true"
-        value="Layers represent data on 2023-11-22.  The designations employed and the presentation of material in the map(s) do not imply the expression of any opinion on the part of WFP concerning the legal of constitutional status of any country, territory, city, or sea, or concerning the delimitation of its frontiers or boundaries."
+        value="Layers represent data on 2023-11-23.  The designations employed and the presentation of material in the map(s) do not imply the expression of any opinion on the part of WFP concerning the legal of constitutional status of any country, territory, city, or sea, or concerning the delimitation of its frontiers or boundaries."
```
ericboucher, line 203: @echaidemenos let's fix the date in tests so that it doesn't change every day no?

PR #973, opened 2023-11-14
Improve failed Kobo authentication behavior in the frontend; COUNTRY=cambodia

echaidemenos:
This fixes issue https://github.com/WFP-VAM/prism-app/issues/560 and https://github.com/WFP-VAM/prism-app/issues/910.

How to test the feature:

- use country `cambodia` 
- from field reports use flood report
- enter incorrect kobo credentials
- see that the auth modal is still showing and "Authentication failed" message appears
- the layer is turned off when cancel is click on the auth modal

Screenshot/video of feature:
Screenshot from 2023-11-14 10-23-44


ericboucher (approved):

frontend/src/utils/error-utils.ts
```diff
@@ -10,3 +10,12 @@ export class LocalError extends Error {
     return this.message;
   }
 }
+
+export class HTTPError extends Error {
+  constructor(
+    public readonly message: string,
+    public readonly statusCode: number,
+  ) {
+    super(message);
```
ericboucher, line 19: Maybe we should add
`this.name = name;` or similar?

wadhwamatic:
This looks good. Thanks for the fix @echaidemenos

ericboucher (approved):

Issue #749, opened 2023-03-09
Deprecate `Layers/BoundaryDropdown` in favor of `Common/BoundaryDropdown`


ericboucher:
### Provide a clear and concise description of what the problem is.

We have duplicate files to create boundary dropdowns. Esp. since #628.

We should consolidate them and leverage `Common/BoundaryDropdown` for GoTo AND Alerts.

### Provide a clear and concise description of what you want to happen.

I would like to have a simplified approach to `BoundaryDropdown` and only one file with the name `BoundaryDropdown.tsx`

#### GoTo, works for admin 3 (using `Common/BoundaryDropdown`)
<img width="1447" alt="Screenshot 2023-03-09 at 5 15 45 PM" src="https://user-images.githubusercontent.com/16843267/224085448-2ec95169-66c5-45ac-b5a5-9d260acca1df.png">

#### Alerts, broken for admin 3 (using `Layers/BoundaryDropdown`)
<img width="868" alt="Screenshot 2023-03-09 at 5 16 05 PM" src="https://user-images.githubusercontent.com/16843267/224085467-74feae39-d8cb-47bb-b90e-74bc94d11a1e.png">

Issue #937, opened 2023-10-19
[Bug]: alerting area dropdown does not seem to work for Cambodia


laurentS:
### What happened?

The alerting menu includes a region selection dropdown which does not always show region names correctly. For instance, the screenshot below for Cambodia shows multiple regions under the same name (I picked 4 visible on the map to confirm they are indeed different areas).

![image](https://github.com/WFP-VAM/prism-app/assets/1745184/55c864dc-c133-487d-95a4-dc457fe7db71)

### Which country / deployment are you running?

This is for Cambodia, I have only checked that Mozambique works ok, I don't know about other countries.

### Add a screenshot (if relevant)

This menu, to avoid confusion:
![image](https://github.com/WFP-VAM/prism-app/assets/1745184/ff5da182-b1a4-471a-a1b3-8cddd6671f5e)

wadhwamatic:
Thanks for catching this. The alert feature hasn't been deployed so I'm not surprised there are issues. 

@ericboucher - we had talked about trying to unify the location list approach between the go to feature and others. If the go to feature has been implemented well, can we also utilize it here?

laurentS:
I think the underlying issue I'm seeing in various parts of the code related to selecting a location is that the shape of the geo data (admin areas level 0, 1, 2, names...) does not seem to be standardised across countries, which leads to the code in these dropdowns (and things like displaying the name of the location in a localised way) being overly complicated. I wonder if preprocesssing this data into one common shape could help simplify the code in a bunch of places?

wadhwamatic:
@laurentS - I'm not sure what preprocessing into one common shape would entail. It sounds like you're suggesting a centrally managed single set of admin boundaries. That would usually work for WFP managed deployments down to admin 2, but I think would force some common definition of boundaries that is problematic for a long list of reasons and gets messy when we start getting to admin 3 level data. @ericboucher - sounds like we should think further on this to find a longer term solution. Let me know when you'd like to discuss

laurentS:
Related to #749

PR #974, opened 2023-11-15
Refactor area dropdowns into a single component; COUNTRY=rbd

laurentS:
This fixes issue #749 and #937.

How to test the feature:

- [ ] Set `alertFormActive: true` in `config/<country>/prism.json`
- [ ] Open the "GO TO" dropdown and pick any area. The map should pan/zoom to it, no matter what level
- [ ] Open the "create alert" menu, and the "regions" drop down. Its content should be the same as the previous dropdown. Any area/subarea can be selected. The map updates as selections are made, and conversely when updating selection on the map directly.

### Notes
- items in the menu are sorted in alphabetical order of their name (should be localized to the selected language, but I forgot the khmer alphabet :sweat_smile:)
- Menus appear to work ok with 1 (Afghanistan), 2 (Mozambique) or 3 (cambodia) admin levels, as well as multi-country
- Formatting in the menu could be improved. Right now it's based on relative depth compared to root level, instead of actual admin level (so in multi country, country name is uppercase, but in single country, it's admin level 1)
- I've used `<MenuItem>` everywhere, which appears to make menus faster, but for RBD in particular, they're still sluggish. I can try to figure out why if you think it's worth the effort.

### outstanding points
- [x] improve typing in some places
- [x] cleanup / remove disused code
- [x] move the 3 level location selector in Charts panel to use the new code

# Screenshot/video of feature:
## Jordan
image
and in Arabic
image

## Cambodia
image
image

## RBD
image


frontend/src/components/Common/BoundaryDropdown/utils.ts
```diff
@@ -2,18 +2,26 @@ import type { Feature, MultiPolygon, BBox } from '@turf/helpers';
...
   levels: number[];
   relations: BoundaryRelation[];
 };
 
 export type BoundaryRelation = {
   name: string;
+  adminCode: AdminCodeString;
   parent: string;
-  level: number;
+  level: number; // 0-indexed, so not compatible with AdminLevelType
```
ericboucher, line 24: AdminLevelType is now 0 indexed as well.

frontend/src/components/MapView/Layers/BoundaryDropdown.tsx
```diff
@@ -3,23 +3,29 @@ import {
   FormControl,
   InputAdornment,
   InputLabel,
-  ListSubheader,
+  // ListSubheader,
```
ericboucher, line 6: 👣

ericboucher (approved):

frontend/public/data/rbd/wca_admbnda_adm2_ocha.json
```diff
@@ -791,7 +791,7 @@
...
...92, 6.263605358 ], [ -4.463962845, 6.26247451 ], [ -4.463324786, 6.261635827 ], [ -4.462811749, 6.260764575 ], [ -4.462312957, 6.259733331 ], [ -4.461602241, 6.257951512 ], [ -4.46117784, 6.256554368 ], [ -4.460911886, 6.255159605 ], [ -4.460595199, 6.254005201 ], [ -4.459232966, 6.252728222 ], [ -4.457603043, 6.251870238 ], [ -4.45559599, 6.251889366 ], [ -4.453299143, 6.252304371 ], [ -4.451467526, 6.253165165 ], [ -4.449836791, 6.253547795 ], [ -4.448413037, 6.253715633 ], [ -4.446999575, 6.253929789 ], [ -4.445811155, 6.254744708 ], [ -4.444398388, 6.255521806 ], [ -4.442901666, 6.255963961 ], [ -4.441277088, 6.256024378 ], [ -4.439440594, 6.255944069 ], [ -4.438009071, 6.255342081 ], [ -4.43672705, 6.254571246 ], [ -4.435704589, 6.25380535 ], [ -4.434538653, 6.253450188 ], [ -4.433474547, 6.253671266 ], [ -4.43222133, 6.254347858 ], [ -4.430780286, 6.255433608 ], [ -4.429170499, 6.256470643 ], [ -4.427789102, 6.257374226 ], [ -4.426146448, 6.258376864 ], [ -4.425361446, 6.258839649 ], [ -4.424618456, 6.259277449 ], [ -4.423086972, 6.260350433 ], [ -4.421922625, 6.261109285 ], [ -4.421020724, 6.261688817 ], [ -4.420264905, 6.262294517 ], [ -4.418398219, 6.263387614 ] ] ] ] } },
```
ericboucher, line 794: @laurentS what changed here?

ericboucher: @wadhwamatic ^

wadhwamatic: Thanks for catching and fixing this @laurentS. I see that the error is in the original source boundary file. I'll let the team in Dakar know. The fix you made is correct.

I see others that should be fixed too: CI1302, CI1303, CI1304

wadhwamatic:
Great work @laurentS! I haven't found any issues. We might hear some feedback on the Khmer and Arabic versions but can open issues as needed if something is reported

wadhwamatic (approved):

frontend/public/data/rbd/wca_admbnda_adm2_ocha.json
ericboucher: Indeed, and we need to update the links for the ones we change in the admin data files.

ericboucher: @wadhwamatic I fixed the others. However for 01 and 02 we have a problem because there are two data info with the same keys in the admin data level datasets and we don't have a way to know which datapoint is for which region. In https://github.com/WFP-VAM/prism-app/pull/974/commits/d231af6af4e716f6a0d5553b1380440dac26c22b I was able to fix the ica and multi risk datasets for 03 and 04 but not for 01 and 02

laurentS:
> Great work @laurentS! I haven't found any issues. We might hear some feedback on the Khmer and Arabic versions but can open issues as needed if something is reported

yes. I anticipate there might be issues with sorting of names in those alphabets, I don't know how valid the javascript `sort()` is for non-latin alphabets. But it would be an easy fix if we have some reference of how to sort.

Issue #901, opened 2023-07-11
Remove some usage of useMemo and useCallback


JulienPathtech:
### Provide a clear and concise description of what the problem is.

There is a lot of (if not always) `useMemo` / `useCallback` in the code. This make the code less readable and can cause performance overhead in some situation

### Provide a clear and concise description of what you want to happen.

Remove usage of memoization when the memoized code is `light` or need small computation time. According to this [article](https://medium.com/credera-engineering/when-should-usememo-and-usecallback-be-used-and-when-not-complete-with-examples-df13cd7a6cf4#:~:text=Every%20pro%20has%20a%20con,you'd%20expect%20it%20to.),there is a performance overhead : 

This will : 
- Improve code readability
- Improve performance

PR #979, opened 2023-11-20
Add date range slider for charts

echaidemenos:
fixes https://github.com/WFP-VAM/prism-app/issues/965.
fixes https://github.com/WFP-VAM/prism-app/issues/989
fixes https://github.com/WFP-VAM/prism-app/issues/739

How to test the feature:

- navigate to charts tab
- enable some chart or charts
- use the slider 

Screenshot/video of feature:

Screenshot from 2023-12-04 16-41-59
Screenshot from 2023-12-04 16-42-22


ericboucher:
Let's handle https://github.com/WFP-VAM/prism-app/issues/952 as part of this PR @echaidemenos

frontend/src/context/leftPanelStateSlice.ts
```diff
@@ -1,8 +1,14 @@
 import { createSlice, PayloadAction } from '@reduxjs/toolkit';
 import type { RootState } from './store';
 
+export enum Panel {
+  Layers = 0,
```
ericboucher, line 5: we could even use explicit string here no? or are the numbers useful?
Layers = "layers", ...

frontend/src/components/MapView/LeftPanel/ChartsPanel/index.tsx
```diff
@@ -514,43 +580,91 @@ const ChartsPanel = memo(
...
       t,
+      secondAdmin0Key,
     ]);
 
     useEffect(() => {
       if (showChartsPanel) {
         setPanelSize(PanelSize.xlarge);
         setResultsPage(
-          <Box className={classes.chartsPanelCharts}>{renderResultsPage}</Box>,
+          <Box className={classes.chartsContainer}>
```
ericboucher, line 611: It could be worth creating a "DateRangeSlider" component

frontend/src/components/MapView/LeftPanel/ChartsPanel/index.tsx
```diff
@@ -336,11 +387,16 @@ const ChartsPanel = memo(
...
+              height: '50vh',
               width: '100%',
             }}
           >
             <ChartSection
+              setChartSelectedDateRange={setChartSelectedDateRange}
+              setMaxDataTicks={setMaxDataTicks}
+              chartDateRange={chartDateRange}
+              setChartDateRange={setChartDateRange}
+              chartRange={comparePeriods ? undefined : chartRange}
```
ericboucher, line 399: do we need to pass in all these?

frontend/src/components/MapView/LeftPanel/ChartsPanel/index.tsx
```diff
@@ -514,43 +580,91 @@ const ChartsPanel = memo(
       ) : null;
       return [locationBox, ...titles, ...zipped];
     }, [
-      admin0Key,
-      adminLevel,
       adminProperties,
-      classes.textLabel,
+      startDate1,
```
ericboucher, line 584: is there a reason for the re-order?

echaidemenos: I am pretty sure it was done automatically. I am reordering them.

frontend/src/components/MapView/LeftPanel/ChartsPanel/index.tsx
echaidemenos: - `chartRange` is the output of the slider, setting what appears on the chart.
- `chartMaxDateRange` and `setChartMaxDateRange` are about knowing which dataset has the largest available data range, so we can extend the others to match it.
- `setMaxDataTicks` is for setting the sliders ticks
- `setChartSelectedDateRange` is for getting the the dates which the slider has selected, so we can show then on the label. This is necessary, since we don't have the dates in this component, the datasets hold the date-value objects

frontend/src/components/MapView/LeftPanel/AnalysisPanel/index.tsx
```diff
@@ -112,7 +113,7 @@ import ExposureAnalysisTable from './AnalysisTable/ExposureAnalysisTable';
 import ExposureAnalysisActions from './ExposureAnalysisActions';
 import { getExposureAnalysisTableData } from '../../utils';
 
-const tabIndex = 2;
+const tabIndex = Panel.Analysis;
```
ericboucher, line 116: To make it clearer it might be better to change from "index" to "panelId" or something like that? `index` is used in a few places

wadhwamatic:
@echaidemenos - the functionality for this feature works great, nice work. I feel like some minor UI changes would make it a bit more polished. @ericboucher - did we get any feedback from Marc on this?

frontend/src/components/MapView/LeftPanel/ChartsPanel/index.tsx
laurentS: Does this question suggest some comments would be helpful? :wink:

frontend/src/components/Common/Chart/index.tsx
```diff
@@ -207,11 +210,23 @@ const Chart = memo(
...
+      }));
+
       return {
         labels,
-        datasets: datasetsWithThresholds,
+        datasets: datasetsTrimmed,
       };
-    }, [EWSthresholds, indicesDataSet, labels, tableRowsDataSet, transpose]);
+    }, [
+      EWSthresholds,
```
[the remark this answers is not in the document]
echaidemenos: i will be removing the `useMemo` here. From the article referenced in #901 the benefits of `useMemo` starts to kick in after 1000 elements on a sort algorithm. Here we have less than 200 elements and slicing (O(n))

frontend/src/components/MapView/LeftPanel/ChartsPanel/index.tsx
```diff
@@ -251,6 +259,40 @@ const ChartsPanel = memo(
...
+    >(['', '']);
+    const [showSlider, setShowSlider] = useState(true);
+
+    // Reset slider when new charts selection is changed
+    useEffect(() => {
+      setMaxDataTicks(0);
+      setChartRange([0, 0]);
+      setChartSelectedDateRange(['', '']);
+      setChartMaxDateRange(['', '']);
+    }, [selectedLayerTitles]);
```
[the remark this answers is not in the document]
echaidemenos: yes this "resets" slider specific state when charts selection changes, as the comment above describes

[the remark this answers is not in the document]
laurentS: gosh, I find this so confusing...

api/app/report.py
```diff
@@ -80,7 +80,9 @@ async def mock_prism_api_stats_call(route):
         await click_target_exposure_analysis(page, layerIdParam)
 
         # Wait for page to be loaded on exposure analysis
-        await page.wait_for_selector('div[id="full-width-tabpanel-2"]', state="visible")
+        await page.wait_for_selector(
```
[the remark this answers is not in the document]
ericboucher: Not that I can think of at the moment. We could imagine opening the tab through a "url trigger"

ericboucher:
> We lost the ability to go back further in time. I think it's ok to limit how far back you can slide back, but we should keep the date input field so a user can specify a date of interest. Hope that's clear and agreed upon.


@wadhwamatic  - I just pinged Marc. We didn't lose the ability to go back in time since we're grabbing all the available data for all datasets and use the max range. So I think it's a bit less confusing this way personally instead of manipulating dates from two different places which will likely lead to edge cases

wadhwamatic:
@ericboucher - ok, got it regarding date fetching. As soon as we add a limit, someone will ask to make it longer. Can we start with 20 years of data?

wadhwamatic:
Thanks @ericboucher and @echaidemenos. The bar and handles feel disproportionately large to me. Can we reduce the height of the bar/handles by 25%, and the width of the handles by 50%?

The text 'start' and 'end' above the slider bar look too close to the bar. Can also increase that spacing a bit? This might already get fixed by reducing the bar height.

api/app/geotiff_from_stac_api.py
```diff
@@ -31,7 +31,10 @@
 
 @timed
 def generate_geotiff_from_stac_api(
-    collection: str, bbox: [float, float, float, float], date: str, band: str
+    collection: str,
+    bbox: [float, float, float, float],
+    date: str,
+    band: str | None = None,
```
ericboucher, line 37: @laurentS looks better like this?

frontend/src/components/MapView/LeftPanel/ChartsPanel/DateSlider/index.tsx
```diff
@@ -0,0 +1,78 @@
+import { Box, Typography, createStyles, makeStyles } from '@material-ui/core';
+import React from 'react';
+import RangeSlider from 'react-range-slider-input';
+import 'react-range-slider-input/dist/style.css';
+import './slider.css';
+
+const useStyles = makeStyles(() =>
+  createStyles({
+    sliderContainer: {
+      width: 'calc(100% - 4rem)',
```
ericboucher, line 10: Is there a reason you're using a calc instead of setting it to 100% and having 2rem margins on each side?

ericboucher (approved):

Issue #980, opened 2023-11-21
Merge EWS Data Viewer into PopupChart


ericboucher:
### Provide a clear and concise description of what you want to happen.

The DataViewer is currently used to display EWS charts for Cambodia.
We would like to build on the work achieved in #958  to move the EWS logic to the tooltip PopupChart as well.
We should then also enable the `Download` button for EWS data.

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

Remove the DataViewer component once unused.

PR #983, opened 2023-11-22
Merge dataviewer into popupchart; COUNTRY=cambodia

jbperidypathtech:
This fixes issue #980.

How to test the feature:

- Select Early Warning and Rainfall Amount layers
- Click on map area
- Navigate in charts
- Download datas

Screenshot/video of feature:
image
image


wadhwamatic:
I like that we're consolidating these features into a single component, but we're now also adding an additional click for users to see the river level trend data. 

For other chart data, an intermediary step is needed to determine which admin level a user wants to visualize. For point data, that condition doesn't exist. 

Can we consider going straight to chart view for point data?

jbperidypathtech:
Hi @wadhwamatic,
I change the behavior. Let me know if it is better for your now like this.
I will have a check tomorrow, It seems there is a board effect when you use Early Warning layer with others layers. I notice popup details information were not always displayed... Let me know if you have also this behavior.

ericboucher:
@jbperidypathtech the latest changes are a huge regression for me in terms of UI. I have a lot of bugs with the tooltip:
- sometimes I just see a floating "tip" / corner
- the tooltips are placed on the side instead of on the top
- the "tip" is white with a dark square around it

Screenshot 2023-11-23 at 11 38 14 PM
Screenshot 2023-11-23 at 11 46 44 PM
Screenshot 2023-11-23 at 11 47 10 PM

jbperidypathtech:
@ericboucher I have the same, it works normally locally but I don't understand why it is not working on surge... It seems something not working right on mapboxgl lib.
I start new deploy with reinstallation of dependencies, but no success.

frontend/src/components/MapView/MapTooltip/index.tsx
```diff
@@ -117,6 +55,99 @@ const styles = () =>
...
+        <PopupPointDataChart adminLevelsNames={() => [...adminLevelsNames()]} />
+      </Popup>
+    );
+  }
+
+  return (
+    <Popup
+      coordinates={popup.coordinates}
+      className={classes.popup}
+      style={{ zIndex: 5 }}
```
ericboucher, line 127: you can probably just add this to the popup class

frontend/src/components/MapView/MapTooltip/index.tsx
```diff
@@ -117,6 +55,99 @@ const styles = () =>
...
+      ? popup.locationName
+      : popup.locationLocalName;
+    const splitNames = locationName.split(', ');
+
+    const adminLevelLimit =
+      adminLevel === undefined
+        ? availableAdminLevels.length
+        : adminLevel + (multiCountry ? 1 : 0);
+    // If adminLevel is undefined, return the whole array
+    // eslint-disable-next-line fp/no-mutating-methods
```
ericboucher, line 103: some of this logic seems related to RedirectToDMP only, could it be moved in there?

wadhwamatic:
This is looking good, thanks. Great to have download capabilities for the EWS data now. Can we add the date to the timestamp column?

frontend/src/components/MapView/MapTooltip/index.tsx
jbperidypathtech: I didn't managed to do that. When I tried it was overided by native component class (.MapTooltip-popup-195). I tried several solution but only this one worked. If you see a way to achieved that let me know :).
image

ericboucher: even with !important?

jbperidypathtech:
Hi @wadhwamatic date should be available now :)

wadhwamatic:
Thanks for adding the dates, @jbperidypathtech. One last input from me - it seems hovering over the chart for EWS-1294 is inconsistent. I can sometimes see appear on hover, but not always. See the video below. I only see this on the EWS-1294 layer, not on other charts in the tooltip.

https://github.com/WFP-VAM/prism-app/assets/3343536/f72626ba-78b8-4413-8688-8dc380038640

ericboucher (approved):

ericboucher:
@wadhwamatic I'm merging this PR for now but feel free to open an issue to tackle the chart hover problem. As @jbperidypathtech mentioned we might be able to fix this simply by downsampling the EWS data

PR #991, opened 2023-11-24
Hotfix charts dropdown issue

ericboucher:
The admin1 dropdown was not populating properly for admin 1 in single country deployments

Screenshot 2023-11-24 at 2 54 15 PM


laurentS (approved):
Sorry I missed this. I was pretty sure I had tested over multiple countries. Looks like the PR template would have helped ;)

PR #993, opened 2023-11-27
Add admin level codes to all countries

ericboucher:
PR #974 made admin level codes mandatory for all layers.

@wadhwamatic @laurentS please have a look to make sure the selected admin codes are reasonable


wadhwamatic:
@ericboucher - Would this also address https://github.com/WFP-VAM/prism-app/issues/604?

wadhwamatic:
@ericboucher - I added a few changes. Looks good from my side

frontend/src/config/mongolia/layers.json
```diff
@@ -4,7 +4,7 @@
     "path": "data/mongolia/admin_boundaries.json",
     "opacity": 0.8,
     "admin_code": "NSO_CODE",
```
ericboucher, line 6: @wadhwamatic then this should likely be updated as well no?

laurentS: I took a guess when I copied `admin_code` into `admin_level_codes` here, but I was a little puzzled by the fact that the format of codes did not seem consistent with other countries (5 digits vs 6).

laurentS (approved):

PR #1001, opened 2023-11-30
Add commit hash text, hidden on the bottom of the tabs

### LINKED ISSUE #984: Add commit tag to the frontend and the API

Add commit tags to the API and the frontend to easily know what is currently deployed.
For example the API could have an endpoint that would return:
`{commit-hash: hash, commit-message: message}`

And the frontend could have a similar data made accessible, maybe hidden in the "about page" or as a standalone page. Thoughts on this @laurentS @echaidemenos since we did something similar on Ovio?

echaidemenos:
### Description

This fixes #984 

Adds hidden text at the bottom of the tabs, showing the commit hash of the code deployed.

## How to test the feature:

- navigate to the bottom eg. layers tab
- highlight the hidden text to see the commit hash. 
- confirm it is the same as the hash of the code of this deployment

## Checklist - did you ...

Test your changes with

Screenshot from 2023-11-30 15-36-59

### DIFF
--- a/frontend/config-overrides.js
+++ b/frontend/config-overrides.js
@@ -1,3 +1,4 @@
+/* eslint-disable fp/no-mutation */
 const {
   override,
   disableChunk,
@@ -8,6 +9,11 @@ const path = require('path');
 const RemovePlugin = require('remove-files-webpack-plugin');
 
 const country = process.env.REACT_APP_COUNTRY;
+
+// In case GIT_HASH is not set we are in github actions environment
+process.env.REACT_APP_GIT_HASH = process.env.GITHUB_SHA
+  ? process.env.GITHUB_SHA
+  : process.env.GIT_HASH;
 
 module.exports = override(
   useEslintRc(path.resolve(__dirname, '.eslintrc')),
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -16,9 +16,9 @@
   },
   "scripts": {
     "clean": "rimraf ./node_modules && pushd ../common && rimraf ./node_modules && popd",
-    "start": "cross-env EXTEND_ESLINT=true react-scripts start",
+    "start": "export GIT_HASH=$(git rev-parse --short HEAD) && cross-env EXTEND_ESLINT=true react-app-rewired start",
     "analyze": "yarn build && npx source-map-explorer 'build/static/js/*.js'",
-    "build": "cross-env EXTEND_ESLINT=true react-app-rewired build",
+    "build": "export GIT_HASH=$(git rev-parse --short HEAD) && cross-env EXTEND_ESLINT=true react-app-rewired build",
     "test": "cross-env REACT_APP_TESTING=true && react-scripts test",
     "prettier:json-fix": "prettier --write */**.json",
     "prettier:json-check": "prettier --check */**.json",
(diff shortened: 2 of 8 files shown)


frontend/config-overrides.js
```diff
@@ -9,6 +10,11 @@ const RemovePlugin = require('remove-files-webpack-plugin');
 
 const country = process.env.REACT_APP_COUNTRY;
 
+// In case GIT_HASH is not set we are in github actions environment
+process.env.REACT_APP_GIT_HASH = process.env.GITHUB_SHA
```
ericboucher, line 14: this could probably become a const instead of an env at this point no?

ericboucher:
Great start
- I think the text can be smaller and we can just show the short hash (8 first characters)
- Let's call it "version hash", @wadhwamatic does that work for you?
- Let's make it clickable to open up GitHub as discussed

frontend/package.json
```diff
@@ -16,9 +16,9 @@
   },
   "scripts": {
     "clean": "rimraf ./node_modules && pushd ../common && rimraf ./node_modules && popd",
-    "start": "cross-env EXTEND_ESLINT=true react-scripts start",
+    "start": "export GIT_HASH=$(git rev-parse --short HEAD) && cross-env EXTEND_ESLINT=true react-app-rewired start",
```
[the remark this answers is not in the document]
echaidemenos: added it on start command also, so the view is consistent between development and deployment

frontend/config-overrides.js
echaidemenos: I don't think this code is accessible from within the application. It run before build and sets up the environment/configuration. I believe we did the same on OVIO right?

frontend/src/components/Common/HashText/index.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
ericboucher, line 1: Let's add some comments at the top of the component

frontend/src/components/Common/HashText/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <p
+      style={{
+        color: '#f5f7f8',
+        position: 'absolute',
+        bottom: '0',
+        fontSize: '0.7rem',
+        left: '1rem',
+      }}
+    >
+      version hash:{' '}
```
ericboucher, line 15: let's make the whole text clicable. If no hash is provided, simply link to the repo

frontend/config-overrides.js
```diff
@@ -9,6 +10,12 @@ const RemovePlugin = require('remove-files-webpack-plugin');
 
 const country = process.env.REACT_APP_COUNTRY;
 
+// In case GIT_HASH is not set we are in github actions environment
+process.env.REACT_APP_GIT_HASH = (process.env.GITHUB_SHA
```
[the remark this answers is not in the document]
echaidemenos: `'' ?? '42'` would result in `''`. I am not sure if `process.env.GITHUB_SHA` can be `''`, but in that case we would prefer to use `process.env.GIT_HASH`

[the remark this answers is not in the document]
echaidemenos: Thinking it a bit better maybe `process.env.GITHUB_SHA || process.env.GIT_HASH` does the same :)

ericboucher (approved):

PR #1006, opened 2023-12-04
Optionally pass band info for STAC download; COUNTRY=zimbabwe

### LINKED ISSUE #1005: [Feature Request]: modify STAC download request to handle bands (aka assets)

The structure of blended precipitation products provided through HDC is different than other rainfall products. Most HDC rainfall products have a distinct layer for each time aggregation (i.e. 10-day, 1-month, 3-month, etc.). But HDC blended products have a single layer for rainfall aggregation and rainfall anomaly, each with their own set of styles for each time aggregation.

I need to be able to download a blended rainfall layer at any of the available time aggregations. To do this, we need to update the STAC download query to use the 'bands' (aka 'assets' in STAC terminology) available for each layer. 

For example, the data layer 'rfq_blended_zwe_dekad' has 10 bands available, one for each available time aggregation. If I wanted to access the 3-month anomaly, I would need to refer to the asset / band named 'r3q', which is seen here:
 https://data.earthobservation.vam.wfp.org/stac#/collections/rfq_blended_zwe_dekad/items/rfq_blended_zwe_dekad-202305d1?.asset=asset-r3q

Here's a screenshot of the assets associated with the latest blended rainfall anomaly for Zimbabwe
Screenshot 2023-12-04 at 17 44 50

Until this is implemented, the download button for blended precipitation products only gets the first band which in this case is r1q (1-month aggregation)

ericboucher:
### Description

This tentatively fixes #1005. By adding a "band" additional_query_parameter.

Unfortunately it doesn't seem super easy to infer the band name from the style name so I added an additional parameter for now. This won't work as is since it's not deployed to the API but you can give it a try locally.

The band info is now added to the filename if it is not simply "band".

## How to test the feature:

- [ ] check that you can download a geotiff for this layer using a band - https://prism-1006.surge.sh/?hazardLayerIds=precip_blended_3m&date=2023-11-11
- [ ] check that you can download a geotiff for this layer that does not use a band - https://prism-1006.surge.sh/?hazardLayerIds=rain_anomaly_2month&date=2023-11-11

## Checklist - did you ...

Test your changes with

### DIFF
--- a/api/app/geotiff_from_stac_api.py
+++ b/api/app/geotiff_from_stac_api.py
@@ -43,7 +43,9 @@ def generate_geotiff_from_stac_api(
     if not items:
         raise HTTPException(status_code=500, detail="Collection not found in stac API")
 
-    print(items[0].assets)
+    # TODO - confirm that input band exists
+    available_bands = items[0].assets.keys()
+    print(available_bands)
 
     bands = [band] if band else None
     try:
--- a/api/app/geotiff_from_stac_api.py
+++ b/api/app/geotiff_from_stac_api.py
@@ -34,7 +34,6 @@ def generate_geotiff_from_stac_api(
     collection: str, bbox: [float, float, float, float], date: str, band: str
 ) -> str:
     """Query the stac API with the params and generate a geotiff"""
-    file_path = f"{collection}_{date}_{str(uuid4())[:8]}.tif"
     catalog = Client.open(STAC_URL)
 
     query_answer = catalog.search(bbox=bbox, collections=[collection], datetime=[date])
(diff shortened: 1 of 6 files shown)


wadhwamatic:
@ericboucher - thanks for the quick fix on this. I see this working locally as expected. 
Screenshot 2023-12-05 at 05 31 47

Please let me know when you've got the backend setup as well and I'll test again. I'm not sure if the STAC parameter will be 'band' or 'asset', FYI.

Related to this, we had implemented a way to bypass adding layers to the analysis feature to prevent this issue from using the wrong data. See PR https://github.com/WFP-VAM/prism-app/pull/829
Can we add the 'band' parameter to the analysis request as well? I think for WCS requests, the parameter is 'RangeSubset'.
Even better, we should use STAC requests instead of WCS when we're accessing STAC-enabled data sources: https://github.com/WFP-VAM/prism-app/issues/644

I realize that's a lot. Let me know what's feasible in the short term.

wadhwamatic (approved):

PR #1030, opened 2023-12-15
Improve app reactivity; COUNTRY=rbd

### LINKED ISSUE #986: Performance improvements
### What happened?

The app is quite slow for large deployments like RBD. We should investigate and update the code to be more efficient.
In particular, we need to investigate, dates access, unecessary re-renders, tooltip load time.

A few ideas and examples:
- Prep time in src/components/MapView/index.tsx is super slow. The logic needs to be revisited.
- Some components are running calculations for layers that are not selected. Eg. Download button
- Dates are stored as arrays, we need to see if it is slowing us down and if saving them as a SortedMap in some places could speed things up. Eg. in getRequestDate and to find date intersections.
- Tooltips are super slow to show up. Why? Similarly, admin areas are stored as lists but maybe storing them as maps in some places could be useful. To be investigated.

All, but RBD is a good test case

_No response_

jbperidypathtech:
### Description

This fixes #986.

Main changes made: 
- Split components in more functional parts to avoid all states changes at the root component and target lot of not necessary child components rereders.
- Remove not usefull listener of MapGl components, generating not usefull renders.
- Activate some perfomance tips from MUI to avoid to render not visible content.
- Change tab behavior to not render not visible content. 

In local mode we observe that redux is very slow. I created an issue to upgrade redux libs #1033.
I think we should try to also upgrade mapbox-gl and react-mapbox-gl to latest #1035.

## How to test the feature:
- Use layer selection and compare response time with master. 

## Checklist - did you ...

Test your changes with


frontend/src/components/MapView/LeftPanel/LeftPanelTabs/TabPanel.tsx
```diff
@@ -22,7 +22,7 @@ const TabPanel = memo(({ children, value, index, ...other }: TabPanelProps) => {
       }}
       {...other}
     >
-      {children}
+      {index === value && children}
```
jbperidypathtech, line 25: I add this to nos render contents until tab is selected. Else it is rendered but not visible.

frontend/src/components/MapView/LeftPanel/layersPanel/MenuItem/MenuSwitch/index.tsx
```diff
@@ -129,6 +129,7 @@ const MenuSwitch = memo(({ title, layers, extent }: MenuSwitchProps) => {
       elevation={0}
       classes={{ root: classes.root }}
       onChange={handleAccordionExpand}
+      TransitionProps={{ unmountOnExit: true }}
```
jbperidypathtech, line 132: This options allow to not render content until the accordion is open (not good for seo but for us and performance it is better).

frontend/src/components/MapView/Map/index.tsx
```diff
@@ -171,7 +171,6 @@ const MapComponent = memo(
       (map: Map) => {
         // Track with local state to minimize expensive dispatch call
         const layerIds = new Set<LayerKey>();
-        map.on('sourcedataloading', mapSourceListener(layerIds));
```
jbperidypathtech, line 174: Not usefull because we already have sourcedata listener.
And make a lot of rerenders...
image

laurentS:
This PR improves things for me, at least when running a "production, optimised" build (as opposed to the development setup). For most interactions that don't require network connections, things run smoothly on my machine (a recent intel CPU with 64GB ram, which I'd expect to be more powerful than most users' machines), using a clean firefox nightly with no plugins (so a fairly ideal setup).

The slowest interaction is when opening the "goto" menu, I wait around 900ms (according to firefox profiler) before it appears, and more or less the same to close the dropdown. Even with a profiler, it is difficult to figure out what takes so much time. "Our" code to build the list of items to show runs in 10-15ms, the rest of the time seems to be spent in react internals (by looking at the stack shown in the profiler output). (on countries other than RBD, the menu opens in very reasonable times).
Popups open pretty much instantly for me. At this point, the production build seems fine to me (again, on a fast machine with very fast internet connection).

Memory usage climbs to about 1GB for rbd, which may or may not be a problem, depending on the hardware used. If machines have to start swapping, then this might be part of the slowness.

wadhwamatic:
@jbperidypathtech - thanks for all the work on this. I see a general improvement, but it's not very easy for me to compare. I tried using Chrome's profiler for a build based on the current main branch vs this PR. A flow from load time, to loading a CH layer and then using the go to feature looked be about 8 seconds faster using your PR. 

Here's a screenshot of the profiler summary with the main branch on the left, and this PR on the right following the steps mentioned above:
Screenshot 2024-01-08 at 18 22 00

And here are the profiler json files if helpful:
- Main: https://www.dropbox.com/scl/fi/td06kxlpfc8w4i01t638j/Trace-main.json?rlkey=l8d5clkkz2gakfzkhsxdhyr08&dl=0
- PR 1030: https://www.dropbox.com/scl/fi/6z1oa42eh4fi18483zb3n/Trace-PR-1030.json?rlkey=cgkedf8yt9cflup86xqoovuwe&dl=0 

Functionality on this PR worked as expected for RBD. If there's anything specific I should test, or other ways to compare improvements, let me know. Thanks!

ericboucher (approved):

PR #1036, opened 2023-12-19
Select correct boundary area to request data for

laurentS:
### Description

This is a hotfix for a small bug in popup chart whereby the data sent to the chart was never the correct one.

To reproduce the bug:
- country=cambodia (but any other will show the bug)
- activate a data layer, for instance "rainfall amount"
- click any are of the map to bring up the popup, and load a chart
- save CSV data for easier comparison (or just a good look at the shape of the chart will do)
- click some other part of the map. the chart reloads the same data

(this can be confirmed by looking at the network requests which all relate to the same boundary code)

This PR fixes this, so that data is fetched for the area that was indeed clicked.

## How to test the feature:

Same as above, but the network requests are now different for each click, and admin codes can be confirmed with the json data.

## Checklist - did you ...

Test your changes with

### DIFF
--- a/frontend/src/components/MapView/Layers/BoundaryLayer/index.tsx
+++ b/frontend/src/components/MapView/Layers/BoundaryLayer/index.tsx
@@ -65,6 +65,8 @@ function BoundaryLayer({ layer, before }: ComponentProps) {
 
   const onClickShowPopup = (evt: any) => {
     const coordinates = evt.lngLat;
+    const locationSelectorKey = layer.adminCode;
+    const locationAdminCode = evt.features[0].properties[layer.adminCode];
     const locationName = getFullLocationName(
       layer.adminLevelNames,
       evt.features[0],
@@ -75,7 +77,15 @@ function BoundaryLayer({ layer, before }: ComponentProps) {
       evt.features[0],
     );
 
-    dispatch(showPopup({ coordinates, locationName, locationLocalName }));
+    dispatch(
+      showPopup({
+        coordinates,
+        locationSelectorKey,
+        locationAdminCode,
+        locationName,
+        locationLocalName,
+      }),
+    );
   };
 
   const onClickFunc = (evt: any) => {
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -53,6 +53,7 @@
     "@turf/bbox": "^6.5.0",
     "@turf/bbox-polygon": "^6.5.0",
     "@turf/boolean-point-in-polygon": "^6.5.0",
+    "@turf/centroid": "^6.5.0",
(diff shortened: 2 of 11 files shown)


ericboucher (approved):

laurentS:
Actually, I found a bug in my fix, please don't merge yet. Under analysis tab, run some analysis, then "view table", select a location, then try to open a chart, the app crashes...

wadhwamatic:
Thanks @laurentS. Please ping me when ready to review. This will need some quick turnaround and redeployments when ready

ericboucher:
@laurentS I added a proposed fix - clicking on the map at the given coordinate. The reactivity (and refresh behavior) needs to be improved a bit but it's probably ok for now. Let me know what you think of the approach. I believe that this is a cleaner approach and having only one showPopup trigger will help ensure replicability.

frontend/src/components/MapView/Layers/BoundaryLayer/index.tsx
```diff
@@ -65,6 +65,8 @@ function BoundaryLayer({ layer, before }: ComponentProps) {
 
   const onClickShowPopup = (evt: any) => {
     const coordinates = evt.lngLat;
+    const locationSelectorKey = layer.adminCode;
+    const locationAdminCode = evt.features[0].properties[layer.adminCode];
```
ericboucher, line 69: how about returning the full properties object instead in the popup state?

laurentS:
@ericboucher I like the approach, but it's not really working when I try it locally. At least in some cases, the coordinates are outside the boundary we actually want. I don't know how they're calculated, but on 2-3 randomly picked examples in Mozambique, the point in question was either just outside the boundary.
For instance, click on "Gaza, Chibuto", for me it opens "Inhambane, Panda" instead (which is the area just to the east of it).
Sadly, I don't think we can use these coordinates here :(

ericboucher:
@laurentS  I pushed a commit to use the centroid of the geometry features. Since we were taking the corner, we were always clicking on conflicting boundaries, using the centroid seems to help. There is a small "refresh" problem with the position but I don't think it's a big deal at the moment.

laurentS:
Thanks for the fix, I was just looking into the same solution. I think that long term, we should pass unique IDs around, instead of names, local names and coordinates. They would be much cheaper, and more useful, but as you said, this works fine for now.
I just see a little problem with refreshing the location of the popup if I click one row in the table, then another one, the popup doesn't move to the correct location until I act on the map (zoom, pan...). But not a huge problem as you mentioned.

laurentS:
@wadhwamatic this should be good to go now. Both Eric and I have tested the last fix, which ensures the popup actually moves as expected.

PR #1043, opened 2023-12-29
Virtualize goto dropdown to improve memory usage; COUNTRY=rbd

ericboucher:
Tentatively virtualize goto dropdown to improve memory usage. @laurentS please let me know if you seen an improvement on the deployed version. (i.e. if it's worth fixing the styling details)

This PR also resolves a small UI bug where the goto loading would shift the dropdown size


laurentS (approved):
This does speed things up enormously for the goto menu, to the point that there is no more jank in the profiler output. The menu now displays with only 10 items (which comes from the 350/35 ratio in the code changes, I would probably change it 700/35 to show 20 items instead).

I wonder if it would help on the date selector at the bottom of the screen since it includes tooltips for all dates since 2005 or something. But since not all of them are shown, this probably won't have as much impact.

frontend/src/components/MapView/Layers/BoundaryDropdown.tsx
```diff
@@ -308,23 +302,36 @@ export function SimpleBoundaryDropdown({
...
+      <MenuItem
+        classes={{ root: clsName[(area.level - rootLevel) as number] }}
+        key={area.adminCode}
+        value={area.adminCode}
+        style={style}
+      >
+        {area.label}
+      </MenuItem>
+    );
+  };
```
jbperidypathtech, line 323: Not sure we need to create sub-component for that. I think it could be easier to write it directly line 388 (when it is called)

jbperidypathtech (approved):

ericboucher:
Unfortunately, it seems that mui `Select` is not fully compatible with virtalization. I had to move the click handling from the `Select` `onChange` to the menu item `onClick`. However, the current change breaks the multi-select behavior for alerts...

ericboucher:
I believe I managed to fix the issue on the alert dropdown, please have a look and let me know if it works ok @laurentS @wadhwamatic

PR #1056, opened 2024-01-10
Use download icons for EWS charts; COUNTRY=cambodia

### LINKED ISSUE #1055: [Feature Request]: Make chart tooltip UI consistent with other charts

The chart UI when viewing a climate hazard layer is slightly different from Cambodia's EWS-1294 chart. It would be good to revisit EWS-1294 charts to have them consistent with the climate hazard layers.

Screenshots below are from Cambodia. First is the chart tooltip for the 10-day rainfall estimate which is configured using 'chart_data' in layers.json. The second is from the EWS-1294 layer which I believe is a unique case of point data with it's own chart feature.

Screenshot 2024-01-09 at 22 53 53

Screenshot 2024-01-09 at 22 54 07

_No response_

ericboucher:
### Description

This fixes #1055

Use download icons for EWS charts.
This makes the experience across charts more consistent.
There is more data in the CSV now (it includes the alert thresholds) but the date field only show the time. In addition, the CSV filename is a bit different and only includes the title. I don't think it's an issue but let us know if you would like any of this to change @wadhwamatic 

Screenshot 2024-01-10 at 6 53 57 PM

## How to test the feature:

- [ ] Activate EWS layer
- [ ] Use the download icons on the top right and check the output

## Checklist - did you ...

Test your changes with

### DIFF
--- a/frontend/src/components/MapView/MapTooltip/PointDataChart/PopupPointDataChart.tsx
+++ b/frontend/src/components/MapView/MapTooltip/PointDataChart/PopupPointDataChart.tsx
@@ -1,16 +1,14 @@
 import Chart from 'components/Common/Chart';
-import { buildCsvFileName, isAdminBoundary } from 'components/MapView/utils';
+import { isAdminBoundary } from 'components/MapView/utils';
 import { ChartConfig } from 'config/types';
 import {
   CHART_DATA_PREFIXES,
   datasetSelector,
 } from 'context/datasetStateSlice';
 import { t } from 'i18next';
-import React, { memo, useRef } from 'react';
+import React, { memo } from 'react';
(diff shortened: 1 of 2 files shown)


wadhwamatic:
Thanks @ericboucher. I created an issue for the date field recently: https://github.com/WFP-VAM/prism-app/issues/1049. 

Can we also add the close button to EWS charts? That will keep it fully consistent with other tooltip charts.

wadhwamatic:
> In addition, the CSV filename is a bit different and only includes the title. I don't think it's an issue but let us know if you would like any of this to change @wadhwamatic

The filename is ok as is

wadhwamatic (approved):
Looks good now, thanks!

Issue #467, opened 2022-04-28
Performance - improve handling of large boundary files (eg. RBD)


ericboucher:
We should augment the caching time allowed for boundary files to speed up app initialization

wadhwamatic:
@ericboucher - I'm not sure if what you have in mind will help with an issue we have with large boundary files with complex geometries. For example, there are three admin boundary files used in the upcoming RBD deployment ranging from 14 - 40 mb in size. Loading time is slow as you'd expect (see https://demo-prism-rbd-v4.surge.sh). 

What are your thoughts?

ericboucher:
In some cases, we also seem to be loading admin boundaries that do not get used, for example, http://prism-510.surge.sh/, we load `admin_boundaries.json` but I am not sure it actually gets used

PysarenkoDS-BWT:
I'm try to save save files locally, but not have some results
Try save in LocalStorage, SessionStorage, Cookies but for all this files to large
Also I'm try to use [Cache](https://developer.mozilla.org/ru/docs/Web/API/Cache#browser_compatibility) this is experimental feature but with this large response it not work too

PysarenkoDS-BWT:
Investigate how we can use [topojson package](https://github.com/topojson/topojson) converting every GeoJson file in TopoJson file doesn't make sense because single TopoJson file takes more space then GeoJson (for example try convert `ecu_admbnda_adm2` to TopoJson and file size increase from 37.2 to 41.6 MB)

But we can covert some GeoJson file in one TopoJson, I'm try this and convert `ecu_admbnda_adm1` and `ecu_admbnda_adm2` their size is 17.9 + 37.2 = 55.1 MB and when I convert it their size in TopoJson is 41.7

So that we can convert and store 1 TopoJson file for each country and on the front side convert them back to GeoJson files to reduce the size of the downloaded data from server.

wadhwamatic:
@ericboucher - I think we should look into using vector tiles for boundaries and potentially generate them using tippecanoe 
https://github.com/mapbox/vector-tile-spec/
https://github.com/felt/tippecanoe

ericboucher:
Also relevant and "new" - https://www.openstreetmap.org/user/daniel-j-h/diary/402706

laurentS:
[...]

PR #1060, opened 2024-01-11
Speed up various date processing functions; COUNTRY=rbd

laurentS:
### Description

This fixes (partially) #467 .

I used firefox profiler to identify CPU heavy functions in the code when loading the app for RBD, and tried to speed up the most time consuming ones. Speedups range from 25% up to 10-20x.

## How to test the feature:

- [ ] Build the app with `REACT_APP_COUNTRY=rbd yarn build --profile` (to prevent webpack from  mangling function names, so that the profiler output is usable)
- [ ] Serve it locally with `npx serve -s build`
- [ ] Load the page with the profiler active (for firefox: https://profiler.firefox.com/docs/)
- [ ] explore the flame graph or stack chart tabs for major time consumers (wider blocks) to see time differences before/after

I suggest reviewing the PR commit by commit, as each of them does one thing, it will make more sense than trying to look at everything together.

## Checklist - did you ...

Test your changes with

### DIFF
--- a/common/src/utils/time.ts
+++ b/common/src/utils/time.ts
@@ -1,5 +1,5 @@
 export function setNoon(date: string): string {
-  return date.split("T")[0] + "T12:00:00Z";
+  return `${date.split("T")[0]}T12:00:00Z`;
 }
 
 export const DEFAULT_DATE_FORMAT = "YYYY-MM-DD";
--- a/common/src/utils/time.ts
+++ b/common/src/utils/time.ts
@@ -1,7 +1,5 @@
-import * as moment from "moment";
-
 export function setNoon(date: string): string {
-  return moment.utc(date.split("T")[0]).set({ hour: 12, minute: 0 }).format();
+  return date.split("T")[0] + "T12:00:00Z";
 }
 
 export const DEFAULT_DATE_FORMAT = "YYYY-MM-DD";
--- a/frontend/src/context/store.ts
+++ b/frontend/src/context/store.ts
@@ -38,7 +38,15 @@ export const store = configureStore({
     // serialize the state
     serializableCheck: false,
     immutableCheck: {
-      ignoredPaths: ['mapState.layersData', 'analysisResultState.result'],
+      // do not check the following state branches for accidental
+      // mutations. This saves a lot of time in dev mode (but has no
+      // impact on production builds).
+      ignoredPaths: [
+        'mapState.boundaryRelations',
+        'mapState.layersData',
+        'analysisResultState.result',
+        'serverState.availableDates',
+      ],
     },
   }).concat(errorToNotificationMiddleware),
 });
--- a/frontend/src/utils/date-utils.ts
+++ b/frontend/src/utils/date-utils.ts
@@ -10,7 +10,13 @@ export const datesAreEqualWithoutTime = (
   date1: number | Date,
   date2: number | Date,
 ): boolean => {
-  return new Date(date1).toDateString() === new Date(date2).toDateString();
+  const d1 = new Date(date1);
+  const d2 = new Date(date2);
+  return (
+    d1.getDate() === d2.getDate() &&
+    d1.getMonth() === d2.getMonth() &&
+    d1.getFullYear() === d2.getFullYear()
+  );
 };
 
 export const generateDatesRange = (
--- a/frontend/src/utils/date-utils.ts
+++ b/frontend/src/utils/date-utils.ts
@@ -10,13 +10,9 @@ export const datesAreEqualWithoutTime = (
   date1: number | Date,
   date2: number | Date,
 ): boolean => {
-  const d1 = new Date(date1);
-  const d2 = new Date(date2);
-  return (
-    d1.getDate() === d2.getDate() &&
-    d1.getMonth() === d2.getMonth() &&
-    d1.getFullYear() === d2.getFullYear()
-  );
+  const d1 = new Date(date1).setHours(0, 0, 0, 0);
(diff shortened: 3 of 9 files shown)


frontend/src/utils/date-utils.ts
```diff
@@ -10,7 +10,13 @@ export const datesAreEqualWithoutTime = (
   date1: number | Date,
   date2: number | Date,
 ): boolean => {
-  return new Date(date1).toDateString() === new Date(date2).toDateString();
```
laurentS, line 13: This is not visible at startup, but when clicking around the app a few times with a layer open (change date with the bottom selector, open a popup, etc...), this function consumes a lot of CPU time (~1s after 4-5 clicks). The new code is about 27% faster according to benchmark.

common/src/utils/time.ts
```diff
@@ -1,7 +1,5 @@
-import * as moment from "moment";
-
 export function setNoon(date: string): string {
-  return moment.utc(date.split("T")[0]).set({ hour: 12, minute: 0 }).format();
+  return date.split("T")[0] + "T12:00:00Z";
```
laurentS, line 2: This runs 10-20x faster in firefox. I'm seeing huge savings at page load time (for `getAllLayerDays`, the overall time dropped from 1400ms to about 80ms). It also seems to save about 100MB of RAM for rbd.

frontend/src/utils/date-utils.ts
ericboucher: Maybe worth extracting into a compareDate function? It might already exist in some places actually

laurentS: Changed it again, now it's about twice as fast. I haven't found any `compareDates` function, but it's quite likely that there is one somewhere indeed :)

laurentS:
Each commit saves a bit of CPU. Overall, CPU time drops by ~1.5-2s on my machine, or ~20-30% of CPU time during initial page load. Some of the speed ups also improve things during later usage.

frontend/src/context/store.ts
```diff
@@ -38,7 +38,15 @@ export const store = configureStore({
...
     immutableCheck: {
-      ignoredPaths: ['mapState.layersData', 'analysisResultState.result'],
+      // do not check the following state branches for accidental
+      // mutations. This saves a lot of time in dev mode (but has no
+      // impact on production builds).
+      ignoredPaths: [
+        'mapState.boundaryRelations',
+        'mapState.layersData',
+        'analysisResultState.result',
+        'serverState.availableDates',
```
laurentS, line 48: Adding this line here saves enormous amounts of time in dev mode, both during initial loading and pretty much any interaction afterwards (for instance, a click on the bottom date selector takes about 1.5s vs 6-7s before).
`redux-toolkit` and its immutable middleware (in dev mode) runs all `redux` actions **twice**, and each time does a deep comparison of the entire state tree to check for accidental mutations. The `ignoredPaths` tells it to skip these parts of the tree. `serverStates.availableDates` holds around 80 layers of 250-3300 dates each, which won't actually change once they've been loaded, so this saves a ton of useless work.

[This](https://github.com/reduxjs/redux-toolkit/blob/b283b1c9d7fc9b7b9bd5729998ab84490db1dbe5/packages/toolkit/src/immutableStateInvariantMiddleware.ts#L166) is the checks it runs (for educational purposes) :wink: 

The more radical option is to set `immutableCheck: false` which entirely disables these checks. At that point, dev mode is virtually as fast as the production build, but the drawback is that we lose some useful checks.

A middle ground might be to disable the checks in dev mode, but turn them on in CI, as any mutation will throw an error. This obviously assumes that our tests cover all such cases.

wadhwamatic:
@laurentS - Things definitely appear to load faster and the UI generally feels snappier. But when I tried to load a CH data layer and then select a rainfall layer without an intersection, the app froze.

To replicate:
1. REACT_APP_COUNTRY=rbd yarn build --profile
2. npx serve -s build
3. Activate layer: Cadre Harmonisé > Phase Classification > Phase classification. The latest data will load (2023-09-30)
4. Activate layer: Rainfall > Rainfall Amount > Rainfall Aggregate 10-day. The date will remain the same
5. Use the handle in the timeline and drag to a future date without an intersection (2023-10-11 as an example)
6. A message will appear saying the closest date was loaded instead of saying the intersection doesn't exist, and the app freezes up

Screenshot for PR 1060:
Screenshot 2024-01-15 at 22 34 05

Note the same steps works on the main branch with a message correctly saying the intersection doesn't exist. Screenshot from main:
Screenshot 2024-01-15 at 22 47 33

ericboucher:
@laurentS I think I fixed the issue, we ended up with overlapping default date and validity dates with slightly different displayDate. So the `sortedUniqBy` was not deduping properly, and default dates ended up appearing like they had multiple layers available. When in reality it was just the one. It might be worth moving the setHour again but I'll let you be the judge of that

wadhwamatic:
Thanks @ericboucher. The date issue is resolved

laurentS:
@wadhwamatic I think this PR is probably pushed as far as we can reasonably go. When profiling, I'm now seeing most of the CPU time spent in dependencies, not our own code. There's still room to improve a bit, but we're now in the less rewarding part of the Pareto curve :)

I'd suggest we merge as is, unless you hit other issues when testing.

wadhwamatic:
I ran a few tests in RBD, Cambodia and Mozambique and wasn't able to break anything =)

Great work on this Laurent. The performance improvements are very pleasing!

ericboucher (approved):

PR #1062, opened 2024-01-17
Add script to check boundary key validity

ericboucher:
### Description

This script checks that the keys listed in layers.json for boundaries exist in the corresponding file.
The script now runs as part of the CI

This unearthed a few missing keys:

```
Key ST_MMR4 does not exist in file ../public/data/myanmar/mmr_polbnda_adm1_250k_mimu.json
Key ST_MMR4 does not exist in file ../public/data/myanmar/mmr_polbnda_adm3_250k_mimu.json
Key DT_MMR4 does not exist in file ../public/data/myanmar/mmr_polbnda_adm3_250k_mimu.json
Key TS_MMR4 does not exist in file ../public/data/myanmar/mmr_polbnda_adm3_250k_mimu.json
Key ProvinceR  does not exist in file ../public/data/tajikistan/tjk_admin2_wgs84_clean.json
```

@wadhwamatic let me know how you want to handle those to turn the CI check green.

## How to test the feature:

- [ ] run `yarn check-boundary-keys`


frontend/scripts/check-boundary-keys-validity.js
```diff
@@ -0,0 +1,67 @@
...
+
+  // Check if the keys exist in the JSON file
+  const keysToCheck = [
+    ...layer.admin_level_codes,
+    ...layer.admin_level_names,
+    ...layer.admin_level_local_names,
+  ];
+
+  keysToCheck.forEach(key => {
+    if (!fileContent.includes(key)) {
```
laurentS, line 21: If we're very rigorous, this only checks that the key string appears in the content of the file, not that it's actually a JSON key, nor that it's in the right place. I guess a more thorough check would be provided by a json schema or equivalent, but that's probably a bit more work to implement?

laurentS (approved):

frontend/scripts/check-boundary-keys-validity.js
ericboucher: I am testing `"${key}":` to make this a bit more robust

wadhwamatic:
@ericboucher - I thought the issue with the Tajikistan boundaries might have been unicode text for Russian names so went ahead and fixed that and removed unused fields as well. The problem was actually in layers.json where there was an extra space in 'ProvinceR'. 

The test runs successfully now

wadhwamatic (approved):

PR #1070, opened 2024-01-22
Fix EWS-1294 date missing in the CSV download; COUNTRY=cambodia

### LINKED ISSUE #1049: [Bug]: Cambodia - EWS-1294 date field is missing info in the CSV download
### What happened?

When I download a CSV file for EWS-1294 data, the date column shows time of day, but not the full date. We should use the full date + time in this column, or have a separate column for the date.

To replicate, use Cambodia and select Flood > Early warning > EWS-1294. Then click on a data point to see the chart, then click the download button.

Cambodia on main branch.

Screenshot 2024-01-08 at 16 38 41

echaidemenos:
### Description

This fixes #1049 .

Fixes a bug where only the time would appear in the CSV for Cambodia - EWS-1294 layer

## How to test the feature:

- select Cambodia -> Flood -> Early Warning
- select array to display chart on the map
- use download CSV button and verify the date show both date and time

## Checklist - did you ...

Test your changes with


frontend/src/components/Common/Chart/index.tsx
```diff
@@ -27,15 +27,19 @@ interface ChartData {
   }[];
 }
 
-function downloadCsv(chartData: ChartData, filename: string) {
+function downloadCsv(
```
ericboucher, line 30: This is a bit counter-intuitive to me. It doesn't feel like this should be the responsibility of "downloadCSV". Whoever is calling downloadCSV should prepare the data.

Also, it looks like we have two functions `downloadCsv` that do basically the same thing, so we need to homogenize this.

I would actually suggest inverting the behavior, keep the full date in the charts data, but have an option to display time only when we visualize the chart. What do you think?

frontend/src/components/Common/Chart/index.tsx
```diff
@@ -153,16 +157,27 @@ const Chart = memo(
...
+          return moment(row[config.category])
+            .locale(t('date_locale') as LocaleSpecifier)
+            .format('YYYY-MM-DD HH:mm');
+        });
+
+      return data.EWSConfig
+        ? [fullLabels.map(x => x.split(' ')[1]), fullLabels]
+        : [
+            fullLabels.map(x => x.split(' ')[0]),
+            fullLabels.map(x => x.split(' ')[0]),
```
laurentS, line 179: You're returning the same values twice, is this normal? Should the second part be `fullLabels` instead for consistency with above?

frontend/src/components/Common/Chart/index.tsx
```diff
@@ -153,16 +157,27 @@ const Chart = memo(
...
-        return moment(row[config.category])
-          .locale(t('date_locale') as LocaleSpecifier)
-          .format(dateFormat);
-      });
+      const fullLabels = tableRows
+        .slice(chartRange[0], chartRange[1])
+        .map(row => {
+          return moment(row[config.category])
+            .locale(t('date_locale') as LocaleSpecifier)
+            .format('YYYY-MM-DD HH:mm');
```
laurentS, line 172: Is there a way to get rid of `moment` here?

frontend/src/components/Common/Chart/index.tsx
```diff
@@ -366,9 +354,15 @@ const Chart = memo(
...
                 <IconButton
-                  onClick={() =>
-                    downloadCsv(chartData, title.split(' ').join('_'))
-                  }
+                  onClick={() => {
+                    const chartData = {
+                      labels: labels.map(x =>
+                        shouldDisplayTime ? x : String(x).split(' ')[0],
+                      ),
+                      datasets: datasetsTrimmed,
```
ericboucher, line 362: Not sure I understand why this is needed?

ericboucher (approved):

PR #1071, opened 2024-01-22
Add CDI; COUNTRY=jordan

ericboucher:
### Description

This PR adds a CDI layer to Jordan.

It adds a new config key "dateLayer"  for the CDI layer which allows us to reference the layer that should be used to determine available dates.

Note that I also made a significant changed around how we store dates for WMS and WCS layers so that we store them with our layerIds instead of the serverlLayerName. This makes dates a bit easier to grab, and also saves up a bunch of memory by not storing dates we do not need. cc @laurentS 

TODO:
- [x] get dates
- [x] add legend
- [x] add date trigger
- [x] pre-build country polygon
- [x] test hexagons
- [ ] update legend
- [x] connect to backend when available

## How to test the feature:

- [ ]  https://prism-1071.surge.sh/?hazardLayerIds=cdi
- [ ]  play with the opacity, dates and so on (note that this is not linked to the backend yet)

## Checklist - did you ...

Test your changes with

- [ ] `REACT_APP_COUNTRY=jordan yarn start`

Screenshot 2024-01-22 at 10 33 12 PM


frontend/src/utils/layers-utils.tsx
```diff
@@ -94,6 +95,14 @@ const useLayers = () => {
     return bbox(boundaryLayerData.data) as Extent; // we get extents of admin boundaries to give to the api.
   }, [boundaryLayerData]);
 
+  // TODO - investigate if this is not too time consuming for larger countries.
```
ericboucher, line 98: @laurentS we need to investigate if this is the right place to do this calculation and what the performance impact is on a larger deployment like RBD.

frontend/src/components/MapView/Layers/CompositeLayer/index.tsx
```diff
@@ -26,7 +59,34 @@ const CompositeLayer = ({ layer }: Props) => {
     }
   }, [dispatch, layer, data, date]);
 
-  return <>test</>;
+  const { adminBoundaryLimitPolygon } = useLayers();
+
+  if (data && adminBoundaryLimitPolygon) {
+    const filteredData = {
```
ericboucher, line 65: Another option is to use mapbox filtering directling using the `within` command. When I used it locally it seemed slower but it could be worth checkin on deployed versions. Here is the code for reference:

```
  const { adminBoundaryLimitPolygon } = useLayers();

  const paintProps: (
    opacity: number | undefined,
  ) => CircleLayerSpecification['paint'] = (opacity?: number) => ({
    'circle-opacity': [
      'case',
      ['within', adminBoundaryLimitPolygon],
      opacity || 0.5,
      0,
    ],
    'circle-color': [
      'interpolate',
      ['linear'],
      ['get', 'value'],
      -2,
      '#FFFF33',
      2,
      '#FF8433',
    ],
    'circle-radius': [
      'interpolate',
      ['exponential', 2],
      ['zoom'],
      0,
      1 * scale,
      15,
      16 * scale,
    ],
  });
```

frontend/src/utils/layers-utils.tsx
ericboucher: Well it crashed for Cambodia so that's an answer 😅

frontend/src/components/MapView/Layers/CompositeLayer/index.tsx
ericboucher: Note that it also seems to accept only one feature / polygon at a time, so we still need to do the merge on our own...

frontend/src/utils/layers-utils.tsx
ericboucher: Maybe we should simply run a script to save this boundary once per country

laurentS:
Just a visual comment about the rendering of the layer. I suppose it's intended that way, but I find the circle rendering a bit hard to read in some cases, depending on how it's zoomed in.

When zoomed out, I find the colour scale to be readable:
image

But when zooming in, I find it hard to tell the values apart.
image

I wonder if it would help to increase the radius as you zoom in?

frontend/src/components/MapView/Layers/layer-utils.tsx
```diff
@@ -122,3 +124,28 @@ export const addPopupParams = (
     ),
   );
 };
+
+// Define the type for boundary data
+type BoundaryData = GeoJSON.FeatureCollection;
+
+// Define the function
```
laurentS, line 131: You have written more inspiring comments before :sweat_smile:

frontend/src/utils/layers-utils.tsx
laurentS: To be sure I'm not misreading the code, this calculates a single boundary polygon that includes the entire country's (or multicountries' for RBD) area. I can't really see the value in recalculating this on every app run. By definition, it won't change more often than the underlying boundary files. I'd vote for precomputing this, either at deployment time (or when the dev version starts), or even add the output in git along with the boundary definitions (produced using the code you wrote here). As a side note, having that one big polygon available could also speed up `boundaryLayerDataAreOutsideOfBoundaryBBox` which is called on every click on the map (to check if we clicked inside the country).

frontend/src/components/MapView/Layers/layer-utils.tsx
ericboucher: 🤖  haha

wadhwamatic:
> Just a visual comment about the rendering of the layer. I suppose it's intended that way, but I find the circle rendering a bit hard to read in some cases, depending on how it's zoomed in.
> 
> When zoomed out, I find the colour scale to be readable: image
> 
> But when zooming in, I find it hard to tell the values apart. image
> 
> I wonder if it would help to increase the radius as you zoom in?

I was also wondering if points are the correct way to visualize this. I believe CDI is calculated per each 5km square in a grid. Would a square or a hexagon be a better way to represent the response? @lowriech - what drove the decision to use points here?

ericboucher:
I played around a bit with the point size based and zoom and it should look a bit better now.
@wadhwamatic I think Points are a good first choice to test things out. Then we will need to think about the best format a bit more, taking into account size constraints. Unfortunately, hexagon files are a lot larger for example. See the size discussion in https://github.com/WFP-VAM/hip-service/issues/6 for reference

laurentS:
This might be a stupid idea, but it never hurts to suggest those ;)
My understanding is that you're trying to visualise 2 dimensions at the same time, the CDI score (currently rendered as the colour of this point) and some other metric (rendered as the color of the rest of the square).
Could it work to use some other dimension than colour? Say a hash pattern with black lines of varying thickness depending on the CDI score? Though it might be difficult to read when zooming far out, because the data points get closer on the screen.
I'm saying this because I found it difficult to read the colour scale. Maybe a wider range of colours is enough to fix that though (I am not colour-blind, by the way, as far as I know).

ericboucher:
I just pushed a change to transform the points into h3 hexagons. It works well for Jordan. I am not too sure how it would behave for large countries though.

Screenshot 2024-01-26 at 2 47 52 PM

wadhwamatic:
> My understanding is that you're trying to visualise 2 dimensions at the same time, the CDI score (currently rendered as the colour of this point) and some other metric (rendered as the color of the rest of the square).

@laurentS - for CDI, it's just a single variable that we're visualizing. We do in the near future want to visualize two variables together. There's an old issue about using bivariate choropleth maps to try and address this: https://github.com/WFP-VAM/prism-app/issues/229

> I'm saying this because I found it difficult to read the colour scale. Maybe a wider range of colours is enough to fix that though (I am not colour-blind, by the way, as far as I know).

Yeah, I think we can improve the color choice here. Something like this perhaps https://colorbrewer2.org/#type=sequential&scheme=YlOrBr&n=5

ericboucher:
@wadhwamatic I pushed a change to add more color steps and used the one you suggested.
Feel free to play with it in `frontend/src/components/MapView/Layers/CompositeLayer/index.tsx`. We can change the legend once we settle on a color scheme.

lowriech:
A few thoughts on this while I work on understanding the PRISM data model:
- How are we controlling the available date ranges?  If I understand correctly, it seems like we're mapping some concept of available dates (`dateLayer`) onto the CDI to power the light blue icons on the time slider, which then need to map into requests to the hip-service output.  So we basically determine the available dates at build time, which gets around having full user configuration.

Also noting that I can view in the surge site, but not when running locally.  I get a `geojson` response from `hip-service`, but it doesn't get added to the map.

wadhwamatic (approved):

PR #1081, opened 2024-01-26
Register onclick with map.on

echaidemenos:
### Description
changes the way we register layer to use `map.on`. This way `map.fire` triggers onclick events.
note that the event produced by `map.fire`, returns coordinates in an array instead of an object

This fixes #issueNumber.

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with

### DIFF
--- a/frontend/src/components/MapView/Layers/AdminLevelDataLayer/index.tsx
+++ b/frontend/src/components/MapView/Layers/AdminLevelDataLayer/index.tsx
@@ -20,6 +20,7 @@ import {
   firstBoundaryOnView,
   getLayerMapId,
   isLayerOnView,
+  useMapCallback,
 } from 'utils/map-utils';
 import { fillPaintData } from 'components/MapView/Layers/styles';
 import { availableDatesSelector } from 'context/serverStateSlice';
@@ -30,7 +31,6 @@ import {
 } from 'components/MapView/Layers/layer-utils';
 import { convertSvgToPngBase64Image, getSVGShape } from 'utils/image-utils';
 import { FillLayerSpecification } from 'maplibre-gl';
-import { useSafeTranslation } from 'i18n';
 
 const onClick = ({
   layer,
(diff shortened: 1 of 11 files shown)


frontend/src/components/MapView/Layers/BoundaryLayer/index.tsx
```diff
@@ -76,26 +75,44 @@ export const onClick = ({
...
+  const layerId = getLayerMapId(layer.id, 'fill');
+
+  useEffect(() => {
+    if (!map) {
+      return () => {};
+    }
+
+    map.on('click', layerId, onClick({ dispatch, layer, t }));
+    map.on('mouseenter', layerId, onMouseEnter({ dispatch, layer, t }));
+    map.on('mouseleave', layerId, onMouseLeave({ dispatch, layer, t }));
```
ericboucher, line 109: do we need these?

frontend/src/components/MapView/Layers/AdminLevelDataLayer/index.tsx
```diff
@@ -106,6 +110,21 @@ const AdminLevelDataLayers = ({
...
+
+    map.on('click', getLayerMapId(layer.id), onClick({ dispatch, layer, t }));
+    return () => {
+      map.off(
+        'click',
+        getLayerMapId(layer.id),
+        onClick({ dispatch, layer, t }),
+      );
+    };
+  }, [dispatch, layer, layer.id, map, t]);
```
ericboucher, line 126: Let's test things a bit but if this works well let's turn this into a custom useEffect that takes a custom onClick and data as input. Then we can reuse these functions accross the different layers

You probably just need `map.off('click', layerId);` for the cleanup part. wdyt?

frontend/src/utils/map-utils.ts
```diff
@@ -89,3 +102,27 @@ export const getLayerMapId = (layerId: string, type?: 'fill' | 'line') =>
 // evt emitted by map.fire has array of coordinates, but other events have an object
 export const getEvtCoords = (evt: MapLayerMouseEvent) =>
   Array.isArray(evt.lngLat) ? evt.lngLat : [evt.lngLat.lng, evt.lngLat.lat];
+
+export function useMapCallback<T extends keyof MapLayerEventType, U>(
+  type: T,
+  layerId: string,
+  layer: U,
```
ericboucher, line 109: maybe this could be a more flexible `data` input, wdyt?

PR #1086, opened 2024-01-31
Add raster bounds and country mask

### LINKED ISSUE #170: Mask out raster data outside of national boundaries
When a user loads a raster hazard layer (such as NDVI), it's easy to lose focus on the area of interest (aka the deployment country). This is especially true in landlocked countries, and countries with complex boundaries. See the examples from Tajikistan below.

Screen Shot 2021-04-22 at 12 29 24

Screen Shot 2021-04-22 at 12 29 10

We need to explore options to visually keep focus on the area of interest while also recognizing that many hazard layers should be seen in comparison across national borders. We could consider a mechanism that 'greys out' countries outside of the area of interest, or to fully mask data outside of the area of interest. See a relevant tutorial here: https://www.mapbox.com/blog/masking-in-studio-using-high-resolution-administrative-country-polygons

ericboucher:
### Description

This fixes #170.

- Limit the bounds of the raster layer we are fetching
- Adds an opacity mask when using raster layers, @wadhwamatic let's decide if we would like to make this optional, or if we only want to display the mask in print mode maybe. Thoughts?

## How to test the feature:

- [ ] activate any WMS layer
- [ ] verify that the mask looks good for every country, I had to run a simplification algorithm to avoid the weird edge cases we were seeing for moz.

## Checklist - did you ...

Test your changes with

Screenshot 2024-02-08 at 12 24 00 PM


ericboucher:
Let's only apply the mask in print (optional).
There is also a weird bug in mz at the moment:
Screenshot 2024-02-07 at 7 36 56 PM

wadhwamatic:
@ericboucher - this is looking good. Small thing - the north arrow looks to be behind the mask

ericboucher:
@wadhwamatic it was on top but transparent. I changed the image and put it in the middle of the scale bar

wadhwamatic (approved):

PR #1090, opened 2024-02-02
Add tests around date usage/formating

### LINKED ISSUE #1063: Remove momentjs dependency

`momentjs` is deprecated and slow, we should remove it as a dependency 

_No response_

echaidemenos:
This PR aims to add test around various components and function, which use dates, making it easier to resolve #1063 and to increase app test coverage overall


frontend/src/components/MapView/DateSelector/__snapshots__/index.test.tsx.snap
```diff
@@ -62,15 +62,21 @@ exports[`renders as expected 1`] = `
...
             class="memo-timeline-9 react-draggable"
             id="dateTimelineSelector"
-            style="transform: translate(0px,0px);"
+            style="transform: translate(-690px,0px);"
           >
             <div
               class="MuiGrid-root memo-dateLabelContainer-8 MuiGrid-container"
-            />
+            >
+              <mock-timelineitems
```
ericboucher, line 70: let's mock this again

frontend/package.json
```diff
@@ -21,7 +21,7 @@
     "analyze": "yarn build && npx source-map-explorer 'build/static/js/*.js'",
     "build": "export GIT_HASH=$(git rev-parse --short HEAD) || true && cross-env EXTEND_ESLINT=true react-app-rewired build",
     "profile:serve": "yarn build --profile && npx serve -s build",
-    "test": "cross-env REACT_APP_TESTING=true && react-scripts test",
+    "test": "cross-env REACT_APP_TESTING=true TZ=UTC react-scripts test",
```
ericboucher, line 24: Is the change in global-setup not enough?
Ideally we would want the test to go over mutiple timezones and make sure nothing breaks in any timezone. We need to test for:

- far wes, eg. California Timezone
- europe, eg. Paris timezone
- far east, eg. China timezone
- edge cases, eg. timezone with +Xh30min diff

frontend/src/utils/server-utils.ts
```diff
@@ -207,7 +207,7 @@ export const getStaticRasterDataCoverage = (layer: StaticRasterLayerProps) => {
     return [];
   }
   // raw data comes in as {"dates": ["YYYY-MM-DD"]}
-  return dates.map(v => moment(v, 'YYYY-MM-DD').valueOf());
+  return dates.map(v => moment.utc(v, 'YYYY-MM-DD').valueOf());
```
ericboucher, line 210: why make the change here but not on line `201`?

ericboucher: Is it worth doing what we do for other layers?
`.map(item => moment.utc(item.date).set({ hour: 12, minute: 0 }).valueOf())`

frontend/src/components/MapView/Layers/StaticRasterLayer/index.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    timezoneMock.unregister();
+  });
+
+  test('Should work with UTC', () => {
+    timezoneMock.register('UTC');
+    f();
+  });
+
+  test('Should work with US/Pacific', () => {
+    timezoneMock.register('US/Pacific');
```
laurentS, line 25: Maybe add a test with a timezone on the east side of UTC? like UTC+6 or whatever. Ideally, we'd have a list of all timezones, and run the test in a loop for each one. Maybe something like https://jestjs.io/docs/api#testeachtablename-fn-timeout ?

frontend/src/utils/date-utils.test.ts
```diff
@@ -67,3 +75,105 @@ describe('Binary search in ordered arrays of timestamps', () => {
     expect(binaryFind<number>(arr, 42, x => x)).toEqual(-1);
   });
 });
+
+const defaultFormateTests = [
```
laurentS, line 79: typo?
```suggestion
const defaultFormatTests = [
```

frontend/src/utils/date-utils.ts
```diff
@@ -90,3 +94,28 @@ export function binaryIncludes<T extends any>(
 ): boolean {
   return binaryFind(arry, timestamp, callback) > -1;
 }
+
+export const dateStrToUpperCase = (dateStr: string): string => {
+  return `${dateStr.slice(0, 1).toUpperCase()}${dateStr.slice(1)}`;
+};
+
+export const getDateFormat = (
```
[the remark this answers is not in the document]
ericboucher: Fixed in the next PR

frontend/src/utils/ews-utils.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    timezoneMock.register('UTC');
+    f();
+  });
+
+  test('Should work with US/Pacific', () => {
+    timezoneMock.register('US/Pacific');
+    f();
+  });
+
+  test('Should work with Etc/GMT-1', () => {
```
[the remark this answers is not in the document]
ericboucher: I think that's fine as is

frontend/src/components/MapView/Layers/StaticRasterLayer/index.test.tsx
ericboucher: Let's do india standard time, it's both ahead of UTC and 30min off @echaidemenos

ericboucher (approved):

PR #1102, opened 2024-02-04
Add a filename_override parameter for geotiff downloads

ericboucher:
### Description

This fixes #issueNumber.

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with

### DIFF
--- a/api/app/geotiff_from_stac_api.py
+++ b/api/app/geotiff_from_stac_api.py
@@ -109,14 +109,23 @@ def get_geotiff(
     bbox: [float, float, float, float],
     date: str | None = None,
     band: str | None = None,
+    filename_override: str | None = None,
 ):
     """Generate a geotiff and return presigned download url"""
     s3_filename = generate_geotiff_and_upload_to_s3(collection, bbox, date, band)
 
     s3_client = boto3.client("s3")
+
+    params = {
+        "Bucket": GEOTIFF_BUCKET_NAME, 
+        "Key": s3_filename
+    }
+    if filename_override is not None:
+        params["ResponseContentDisposition"] = f"attachment; filename=\"{filename_override or s3_filename}\""
+
     presigned_download_url = s3_client.generate_presigned_url(
         "get_object",
-        Params={"Bucket": GEOTIFF_BUCKET_NAME, "Key": s3_filename},
+        Params=params,
         ExpiresIn=3600,
     )
     return presigned_download_url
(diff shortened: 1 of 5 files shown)


wadhwamatic:
Thanks @ericboucher. Looks good, but I don't think we've handled the 'band' issue and/or we're not getting the right layer name when we use the drop down menu to choose a layer with a different time aggregation (i.e. 10-day vs 3-month)

For example, if I choose the 3-month aggregate instead of 10-day (https://prism-1102.surge.sh/?hazardLayerIds=precip_blended_3m&date=2024-01-21) and try to download the layer, I expected the filename to be precip_blended_3m_2024-01-21 but instead its mozambique_precip_blended_dekad_2024-01-21 and appears to download the 10-day version. 

Side note - I was just checking on master, and I'm not sure if https://github.com/WFP-VAM/prism-app/pull/1006 is working correctly. Using Mozambique on master, I tried to download 10-day rainfall anomaly and 3-month rainfall anomaly. The files are identical (name and the data too). See this zip file:

[rfq_dekad_cfc92607_2024-01-21.zip](https://github.com/WFP-VAM/prism-app/files/14214333/rfq_dekad_cfc92607_2024-01-21.zip)

ericboucher:
@wadhwamatic I couldn't reproduce the name issue you were mentioning, the filename seems correct for me.
However, I was able to reproduce the second error. This was because the band is relying on us setting the band parameter in the config. I created a utility function to use the `style` and just remove `_blended` from it of no band was explicitely set. Does that seem reasonable?

frontend/src/config/utils.ts
```diff
@@ -338,3 +338,20 @@ export const getCompositeLayers = (layer: LayerType): LayerType[] => {
...
   return [];
 };
+
+/**
+ * A utility function to get the STAC band parameter.
+ * WARNING: This function is fragile as it is using the _blended pattern in styles.
+ * @param additionalQueryParams - additional query parameters
+ * @returns the band parameter
+ */
+export const getStacBand = (
```
ericboucher, line 348: @wadhwamatic let me know if this function makes sense

wadhwamatic:
@ericboucher - the name issue happens consistently when you download from the layer panel, but not the legend.

> I created a utility function to use the style and just remove _blended from it of no band was explicitely set. Does that seem reasonable?

yes - that should work fine.

wadhwamatic:
@ericboucher - looks like that change made the download feature disappear for layers without a grouping feature. For example, I can't download forecasts or dry periods from the layer nav panel.

wadhwamatic (approved):
looks good now, thanks @ericboucher !

PR #1116, opened 2024-02-10
add new round of CH data; COUNTRY=RBD

wadhwamatic:
This PR makes a few configuration changes for RBD:
1.  Adds new CH data rounds
2. Adds 'percent of population' layers 
3. Makes a few changes in how layers are organized, using the layer group feature
4. Removes the ability to overlay the general population with special populations

## Checklist - did you ...

Test your changes with


wadhwamatic:
@ericboucher - I'm having a hard time figuring out two things. Not sure if these are bugs, config issues, or data issues. Thanks for any pointers!

1. The fallback data (admin 2 falling back on admin 1), doesn't show the full table view that admin 2 does. I think this is likely as designed but maybe something we should revisit. Not urgent though
2. Tooltip info is missing altogether for ch_pop_3_5 when the layer falls back to admin 1

Issue 1 - tooltip table view is missing for admin 1 fallback. See full table when admin2 data is available, but just the data field when we fallback to admin 1.
Screenshot 2024-02-10 at 17 07 56
Screenshot 2024-02-10 at 17 08 01

Issue 2 - tooltip info is missing for ch_pop_3_5. I tried many things on the config to try and address this. 
Tooltip fallback shows data field for ch_phase
Screenshot 2024-02-10 at 17 03 50

But for ch_pop_3_5, the tooltip doesn't show any data, even though the map symbology shows a value exists.
Screenshot 2024-02-10 at 17 04 17

wadhwamatic:
@ericboucher - there's an edge case that's causing an issue here that I didn't catch before. There's a round of data where there is 'urban' data at admin1 level, but not at admin2 level (2021-02-01). In that scenario, we don't see the option in the date picker or timeline to show data for Feb 2021 for any of the CH urban layers. 

I'm guessing we assume the admin2 layer is where to get the available dates. We should instead look at both the main layer, and the fallback data layer to populate available dates. Is that what is happening or is there something else going on here?

ericboucher:
@wadhwamatic I might be missing something but it looks like the fallback data is not configured for the urban datasets yet and the layer that's missing is not configured in the layers.json, so their dates never get pulled.

So we need to:
1. create new layers with basic info for `urb_admin1_`
2. add these layers as fallbackLayer for the existing `urb_admin2` layers

wadhwamatic:
@ericboucher - found two more issues. Let me know if we want to resolve them within this PR or separately.
https://github.com/WFP-VAM/prism-app/issues/1134
https://github.com/WFP-VAM/prism-app/issues/1135

ericboucher:
I think I fixed both but let me know if it works for you too. For the date issue we should remember to check that it still works on https://github.com/WFP-VAM/prism-app/pull/1121

wadhwamatic:
This looks good now, thank you @ericboucher. I'll go ahead and merge and will revisit PR 1121 after.

PR #1121, opened 2024-02-13
Remove momentjs dependency; COUNTRY=cambodia

### LINKED ISSUE #1063: Remove momentjs dependency

echaidemenos:
### Description

This fixes #1063 and removes dependency to `momentjs` from the project

## How to test the feature:

Make sure that everything date related works as expected and translates properly. @wadhwamatic let's QA this extensively before merging.

- [ ] date for layers
- [ ] charts
- [ ] CSVs
- [ ] date picker

## Checklist - did you ...

Test your changes with


frontend/src/utils/date-utils.ts
```diff
@@ -100,22 +114,46 @@ export const dateStrToUpperCase = (dateStr: string): string => {
 };
 
 export const getDateFormat = (
```
ericboucher, line 116: might be worth renaming to `getFormattedDate`, wdyt?

frontend/src/utils/ews-utils.ts
```diff
@@ -57,16 +57,19 @@ type EWSTriggerLevels = {
...
-  const tempDate = moment.utc('2021-01-01');
+  const endDate = testEndDate || now.setUTCHours(0, 0, 0, 0);
+
+  let tempDate = new Date('2021-01-01');
 
   while (tempDate.valueOf() <= endDate) {
+    const clone = new Date(tempDate.getTime());
     // eslint-disable-next-line fp/no-mutating-methods
-    datesArray.push(tempDate.clone().set({ hour: 12, minute: 0 }).valueOf());
+    datesArray.push(clone.setHours(12));
```
ericboucher, line 69: we also need to set the minutes to 0 for 30min timezones
it might be worth creating a utility `setToNoon` to reuse across the codebase.

frontend/src/utils/date-utils.ts
```diff
@@ -100,22 +114,46 @@ export const dateStrToUpperCase = (dateStr: string): string => {
...
     case 'snake':
-      return moment(date).format(DEFAULT_DATE_FORMAT_SNAKE_CASE);
+    case DEFAULT_DATE_FORMAT_SNAKE_CASE:
+      return `${year}_${month}_${day}`;
+    case 'YYYY-MM-DD HH:mm':
+      return `${year}-${month}_${day} ${hours}:${minutes}`;
+    case 'DD_MM_YYYY':
+      return `${day}_${month}_${year}`;
+    case 'YYYY-MM-DDTHH:mm:ss':
+      return `${year}-${month}-${day}T${hours}:${minutes}:${seconds}`;
```
ericboucher, line 151: Did you look into `date-fns`? Seems like we're reinventing the wheel a bit here no?

echaidemenos: This is as vanilla javascript and as simple as it gets. If you want we can use `date-fns`

frontend/src/components/MapView/DateSelector/utils.ts
```diff
@@ -22,19 +22,21 @@ export function findClosestDate(
   date: number,
   availableDates: ReturnType<Date['getTime']>[],
 ) {
-  const dateToCheck = moment(date);
-
   // TODO - better handle empty arrays.
   if (availableDates.length === 0) {
-    return dateToCheck;
+    return date;
```
ericboucher, line 27: The function docstring is not accurate anymore. Should this function return a date object maybe? Not sure how it's used atm

frontend/src/components/MapView/LeftPanel/ChartsPanel/ChartSection/index.tsx
```diff
@@ -49,8 +50,8 @@ function generateDateStrings(startDate: Date, endDate: Date) {
   const result = [];
   const interval = [1, 11, 21];
   const currentDate = new Date(startDate);
-  currentDate.setHours(10);
-  endDate.setHours(12);
+  currentDate.setHours(10, 0, 0, 0);
```
ericboucher, line 53: not from this PR but curious why we set it to 10 here?

frontend/src/components/MapView/DateSelector/utils.ts
echaidemenos: updated the docstring to number, timestamp in milliseconds

frontend/src/components/MapView/LeftPanel/ChartsPanel/ChartSection/index.tsx
echaidemenos: It is like that so in the `while` bellow the current is before endDate

ericboucher: the while loop uses a `<=` so it would fo through anyway for the first day, no?

echaidemenos: I am adding a test for this function (on the tests PR #1090 ). I believe it has to do with daylight savings. I will confirm and add a comment also

echaidemenos: seems to work ok with 12, so i set it to 12

frontend/src/utils/date-utils.ts
laurentS: I agree, I made the same comment on the other PR about testing, before seeing your comment here.

ericboucher: I pushed the change yesterday :)

frontend/src/utils/date-utils.ts
```diff
@@ -99,23 +113,47 @@ export const dateStrToUpperCase = (dateStr: string): string => {
...
     return undefined;
   }
 
+  const jsDate = new Date(date);
+  const year = jsDate.getUTCFullYear();
+  const month = String(jsDate.getUTCMonth() + 1).padStart(2, '0');
+  const day = String(jsDate.getUTCDate()).padStart(2, '0');
+  const hours = String(jsDate.getUTCHours()).padStart(2, '0');
+  const minutes = String(jsDate.getUTCMinutes()).padStart(2, '0');
+  const seconds = String(jsDate.getUTCSeconds()).padStart(2, '0');
```
laurentS, line 137: These 3 lines are only needed in some cases, maybe put them behind a test to only compute them if needed, as they're probably quite expensive, relatively speaking.
Overall, this function is going to be called quite a lot, so I think it's worth speeding it up where possible.

frontend/src/utils/ews-utils.ts
```diff
@@ -57,16 +58,19 @@ type EWSTriggerLevels = {
...
+
+  while (tempDate.getTime() <= endDate) {
+    const clone = new Date(tempDate.getTime());
     // eslint-disable-next-line fp/no-mutating-methods
-    datesArray.push(tempDate.clone().set({ hour: 12, minute: 0 }).valueOf());
+    datesArray.push(clone.setUTCHours(12, 0, 0, 0));
 
-    tempDate.add(1, 'days');
+    // eslint-disable-next-line fp/no-mutation
+    tempDate = new Date(tempDate.getTime() + oneDayInMs);
```
laurentS, line 73: It's a bit of a weird corner case, but I wonder how much of an impact the [leap second](https://en.wikipedia.org/wiki/Leap_second) has on our date management. It looks like there's only been 5 in the past 20 years, so likely nothing to worry about :crossed_fingers: 

That being said, I would be tempted to modify this `while` loop to add 1 day at a time with a counter, and do `datesArray.push(tempDate.setUTCDate(counter))` so you can avoid creating all these temp date clones.

frontend/src/utils/ews-utils.ts
```diff
@@ -99,16 +103,16 @@ export const fetchEWSDataPointsByLocation = async (
...
 ): Promise<EWSSensorData[]> => {
-  const endDate = moment(date)
-    .clone()
-    .set({ hour: 23, minute: 59, second: 59 });
+  const endDate = new Date(date);
+  endDate.setHours(23, 59, 59, 999);
   // FIXME: pass start/end here? why the 24h delta?
-  const startDate = endDate.clone().subtract(1, 'days');
+  const startDate = new Date(endDate.getTime() - oneDayInMs);
   const format = 'YYYY-MM-DDTHH:mm:ss';
```
laurentS, line 110: Can we create an enum of all the valid formats we use across the codebase to avoid having these random strings?

frontend/src/utils/server-utils.ts
```diff
@@ -302,21 +304,25 @@ export function generateIntermediateDateItemFromValidity(layer: ValidityLayer) {
...
   const dateItemsWithValidity = sortedDates
     .filter(date => date > earliestDate)
-    .map(d => moment(d).set({ hour: 12, minute: 0 }))
-    .reduce((acc: DateItem[], momentDate) => {
-      // We create the start and the end date for every moment date
-      let startDate = momentDate.clone();
-      let endDate = momentDate.clone();
+    .map(d => {
+      const date = new Date(d);
+      date.setHours(12, 0, 0, 0);
```
laurentS, line 314: Should this be `setUTCHours` to ensure the user's locale has no impact? (I guess the same comment applies to all the calls of this method).

frontend/src/utils/server-utils.ts
```diff
@@ -616,7 +622,16 @@ export function formatFeatureInfo(
   labelMap?: { [key: string]: string },
 ): string {
   if (type === DataType.Date) {
-    return `${moment(value).utc().format('MMMM Do YYYY, h:mm:ss')} UTC`;
+    return new Date(value).toLocaleString('default', {
```
laurentS, line 625: Why is this not using `getFormattedDate`? Would it make sense to use this code in that function instead of instantiating all those strings?

frontend/src/utils/ews-utils.ts
echaidemenos: we are setting hours to 12, so leap second can't change the date, i believe

ericboucher (approved):

PR #1125, opened 2024-02-16
remove validity for forecast data; COUNTRY=mozambique

wadhwamatic:
### Description

remove validity period for forecast layer in Mozambique

## Checklist - did you ...

Test your changes with

### DIFF
--- a/frontend/src/config/mozambique/layers.json
+++ b/frontend/src/config/mozambique/layers.json
@@ -5255,10 +5255,6 @@
     "additional_query_params": {
       "styles": "rfh_16_0_300"
     },
-    "validity": {
-      "days": 10,
-      "mode": "forward"
-    },
     "opacity": 0.7,
     "legend_text": "CHIRPS-GEFS global dekadal rainfall forecasts",
     "legend": [


wadhwamatic:
@ericboucher - there's a less than ideal implementation of rainfall forecast anomaly on the HDC side that I'm trying to get around to add this layer in PRISM.

The HDC layer `rfq_dekad` is a combination of a forecast and observation data. The latest available data is a forecast, while previous versions are observations. Older forecasts get overwritten by observations.

We want to add a rainfall forecast anomaly layer that only shows forecasts, i.e. the latest available. I created a layer for this `dekad_rainfall_anomaly_forecast`. If I set `"start_date": "today"`, no data is found because the latest as of now is 21-Feb. 

Can we add another `start_date` option that is something like `last_available` so we only show the most recent round?

Also, while trying to make this work for now, I set `start_date` to `02-21-2024`. I was expecting the format to be DD-MM-YYYY, not MM-DD-YYYY. Note to document this!

ericboucher (approved):

PR #1131, opened 2024-02-22
Sync opacity between layer and legend controls

### LINKED ISSUE #1080: Revamp opacity state management
### What happened?

As a follow up to https://github.com/WFP-VAM/prism-app/pull/1061 we should revamp the way we handle layer opacity.
Currently the state is not persisted between the menu item opacity slider and the legend opacity slider which leads to some weird behaviors. The logic also seems too complex with duplication of code and different subcases that could likely be avoided. As an example, the analysis layer has a slightly differeny opacity management methods although it does not seem necessary.

N/A

_No response_

echaidemenos:
### Description

This fixes #1080 .

Adds a central store for opacity, so it's value and way to update it is consistent between components.
@wadhwamatic one change is that the opaicty is persisted when we switch layers off/on. This is on puporse but can be changed if we prefer.

## How to test the feature:

- enable any layer and make sure the value on the legend is the same with the left panel
- generally play around with opacity

## Checklist - did you ...

Test your changes with


ericboucher (approved):

frontend/src/components/MapView/Legends/LegendItem/__snapshots__/index.test.tsx.snap
```diff
@@ -104,22 +104,21 @@ exports[`renders as expected 1`] = `
               />
               <span
                 class="MuiSlider-track"
-                style="left: 0%; width: 50%;"
+                style="left: 0%;"
               />
               <input
                 type="hidden"
-                value="0.5"
+                value="NaN"
```
ericboucher, line 111: is this expected @echaidemenos ?

frontend/src/components/MapView/LeftPanel/layersPanel/AnalysisLayerSwitchItem/index.tsx
```diff
@@ -40,11 +43,26 @@ const AnalysisLayerSwitchItem = memo(
...
-    const [opacity, setOpacityValue] = useState<number>(initialOpacity || 0);
+    const opacity = useSelector(opacitySelector('analysis'));
+
+    const setOpacity = useCallback(
+      (value: number) =>
+        dispatch(
+          setStateOpacity({
+            map,
+            value,
+            layerId: 'analysis', // We have to pass undefined here so that the function know that we are in an analysis custom layer
```
ericboucher, line 54: update comment

frontend/src/components/MapView/LeftPanel/layersPanel/AnalysisLayerSwitchItem/index.tsx
```diff
@@ -40,11 +43,26 @@ const AnalysisLayerSwitchItem = memo(
...
+            layerType: undefined, // We have also to pass undefined here to show that we have analysis custom layer,
```
ericboucher, line 55: might be worth creating an `analysis` type instead?

frontend/src/components/MapView/Legends/LegendItem/__snapshots__/index.test.tsx.snap
echaidemenos: Yes, initial value of the state is undefined, so the components know it is the first time referenced and they have to set it. That way we keep the value when toggling layers on and off

wadhwamatic (approved):
@echaidemenos - this works well. Persisting the opacity state after toggling layers off / on as you have implemented it is great. Thanks!

frontend/src/components/MapView/LeftPanel/layersPanel/AnalysisLayerSwitchItem/index.tsx
echaidemenos: I am not sure if we need to create an analysis layer type, it maybe worth investigating. For this case i extended `SetOpacityParams` to accept `'analysis'` as a type, so we don't have to use this tricky `undefined` there

ericboucher (approved):

PR #1132, opened 2024-02-23
Augment country mask quality and switch to clipping; COUNTRY=rbd

ericboucher:
### Description

Improves print mask boundary quality via enhanced preprocessing in `preprocess-layers.ts`:

1. **Higher fidelity outlines** — filters tiny islands/holes (≥ 10 km² threshold), unions admin boundaries, then simplifies with tighter tolerance (`0.001`, high quality).
2. **Complex geometries** — Mozambique is always reprocessed with this pipeline. Indonesia uses a manually curated `admin-boundary-unified-polygon.json` (preprocessing skipped).
3. **Clipping** - Switch from masking to clipping.

Previously considered external boundaries (cartographyvectors.com) for cambodia/ecuador/mozambique/myanmar/sierraleone; those countries now rely on the improved preprocessing pipeline instead.

## How to test the feature:

- [ ] Open print/export for a small country (e.g. cambodia, jordan, zimbabwe) and verify mask outline is precise with no triangular artifacts
- [ ] Test mozambique print mask — no stripe artifacts
- [ ] Test indonesia print mask — boundary aligns with admin layers

## Checklist - did you ...

Test your changes with


wadhwamatic:
@ericboucher - we can get admin0 files and add them during the config process going forward. Generally, the most reliable source is HDX, ideally datasets tagged with 'COD'. The format is usually going to be a shapefile.

https://data.humdata.org/dataset/?cod_level=cod-standard&cod_level=cod-enhanced&ext_administrative_divisions=1&q=&sort=last_modified%20desc&ext_page_size=25

I'll add Mozambique to this PR and revisit others soon.

wadhwamatic:
Hmmm. When I added an admin0 boundary for Mozambique using the HDX data, the stripe returns. 
Screenshot 2024-02-28 at 14 55 59

The issue with using the boundary you included @ericboucher is that they are not fully aligned with the other admin boundaries, as since this is for a print feature, fidelity matters. We shouldn't see boundaries like this:
Screenshot 2024-02-28 at 15 01 49

wadhwamatic:
@ericboucher - the last moz boundary file I tried seems to work well. File size went up, but it's still under 2mb.

Here's what I did in QGIS:
1. Dissolve admin1 layers
2. Delete holes (this gets rids of small bodies of water)
3. Multipart to single part. This seems to connect strays
4. Delete unused features
5. Export geojson w/ precision 6

wadhwamatic:
@ericboucher, since we're revisiting this, the current request is to have a more polished map published when we select an admin area in the print map feature. Instead of greying out content outside of the selected admin area, we would prefer to clip the raster data to the selected admin area and continue to have a basemap visible. This mirrors a typical map production workflow. See an example below.


moz-clipped

ericboucher:
@gislawill curious how you envision this working with PMtiles. See #1943 as well.

wadhwamatic:
@ericboucher - the boundary smoothing is working well. I still see occassional stripes at certain zoom levels. See examples from Mozambique and Indonesia

Screenshot 2026-06-11 at 10 43 39
Screenshot 2026-06-11 at 10 49 57

wadhwamatic:
@ericboucher - the lag does indeed look better on a deployed version. I'd like to test this on a few different machines to see how it looks.

I noticed a few other issues:
1. The standard export map (not a batch map) now allows admin clipping in the preview, but the downloaded file is not clipped
2. The exported batch map's legend has a resolution / scale issue. See screenshot of preview and exported PNG. This might exist in the main branch

Screenshot 2026-06-12 at 09 23 25
map_2026-06-01

wadhwamatic:
Another issue (potentially also on main): Sri Lanka admin 1 & admin 2 clipping does not work, but admin 3 does.

https://github.com/user-attachments/assets/8b104cab-c64b-4181-8c27-648898660836

gislawill:
Hey @ericboucher, I was running into performance issues when panning and zooming in the print preview of this branch. I tried out an approach to move the clipping functionality to the tile load process rather than the render process. Checkout #1951. I'm not sure it's fully baked but I think it's pretty close (performance might be a bit worse with vector layers). 

This approach works with PMTiles _if the feature_ is in the viewport when the admin unit is selected. We could add some error messaging around this to "handle" it sufficiently (imo).

Here's the before and after video:
https://github.com/user-attachments/assets/64ca92fe-b8f3-4131-9140-3905a815b604

wadhwamatic:
Thanks for addressing those issues @ericboucher and for the much improved performance @gislawill. I'll test once more when the legend fix is implemented. This is looking very good otherwise

PR #1138, opened 2024-02-26
Allow for both forward and backward validity ranges; COUNTRY=rbd

### LINKED ISSUE #1025: [Feature Request]: Allow for both forward and backward validity ranges

Data layers published by WFP's Humanitarian Data Cube (HDC - https://api.earthobservation.vam.wfp.org/ows/) have multiple time aggregations including 10-day, 1-month, 3-month, etc. 

The reference period for a data layer is straightforward for 10-day products. The date is the beginning of a 10-day period. So, 1-November would refer to 1-November through 10-November; 11-November would refer to 11-November through 20-November; and 21-November refers to 21-November through 30-November. For these products, I can use a validity of 10-day forward and get the result I need. It seems we're somehow also managing to handle 31-day months with this approach with this configuration:
```
    "validity": {
      "days": 10,
      "mode": "forward"
    }
```
However, 1-month aggregations and above are confusing on the HDC side. A 1-month aggregation published on 1-November refers to the beginning of the last dekad in the one month period AND the two dekads prior. So the coverage would be from 11-October through 10-November. As of now, there's no way to create this in the layer configuration.

I need to be able to configure something like:
```
    "validity_periods": [
      {
        "days": 10,
        "mode": "forward"
      },
      {
        "days": 20,
        "mode": "backward"
      }
    ]
```

_No response_

ericboucher:
### Description

This fixes #1025.

- Introduce the concept of dekad mode, as descibed in #1025  
- Allow for both forward and backward validity ranges;

TODO:
- [x] update other `layers.json` validity periods to the new format

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with


frontend/src/utils/server-utils.test.ts
```diff
@@ -168,8 +168,8 @@ describe('Test generateIntermediateDateItemFromValidity', () => {
         new Date('2023-12-11').setHours(12, 0),
       ],
       validity: {
-        days: 10,
-        mode: DatesPropagation.BACKWARD,
+        backward: 10,
+        mode: DatesPropagation.DAYS,
```
echaidemenos, line 172: We could add a test with `dekad` here

frontend/src/utils/server-utils.ts
```diff
@@ -302,17 +302,15 @@ async function generateIntermediateDateItemFromDataFile(
 
 export function generateIntermediateDateItemFromValidity(layer: ValidityLayer) {
```
echaidemenos, line 303: This is not directly linked to this PR, but since we are modifying this function we could remove the usage of `lodash` functions `sortBy` and `sortedUniqBy`

frontend/src/utils/server-utils.ts
```diff
@@ -546,7 +565,7 @@ export async function getLayersAvailableDates(
   const layersWithValidity: ValidityLayer[] = Object.values(LayerDefinitions)
     .filter(layer => layer.validity !== undefined)
     .map(layer => {
-      const layerId = layer.type === 'wms' ? layer.serverLayerName : layer.id;
+      const layerId = layer.id;
```
echaidemenos, line 568: nice catch here, i see that `mapServerDatesToLayerIds` always returns `layer.id` and not `layer.serverLayerName`

frontend/src/utils/server-utils.test.ts
ericboucher: Done ✅

PR #1142, opened 2024-03-01
Add an Anticipatory Action module

### LINKED ISSUE #1098: [Feature Request]: Add an Anticipatory Action module to the PRISM UI

This is a placeholder with more details to come. The Anticipatory Action (AA) module in PRISM will require a new 'tab' which would include a few key elements:

1. The 'trigger' status for each admin area at a given point in time, within a specific monitoring 'window' by severity category
2. The phases of a trigger include: no action, ready, set and go
3. The window is a key monitoring period of crop seasonal calendars - when rainfall amounts are key driver of food security outcomes
4. The trigger status is based upon a forecasted indicator and the 'skill' or reliability of that forecast. This is all managed by WFP's AA and Data Science team through automated data pipelines which then store outputs in CSV files
5. New forecast products are created every month (TBC), and the skill of a forecast generally improves for a shorter forecast window; i.e. the skill of a forecast 6 months for now is lower than the skill of a forecast 1 month from now
6. Users of the dashboard can drill down to districts (admin 2) to see additional details of the trigger over time

Much of the above content is in a few docs that we'll compile and make accessible. For now, it's worth looking more closely at the mockups available [here](https://www.figma.com/file/Xe8H1GuVMJUsIi4JNTYPVo/WFP-Anticipatory-Action-(AA)?type=design&node-id=14-383&mode=design&t=7fHYKJ3VHUvR0HXa-0) which are also pasted below in case that link is inaccessible.

National overview:
Screenshot 2024-02-02 at 21 09 54

District drill down 1:
Screenshot 2024-02-02 at 21 10 04

District drill down 2:
Screenshot 2024-02-02 at 21 10 21

_No response_

echaidemenos:
### Description

This fixes #1098 .

WIP:
So far only the main page layout is implemented.
Next steps can be:
- [ ] adapt CSV and fetch date from there
- [ ] create anticipatory action layer 
- [ ] create AA legend
- [ ] make left panel AA page interactive 

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with

Screenshot from 2024-03-01 17-26-35

### DIFF
--- a/frontend/src/components/MapView/LeftPanel/AnticipatoryActionPanel/HomeTable/index.tsx
+++ b/frontend/src/components/MapView/LeftPanel/AnticipatoryActionPanel/HomeTable/index.tsx
@@ -1,0 +1,162 @@
+import { Typography, createStyles, makeStyles } from '@material-ui/core';
+import { useSafeTranslation } from 'i18n';
+import { borderGray, gray } from 'muiTheme';
+import React from 'react';
+import { Phase } from '../utils';
+
+interface AreaTagProps {
+  name: string;
+  isNew: boolean;
+}
+
+function AreaTag({ name, isNew }: AreaTagProps) {
+  return (
+    <div
+      style={{
+        border: `1px solid ${borderGray}`,
+        height: '2rem',
+        borderRadius: '4px',
+        display: 'flex',
+        alignItems: 'center',
+        gap: '0.25em',
+        padding: '0 0.25em',
+      }}
+    >
+      <Typography>{name}</Typography>
+      {isNew && (
+        <div
+          style={{
+            height: '2em',
+            padding: '0 0.5em',
+            color: 'white',
+            background: '#A4A4A4',
+            fontSize: '10px',
+            borderRadius: '32px',
+            display: 'flex',
+            alignItems: 'center',
+          }}
+        >
+          NEW
+        </div>
+      )}
+    </div>
+  );
+}
+
+export interface RowProps {
+  id: Phase | 'header';
+  iconContent: React.ReactNode;
+  windows: [AreaTagProps[], AreaTagProps[]];
+  header?: string[];
+}
+
+function Row({ iconContent, windows, header }: RowProps) {
+  const classes = useRowStyles();
+  const { t } = useSafeTranslation();
+
+  if (header) {
+    return (
+      <div className={classes.rowWrapper}>
+        <div className={classes.iconCol}>{iconContent}</div>
+        {header.map(name => (
+          <div key={name} className={classes.windowWrapper}>
+            <Typography variant="h3" className={classes.headerText}>
+              {name}
+            </Typography>
+          </div>
+        ))}
+      </div>
+    );
+  }
+
+  return (
+    <div className={classes.rowWrapper}>
+      <div className={classes.iconCol}>{iconContent}</div>
+      {windows.map((col, index) => (
+        // we can actually use the index as key here, since we know each index is a window
+        // eslint-disable-next-line react/no-array-index-key
+        <div key={index} className={classes.windowWrapper}>
+          <div className={classes.windowBackground}>
+            <div className={classes.tagWrapper}>
+              {col.map(x => (
+                <AreaTag key={x.name} {...x} />
+              ))}
+              {col.length === 0 && (
+                <Typography className={classes.emptyText}>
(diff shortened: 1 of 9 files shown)


frontend/src/components/MapView/LeftPanel/AnticipatoryActionPanel/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+  );
+
+  // eslint-disable-next-line no-console
+  console.log(setMild, readyMild);
+
+  // TODO: implement NA and NY
+  // NA means that no proba is above the trigger for that district
+  // NY means that the district is not monitored yet (no rows for the district)
+  const na = React.useMemo(() => [], []);
+  const ny = React.useMemo(() => [], []);
```
ericboucher, line 109: @echaidemenos I think we need to revise this logic a bit and maybe do some pre-processing.
I added some explanations around what NA - No Action and NY - Not Yet Monitored mean.

I also added some filtering to only keep rows were proba > trigger

We also need to include Leve = "MILD" now, so just adding two rows.

And here is some additional info about the logic: to pass to “set”, the trigger must have been “ready” the preceding month. Once Severe Set is attained, the status stays the same for the rest of the window.

Can you propose a better way to do the data preparation? Also in general I think that logic such as language changes should happen as soon as we load the CSV so that the rest of the code does not have to worry about it. Wdyt?

frontend/src/components/MapView/LeftPanel/AnticipatoryActionPanel/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+      iconContent: AAIcons.ny,
+      windows: windowData(ny),
+    },
+  ];
+
+  return (
+    <div className={classes.anticipatoryActionPanel}>
+      {/* header */}
+      <div className={classes.headerWrapper}>
+        {/* Title area */}
```
ericboucher, line 174: It's nice to add comments but some of these feel a bit superfluous.

echaidemenos: Yes i added them when i originally create the layout. now they don't really serve any purpose, i am removing them

frontend/src/components/MapView/LeftPanel/AnticipatoryActionPanel/HomeTable/index.tsx
```diff
@@ -257,34 +280,58 @@ function HomeTable() {
...
     id: -1,
     iconContent: null,
     windows: selectedWindow === 'All' ? AAWindowKeys.map(x => []) : [[]],
     header: selectedWindow === 'All' ? [...AAWindowKeys] : [selectedWindow],
   };
 
-  const dataForRows = React.useMemo(
-    () =>
-      rowCategories.map(r => ({
+  const dataForRows: {
```
ericboucher, line 290: It feels like this whole processing logic, and `getDistrictData` in particular should be closer to where we load the data so that processing happens only once.

Let's try and do that as we transition to the new file

frontend/src/context/anticipatoryActionStateSlice/index.ts
```diff
@@ -0,0 +1,367 @@
...
+      if (aOrder > bOrder) {
+        return -1;
+      }
+      if (aOrder < bOrder) {
+        return 1;
+      }
+      return 0;
+    });
+
+    const filtered = sortedData.filter(x => categoryFilters[x.category]);
```
ericboucher, line 211: This is most often NA I think but we can come back to this later.
This process also does not take into account the fact that once SET is attained it stays available. Or am I missing something? Is there not a way to reuse what's calculated for dataForRows in the HomeTable?

frontend/src/components/MapView/LeftPanel/AnticipatoryActionPanel/HomeTable/index.tsx
ericboucher: Solved in #1170

frontend/src/context/anticipatoryActionStateSlice/index.ts
ericboucher: Solved in #1170

ericboucher (approved):

PR #1143, opened 2024-03-05
update rounding rule for % calculations; COUNTRY=rbd

wadhwamatic:
### Description

Update CD data for RBD to match rounding rules used in official reporting

## Checklist - did you ...

Test your changes with


frontend/public/data/rbd/ch/al_admin2_2020_10_01.json
```diff
@@ -4,6 +4,9 @@
             "adm0_name": "Niger",
             "adm0_pcod2": "NE",
             "adm0_pcod3": "NER",
+            "adm0_name.1": "Niger",
+            "adm0_pcod2.1": "NE",
+            "adm0_pcod3.1": "NER",
```
ericboucher, line 9: is this change expected?

ericboucher (approved):

frontend/public/data/rbd/ch/al_admin2_2020_10_01.json
wadhwamatic: weird. I'm not sure how that got introduced. will look into it.

ericboucher: on a more general note, the CH files are quite big, so it's nice if we can limit the data to what we actually need to limit their size and impact on performance

wadhwamatic: Yes, we can trim data. I kept all fields for now even though they're not all being used as it's still being fleshed out a bit

wadhwamatic:
@ericboucher I created an issue to address file size. Will get to that early April

PR #1158, opened 2024-03-16
Use a single AA geojson layer

ericboucher:
### Description

Simplify the structure of the AA map layers. @echaidemenos wdyt?

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with


frontend/src/components/MapView/Layers/AnticipatoryActionLayer/index.tsx
```diff
@@ -103,46 +103,62 @@ function AnticipatoryActionLayer({ layer, before }: LayersProps) {
...
+            ...feature,
+            properties: {
+              ...feature.properties,
+              fillColor: color || 'grey',
+            },
+          };
+        })
+        .filter(f => f !== null),
+    };
+  }, [data, renderedDistricts, layer.csvWindowKey]);
```
ericboucher, line 137: I am not sure I understand why we need to use layer.csvWindowKey in here?
Why are we not simply returning the final prioritized list of district with their max level in `renderedDistricts` ?
It feels like that's what's creating the issue here as `layer.csvWindowKey` seems to be out of sync

frontend/src/context/anticipatoryActionStateSlice/index.ts
```diff
@@ -174,6 +174,93 @@ export const loadAAData = createAsyncThunk<
...
+      }
+      if (aOrder < bOrder) {
+        return 1;
+      }
+      return 0;
+    });
+
+    const filtered = sortedData.filter(x => categoryFilters[x.category]);
+
+    // let the first window layer to render empty district
```
ericboucher, line 213: This is most often NA I think but we can come back to this later.
This process also does not take into account the fact that once SET is attained it stays available. Or am I missing something? Is there not a way to reuse what's calculated for `dataForRows` in the HomeTable?

PR #1161, opened 2024-03-18
Create QML style file from legend

### LINKED ISSUE #1100: [Feature Request]: Give users the option to download a style file when downloading a GeoTIFF

When a user downloads a GeoTIFF through PRISM, they get a raw data file without the same style attributes that were used in the PRISM frontend. If they want to create a map using this data layer, they would have to manually configure the styles in their desktop GIS software.

The legend configuration in PRISM has the essential information needed to generate a GIS style file. The format we often use in our team is QML which is specific to QGIS. I believe SLD is a standard across GIS systems, while .stylx is used in ESRI software. To start, we should try to support QML and/or SLD.

The UI could be updated to either show two buttons:
Screenshot 2024-02-02 at 21 53 53

Or a single button with a checkbox (default to unchecked):
Screenshot 2024-02-02 at 21 54 44

Open to suggestions on the UI!

A tooltip that shows on hover over the 'with style' options should say: "Choose this option to apply the same style you see now in your desktop GIS software."

If a user chooses to include the style file, we should send a zip package with the GeoTIFF and the style file(s) in the formats mentioned above.

_No response_

ericboucher:
### Description

This fixes #1100

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with

Screenshot 2024-03-18 at 2 47 34 PM


wadhwamatic:
@ericboucher - here's an example qml file. [RFH_16_0_300.qml.zip](https://github.com/WFP-VAM/prism-app/files/14638163/RFH_16_0_300.qml.zip)

ericboucher:
@wadhwamatic the file generation works now. I am not getting the expected result in QGIS but it looks like the data itself is scaled for some reason?

wadhwamatic:
> @wadhwamatic the qml file generation works now. I am not getting the expected result in QGIS but it looks like an issue with the data itself. The geotiffs seem scaled... Eg. for precip_blended_dekad for March in Moz, the data shows rainfall data above `100mm`, but if we run an analysis, the max value in the file is `2`...

That issue should be fixed now that the cache with bad data was cleared

ericboucher:
@wadhwamatic this PR is gtg from my side if you're happy with the resulting QML.

wadhwamatic:
I got an error when I load the style in QGIS for Mozambique, NDVI anomaly. 

The error is from the the legend entry that starts with "<", which is annoying as it's just a label and for some reason the issue doesn't exist for ">". Thoughts on a fix for this?
Screenshot 2024-03-24 at 14 41 05

I think we can safely remove the version reference in the QML file too (`version="3.8.3-Zanzibar"`) and we should test on a few versions of QGIS as well.

ericboucher:
HOLD -  There is a slight issue with how we defined the legends so far that creates a shift in colors because of the way qml uses the values. Here is an example:

[
        “value”: 0,
        “label”: “< 20%“,
        “color”: “#8C4800"
      },
      {
        “value”: 20,
        “label”: “20-40%“,
        “color”: “#AF6B27"
      },
]
When qml assumes 0 as the start and uses the top value instead so it should be:
[
        “value”: 20,
        “label”: “< 20%“,
        “color”: “#8C4800"
      },
      {
        “value”: 40,
        “label”: “20-40%“,
        “color”: “#AF6B27"
      },
]
…
I propose that we revise how we define legends to match how qml works and align with how other GIS tools implement legends.

ericboucher:
Let's wait on #1157 as it introduces shared legends which will make this a lot easier

frontend/src/components/MapView/LeftPanel/layersPanel/MenuItem/MenuSwitch/SwitchItem/LayerDownloadOptions.tsx
```diff
@@ -138,6 +139,78 @@ function LayerDownloadOptions({
...
+        : item.value.toString();
+
+      // TEMPORARY: shift the value index by 1 to account for the 0 value
+      // and match the QML style format. See https://github.com/WFP-VAM/prism-app/pull/1161
+      const shouldShiftIndex =
+        legend[0].value === 0 &&
+        ((legend[0].label as string)?.includes('<') ||
+          (legend[1].label as string)?.includes('<') ||
+          (legend[1].label as string)?.includes('-') ||
+          (legend[1].label as string)?.includes(' to '));
```
ericboucher, line 172: @wadhwamatic I added this logic. I looked through the layers.json to find the different patterns that could lead us to shift the legends. Not 100% I got everything right but we can adapt it in the future if we notice that a specific layer is off

wadhwamatic:
@ericboucher - this looks good, but I found a separate issue where the scale value isn't getting applied and so the resulting map produced locally doesn't match PRISM. This is strange as I'm sure this worked correctly at an earlier stage.

For example, here's the blended SPI for Mozambique
Screenshot 2024-05-22 at 16 54 42

And here's what you see in QGIS for the downloaded raster and QML (see histogram values too)
Screenshot 2024-05-22 at 17 02 07

ericboucher:
@wadhwamatic it's probably because of the switch to STAC. I can rescale the style

wadhwamatic (approved):
@ericboucher - thanks for the scaling update. Good to go now!

Issue #360, opened 2022-01-21
Don't require 'admin_boundaries' as the name of boundary layer


wadhwamatic:
It is confusing to force use of the name 'admin_boundaries' especially as we will have the concept of 'isPrimary' forthcoming with PR https://github.com/WFP-VAM/prism-frontend/pull/338. Not urgent, but I would like to remove this restriction sooner than later

PR #1163, opened 2024-03-19
Select admin level for analysis

### LINKED ISSUE #1052: [Feature Request]: Let users choose admin level when running an analysis

Currently if you choose to run an analysis using admin boundaries, PRISM will use the default boundary layer. But users may not want to aggregate data at that level. For example, the default boundary may be at admin 2 (districts) but a user might want results at admin 1 (province) level. 

We should add an option that allows users to optionally choose an admin level when they select admin boundaries as the baseline layer in the run analysis feature. Ideally, the selection would use a dropdown menu with levels available using a name / label for each option.

An issue is that we don't have a label for admin levels. i.e., there's currently no way to specify that admin 1 refers to 'provinces' and admin 2 refers to 'districts', and these labels can change per country. We could add an admin level label concept to layers.json for type 'boundary'. 

For example:
{
  "admin1_boundaries": {
    "type": "boundary",
    "path": "data/mozambique/moz_bnd_adm1_WFP.json",
    "opacity": 0.8,
    "admin_code": "adm1_source_id",
    **_"label": "province",_**
    "admin_level_codes": [
      "adm1_source_id"
    ],
    "admin_level_names": [
      "adm1_name"
    ],
    "admin_level_local_names": [
      "adm1_name"
    ],
    "styles:": {
      "fill": {
        "fill-opacity": 0
      },
      "line": {
        "line-color": "gray",
        "line-width": 3,
        "line-opacity": 0.8
      }
    }
  },
  "admin_boundaries": {
    "type": "boundary",
    "path": "data/mozambique/moz_bnd_adm2_WFP.json",
    "opacity": 0.8,
    "admin_code": "adm2_source_id",
    **_"label": "district",_**
    "admin_level_codes": [
      "adm1_source_id",
      "adm2_source_id"
    ],
    "admin_level_names": [
      "adm1_name",
      "adm2_name"
    ],
    "admin_level_local_names": [
      "adm1_name",
      "adm2_name"
    ],
    "styles:": {
      "fill": {
        "fill-opacity": 0
      },
      "line": {
        "line-color": "gray",
        "line-width": 1,
        "line-opacity": 0.8
      }
    }
  },

Selection could look like this:
select-admin-level

ericboucher:
### Description

This fixes #1052.

The current changes allow users to run an analysis selecting a different admin level.
@wadhwamatic we need to decide what happens once we run the analysis. Do we change the boundary layers like we do for admin level data? Or do we indicate in the title what level the data is valid for? Let me know what you prefer.

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with

Screenshot 2024-03-19 at 10 18 06 PM

Screenshot 2024-05-21 at 8 28 30 PM

Screenshot 2024-05-21 at 8 35 30 PM


wadhwamatic:
> we need to decide what happens once we run the analysis. Do we change the boundary layers like we do for admin level data? Or do we indicate in the title what level the data is valid for? Let me know what you prefer.

Yes, we should do the same as admin_level layers that are bound to a level besides the default.

ericboucher:
@wadhwamatic bringing from Slack for reference.

Eric: I think I disagree… I actually feel like we should change the behavior for the admin level data. Here are a few reasons:
- It makes it impossible to overlay multiple admin level data and usually the level is pretty obvious from the coloring
- It complicates the admin boundary logic a lot (when it is already quite so 😅  )

ericboucher:
NOTE - for some unknown reason this PR is breaking the report logic in the backend

ericboucher:
@wadhwamatic I think this also fixes #360

wadhwamatic:
This is great, thanks @ericboucher. When we revisit boundaries in https://github.com/WFP-VAM/prism-app/issues/1212, we should also consider giving a name to the levels as part of the config process, i.e. level 0 = country; level 1 = province, etc. and using those names in the drop down list here

wadhwamatic (approved):

PR #1169, opened 2024-03-20
Refactor charts code

### LINKED ISSUE #1111: Charts load time
It seems that charts are loading much faster on Firefox than Chrome, even when working with new requests.

Let's investigate where the difference could stem from between the two and improve performance on Chrome if possible.

Bhavyajain21:
### Description

This fixes #1111 
/claim #1111 

I've added two new useCallback hooks, handleDownloadChartPng and handleDownloadCsv, to handle the download of PNG and CSV files respectively. These hooks are more efficient because they have a stable reference, preventing unnecessary re-renders.

Additionally, the rendering logic for the chart component has been optimized. We only render either a Bar or Line chart based on the config.type property, avoiding unnecessary conditional rendering.

These changes aim to optimize performance and improve code readability.

While these optimizations are contributing to better load time balance between Chrome and Firefox, it's important to note that browser performance can be influenced by various factors beyond the control of the application code. Factors such as browser engine differences, hardware capabilities, network conditions, and browser extensions can all impact performance.


ericboucher:
@Bhavyajain21 Thanks for your work, could you make sure that the linter and type checks are passing please? As a bonus this will create a QA deploy automatically.

Bhavyajain21:
> @Bhavyajain21 Thanks for your work, could you make sure that the linter and type checks are passing please? As a bonus this will create a QA deploy automatically.

@ericboucher, yes I have been trying to do that but the approcahes haven't worked much from my side.

Bhavyajain21:
Thanks for the commit @ericboucher

Bhavyajain21:
@ericboucher Can you please review this?

ericboucher:
@Bhavyajain21 thanks for your work, I will try to finish my review soon but it's a bit hard to QA.
I see how the refactoring makes the code slightly simpler but I do not see how this would solve the performance issue we were seeing.
Aditionnally, the code was not working before my commit so I am not sure to what extent you were able to test this. Have you been able to test a before/after load time on Chrome vs FIrefox?

frontend/src/components/Common/Chart/index.tsx
```diff
@@ -125,79 +120,73 @@ const Chart = memo(
...
-    const labels = React.useMemo(() => {
-      if (!transpose) {
-        return indices.map(index => header[index]);
-      }
-      return tableRows
-        .slice(chartRange[0], chartRange[1])
-        .map(row => row[config.category]);
-    }, [chartRange, config.category, header, indices, tableRows, transpose]);
+    const labels = useMemo(() => {
+      return transpose
```
ericboucher, line 126: @Bhavyajain21 in some cases in this PR you are changing to ternary tetss and in others you are changing a ternary to an if statement, can you explain the rationale behind these changes?

ericboucher (approved):

Bhavyajain21:
Hey @ericboucher, I've been busy from past few days. Sorry for the delayed response.

PR #1189, opened 2024-04-09
Add new layers; COUNTRY=rbd

wadhwamatic:
### Description

Add new layers to RBD and allow point_data layers to be displayed as hexagons.
Add a new round of CH data

TODO
- [x] review point_data layer changes and make them reusable

## How to test the feature:

- [ ] Load new layers (Future of the Lean Season, Cropland Change Analysis, Conflict Analysis)
- [ ] Verify new round of CH data (Feb - May 2024)

## Checklist - did you ...

Test your changes with


wadhwamatic:
@ericboucher - I experimented with the H3 resolution. The ideal state would be to have a more dynamic resolution based on the zoom level. A resolution of 7 is better for data visualization but doesn't appear at the default zoom level. I think the hierarchy functions in H3 could help address this. What do you think? https://h3geo.org/docs/api/hierarchy

ericboucher:
@wadhwamatic possible but a bit tricky. Usually point data like this needs to be in grid structure reasy to be used as hex which does not seem to be the case here?

wadhwamatic:
@ericboucher 

> Usually point data like this needs to be in grid structure ready to be used as hex with a specific resolution based on the distance between points, which does not seem to be the case here?

That's correct, this is point data, not gridded.

> I can simply adjust the resolution with the zoom level for now maybe?

Not sure what you mean - by changing the zoom for RBD overall, or by adjusting the zoom if this layer is selected? If the latter, that works well!

wadhwamatic:
@ericboucher - I added a second point layer for the crop analysis. There's one for 2023 vs 2022, and another for 2023 vs reference years. If I select a second layer from the group, both appear. Instead, we should only show one layer at a time. Hope that's an easy fix.

wadhwamatic:
@ericboucher - this looks good from my side. I tested all new layers in RBD including the cropland change, and also verified that point layers in cambodia still work as expected. please go ahead and merge when you're ready

ericboucher:
Not "related" but one thing I noticed reviewing the PR is that the data files for RBD are getting quite big. I would recommend:
- removing all unnecessary fields
- removing key/values where the value is `null`
- minify JSON files

ericboucher (approved):

PR #1194, opened 2024-04-15
Fix map image would not show in report; COUNTRY=cambodia 

echaidemenos:
Fixes an issue where the map image would not appear in the flood report

#### How to test
- Use country cambodia
- Activate layer Flood / Flood Monitoring / Flood extent and run analysis
- Click on "Preview Report (slow)" button and verify the report


ericboucher (approved):

ericboucher:
Nice catch, I think that's my mistake srry

PR #1204, opened 2024-04-19
Add location name to chart title

### LINKED ISSUE #1195: [Feature Request]: Add location name to chart

When users download charts as a PNG, the context of the location is not specified. We should include the location name along with the indicator so that it can be brought into a report or other outputs while retaining context.

The name could be listed before the indicator in the title, or below the x-axis. The image below shows both, but we only need one or the other.

burkina_faso_boucle_du_mouhoun_10_day_rainfall_estimate_mm

_No response_

echaidemenos:
### Description

This fixes #1195 .

## How to test the feature:

- Navigate to charts panel
- Notice country name is in the title if no admin is selected, else the admin's name is displayed 

## Checklist - did you ...

Test your changes with

Screenshot 2024-04-20 at 10 13 48 AM
Screenshot 2024-04-20 at 10 14 14 AM
Screenshot 2024-04-20 at 10 15 00 AM


ericboucher:
@wadhwamatic let us know what you think, we can also put it on a second line like this:
Screenshot 2024-04-20 at 10 26 32 AM

And we could remove the title above the charts in comparison mode. Though it could still have some values since it shows the whole set of admin levels.

wadhwamatic:
> We could remove the title above the charts in comparison mode. Though it could still have some values since it shows the whole set of admin levels.

@ericboucher - let's keep the title above the charts to keep the full hierarchy in the display. 

I tried this for Cambodia. Couple of things:
1. Not specific to this PR, but the translation didn't show up as an option because it wasn't configured in shared. I thought it would still pull from the local config but did not for some reason.
2. After I added Khmer to the shared folder I could test and see that the admin name doesn't translate in the chart:
Screenshot 2024-04-22 at 09 19 50

ericboucher:
For 1. I'll push a fix, I had a fix locally on another branch but can't find it.

@echaidemenos can you make sure the admin name gets translated?

frontend/src/utils/admin-utils.ts
```diff
@@ -63,6 +63,7 @@ export const getChartAdminBoundaryParams = (
         code: properties[item.id],
         level: item.level,
         name: properties[item.name],
+        localName: properties[`${item.name}_`] || properties[item.name],
```
ericboucher, line 66: what's with the `${item.name}_` ?

echaidemenos: I have updated `levels` so we get the correct property each time

ericboucher (approved):

PR #1209, opened 2024-04-29
Add dropdown to select map style

echaidemenos:
### Description

This PR adds a drop-down on the navbar to select the map style.

Implement this suggestion: https://github.com/WFP-VAM/prism-app/issues/1199#issuecomment-2077806107

TODO: remove the "street" styles, they are just for test the feature is working. Find working satellite styles.

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with

image

### DIFF
--- a/frontend/src/components/MapView/Map/index.tsx
+++ b/frontend/src/components/MapView/Map/index.tsx
@@ -23,7 +23,7 @@ import {
   getLayerMapId,
   isLayerOnView,
 } from 'utils/map-utils';
-import { mapSelector } from 'context/mapStateSlice/selectors';
+import { mapSelector, mapStyleSelector } from 'context/mapStateSlice/selectors';
 import {
   AdminLevelDataLayer,
   BoundaryLayer,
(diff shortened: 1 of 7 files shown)


ericboucher:
We need to update the `firstSymbolId` logic to account for the fact different layers have different symbol layer ids

wadhwamatic:
@ericboucher - I'm not sure what the `firstSymbolId` logic does. Can you explain?

I think that the order might need to be configurable. In this example, if we change the base to satellite, and try to show layers, they do not appear presumably because the satellite layer is on top. This includes the boundary layer which is hidden

I also see that toggling back and forth between styles breaks rendering of admin level data.

wadhwamatic:
@ericboucher - ok understood. Are you suggesting we create a separate PR to address `firstSymbolId`? I'm ok with that approach. Let's just make sure it's captured in an issue so we don't lose track of this.

ericboucher:
@wadhwamatic this is ready for review. However, we currently have the extra map styles in `shared` which means this functionality is active for all countries. This doesn't feel like what we would want righ?
Also, I'm not a fan of the world icon which is often used for languages and recommend using the layers icon instead but that's up to you.

Screenshot 2024-08-01 at 2 46 20 PM

ericboucher:
I meant more from a UX side, the current approach takes a bit too much space for something that might not be used much?

PR #1236, opened 2024-05-04
Initial configuration for COUNTRY=bhutan

wadhwamatic:
### Description

This fixes #issueNumber.

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with


frontend/src/config/bhutan/layers.json
```diff
@@ -0,0 +1,288 @@
...
+        "fill-opacity": 0
+      },
+      "line": {
+        "line-color": "gray",
+        "line-width": 1,
+        "line-opacity": 0.8
+      }
+    }
+  },
+  "rainfall_dekad": {
```
ericboucher, line 40: Let's rely on the shared config as much as possible here.
If we are hindered by admin areas/boundaries, let's make a note of it in https://github.com/WFP-VAM/prism-app/issues/1212

frontend/src/config/bhutan/prism.json
```diff
@@ -0,0 +1,188 @@
+{
+  "country": "Bhutan",
+  "defaultLanguage": "en",
+  "countryAdmin0Id": 31,
+  "alertFormActive": false,
+  "hidePanel": false,
+  "enableNavigationDropdown": true,
+  "icons": {
```
ericboucher, line 8: Same, let's rely on the shared config as much as possible for things like that. Don't hesitate to add to the shared/prism.json

ericboucher (approved):

frontend/src/config/bhutan/layers.json
wadhwamatic: @ericboucher - I added a comment in 1212 as suggested. I could add `chart_data` to `shared/layers.json` but the levels fields will be wrong. Should we add it anyway, or just skip that element in the shared config, knowing either will break the chart feature unless the local config addresses it?

ericboucher: let's keep it as you did for now I think, it's clearer

PR #1252, opened 2024-05-16
specify admin boundary for admin level data; COUNTRY=mozambique

wadhwamatic:
### Description

This fixes an error that was introduced when Mozambique's default was set to admin3 instead of admin 2

## How to test the feature:

- [ ] load vulnerability / exposure layers 
- [ ] verify that data loads as expected

## Checklist - did you ...

Test your changes with

- [ ] `REACT_APP_COUNTRY=rbd yarn start`
- [ ] `REACT_APP_COUNTRY=cambodia yarn start`
- [ ] `REACT_APP_COUNTRY=mozambique yarn start`
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

PR #1253, opened 2024-05-20
Updating CDI to use month range; COUNTRY=jordan

lowriech:
### Description
This PR fixes updates our CDI queries to take a single month range. This is also adds support for calculating CDI using different weights in Jordan (CDI v1 vs CDI v2).

## How to test the feature:
- [ ] Open the CDI layers in Jordan 
- [ ] Load for v1 and v2 on different dates

KNOWN ISSUES: 
- Due to data availability lag, this calculation only works for dates 6+ weeks in the past
  - Long-term fix is tracked here: https://github.com/WFP-VAM/hip-service/issues/12
  - Fix for demo is here: https://github.com/WFP-VAM/prism-app/pull/1295
- Doesn't support seasonal ranges
  - Support is being added here: https://github.com/WFP-VAM/prism-app/pull/1301
- It's slow
  - This will likely be eventually moved so not worrying about it now

## Checklist - did you ...
Test your changes with


frontend/src/context/layers/composite_data.ts
```diff
@@ -13,26 +13,25 @@ export const fetchCompositeLayerData: LazyLoader<CompositeLayerProps> = () => as
...
   { dispatch },
 ) => {
-  // to complete later with new endpoint for composite chart
-
   const { layer, date } = params;
-  const endDate = (date ? new Date(date) : new Date())
-    .toISOString()
-    .split('T')[0];
-  const { baseUrl, inputLayers, startDate } = layer;
+  const startDate = date ? new Date(date) : new Date();
```
ericboucher, line 17: To be discussed but my first guess would have been to do the opposite and use the selectedDate as the endDate and then startDate = selectedDate - 1month

ericboucher: This is actually the way it's done usually in HDC apparently so let's keep it like that.

ericboucher:
@lowriech changes look good to me. Let's implement a switch between seasonal and one-month periods and this should be gtg.

frontend/src/context/layers/composite_data.ts
```diff
@@ -13,26 +14,25 @@ export const fetchCompositeLayerData: LazyLoader<CompositeLayerProps> = () => as
...
+    begin: getFormattedDate(startDate, 'default'),
+    end: getFormattedDate(endDate, 'default'),
     area: {
       min_lon: boundingBox[0],
       min_lat: boundingBox[1],
       max_lon: boundingBox[2],
       max_lat: boundingBox[3],
       start_date: '2002-01-01',
-      end_date: endDate,
+      end_date: '2021-07-31',
```
ericboucher, line 35: Are you sure this should be fixed?

frontend/src/context/layers/composite_data.ts
```diff
@@ -13,26 +14,25 @@ export const fetchCompositeLayerData: LazyLoader<CompositeLayerProps> = () => as
...
-    begin: startDate,
-    end: endDate,
+    begin: getFormattedDate(startDate, 'default'),
+    end: getFormattedDate(endDate, 'default'),
     area: {
       min_lon: boundingBox[0],
       min_lat: boundingBox[1],
       max_lon: boundingBox[2],
       max_lat: boundingBox[3],
       start_date: '2002-01-01',
```
ericboucher, line 34: You might want to replace this by your layer configured start_date?

frontend/src/context/layers/composite_data.ts
ericboucher: If yes, then it should probably be configured inside the layers.json

ericboucher: and used similarly to `areaStartDate`

wadhwamatic:
@lowriech - I tested this through the surge PR deployment. Here's my feedback:
1. I got 'failed to fetch message' and see a 500 error when the latest available date tried to load (May-11 as of now). An older date works though.
2. When I change dates, we're not removing the previous CDI from the map. We should do that. The same behavior exists for other layers in PRISM
3. FYI, loading time was between 10 - 20 seconds

I then added the cdi_v1 config as a layer for Mozambique to test there. 
1. The first load of a CDI layer took 4 mins and 10 secs
2. Second load took 30 seconds

Let me know when you've had a chance to address the first two points for Jordan and I'll test both scenarios again. Thanks!

wadhwamatic:
@lowriech, the last commit @ericboucher pushed addressed point 2: the previous CDI layer now correctly unloads when a new date is requested. Thanks Eric!

ericboucher:
Note - I think we should get rid of the `REACT_APP_USE_LOCAL_HIP_SERVICE` logic before merging to master

wadhwamatic:
@lowriech & @gislawill - between you, can we please address @ericboucher's comments that were shared on May 21st and yesterday?

frontend/src/context/layers/composite_data.ts
gislawill: Yea, I agree with Eric — the end date management should be pulled from the layer.json. I can shift it over now. 

I'll leave the fallback start date at '2002-01-01', I'm not confident I know what to test to ensure they're aren't regressions from removing it

gislawill:
> Note - I think we should get rid of the `REACT_APP_USE_LOCAL_HIP_SERVICE` logic before merging to master

This is now removed ✅

frontend/src/context/layers/composite_data.ts
gislawill: @lowriech, could you weigh in on the ETA on the hip_service side? Feel free to leverage me if I can useful in that

gislawill:
Heads up, I'm seeing these toggles (SPI, 10 day NDVI, etc) linked between both v1 and v2. I think it'd make more sense for these toggles to only be enabled if the top-level toggle is enabled. This doesn't appear to have been introduced in this PR so I'll create a fix in another branch
https://www.loom.com/share/7bcdf5fd8a284446a5a6ba1a64c3cd8e?sid=2b4c7802-9b58-487a-b44a-3611e887f507

gislawill:
> @lowriech changes look good to me. Let's implement a switch between seasonal and one-month periods and this should be gtg.

I'll have an update for this shortly

frontend/src/context/layers/composite_data.ts
lowriech: @ericboucher @gislawill  hip_service is good to go on this.  It just responds to the config.  And having it be more dynamic poses some issues, namely that it won't be able to take advantage of the cache.

gislawill:
> > @lowriech changes look good to me. Let's implement a switch between seasonal and one-month periods and this should be gtg.
> 
> I'll have an update for this shortly

@ericboucher and @wadhwamatic, I opened a draft PR with this update to support a seasonal CDI period. It wound up a bit larger than I expected. I posted some open questions in there. Most crucially, the request to hip service is failing when the bounds are 3 months rather than one month — I'm continuing to dig into that with @lowriech's help.

gislawill:
CDI next steps after discussion with Amit + Chris:
- [ ] Pull in master to #1252
- [ ] Prep #1253 for review (description, comments)
- [ ] Merge #1253
- [ ] Pull in master to #1301
- [ ] Update #1301 with validity (above) and additional configs
- [ ] Update #1295 with latest
- [ ] Create follow-up PR against master that adds finer grain season definitions (s1_start, s1_end)

wadhwamatic:
> CDI next steps after discussion with Amit + Chris:

Thanks for capturing these notes @gislawill. Please let me and @ericboucher know when we've got the go ahead to test this PR

gislawill:
@ericboucher and @wadhwamatic, this is now ready for review. Note: I was getting an intermittent error from hip-service.ovio.org/q_multi_geojson when running this locally. It seems to have now resolved itself? Is this a known issue?

Screenshot 2024-07-10 at 6 00 15 PM

wadhwamatic:
> @ericboucher and @wadhwamatic, this is now ready for review. Note: I was getting an intermittent error from hip-service.ovio.org/q_multi_geojson when running this locally. It seems to have now resolved itself? Is this a known issue?
> 
> Screenshot 2024-07-10 at 6 00 15 PM

@gislawill - thanks for flagging this. I haven't seen this specific error before and wasn't able to replicate it on my side. 

This is GTG from my side. @ericboucher - over to you

wadhwamatic (approved):

frontend/src/context/layers/composite_data.ts
```diff
@@ -12,26 +13,30 @@ export interface CompositeLayerData extends FeatureCollection {}
...
-    // to complete later with new endpoint for composite chart
-
     const { layer, date } = params;
-    const endDate = (date ? new Date(date) : new Date())
-      .toISOString()
-      .split('T')[0];
-    const { baseUrl, inputLayers, startDate } = layer;
+    const startDate = date ? new Date(date) : new Date();
+    const endDate = new Date(startDate);
+    endDate.setMonth(endDate.getMonth() + 1);
```
ericboucher, line 19: Maybe a small comment could be useful here

frontend/src/context/layers/composite_data.ts
```diff
@@ -41,25 +46,22 @@ export const fetchCompositeLayerData: LazyLoader<CompositeLayerProps> =
...
+        headers: {
+          Accept: 'application/json',
+          'Content-Type': 'application/json',
         },
-        `Request failed for fetching boundary layer data at ${baseUrl}`,
-      );
+      });
+
+      // eslint-disable-next-line no-console
+      console.log('Request config used for Qmulti:', {
```
ericboucher, line 61: We can probably get rid of this now? It's always in the network requests if we need it

ericboucher (approved):

frontend/src/context/layers/composite_data.ts
```diff
@@ -12,26 +13,30 @@ export interface CompositeLayerData extends FeatureCollection {}
...
+      end: getFormattedDate(endDate, 'default'),
       area: {
         min_lon: boundingBox[0],
         min_lat: boundingBox[1],
         max_lon: boundingBox[2],
         max_lat: boundingBox[3],
-        start_date: '2002-01-01',
-        end_date: endDate,
+        start_date: areaStartDate ?? '2002-01-01',
+        end_date: areaEndDate ?? '2021-07-31',
```
ericboucher, line 39: Not sure I fully understand so bear with me. But I would be expecting this to take the max of a default and endDate, so max('2021-07-31', endDate).
But maybe we actually want the baseline to be stable over time so 🤷

frontend/src/context/layers/composite_data.ts
gislawill: Yes, definitely

frontend/src/context/layers/composite_data.ts
gislawill: I _think_ the max of 2021-07-31 would be mainly relevant for CDI and not necessarily all future composite layers (but I could totally be wrong about that). 

In general, I'd say we should limit the role of hardcoded dates and prefer for this type of data to come from the layer.json. If the layer.json gives us date to use, prism should trust it and not overrule it. If there are issues with the date provided, we fix the layer.json. I'd prefer to not have 2021-07-31 in here at all and only kept it to reduce risk of regressions.

But take this all with a grain of salt, definitely still learning my way around this project 😅

PR #1259, opened 2024-05-28
Dynamic error cdi; COUNTRY=jordan

lowriech:
@wadhwamatic check it out, pretty simple updates to parameterize a local hip service and to utilize a detail error message


README.md
```diff
@@ -343,7 +343,7 @@ cd frontend
 yarn clean
 yarn install
 yarn setup:common
-REACT_APP_COUNTRY=cambodia yarn start
+REACT_APP_COUNTRY=jordan yarn start
```
ericboucher, line 346: 👣

wadhwamatic:
@lowriech - the date error message correctly appears when the latest date (11-May) is requested. But subsequent requests are not rendering when data is available even though the response is the expected geojson object. It would also be good to test this and https://github.com/WFP-VAM/prism-app/pull/1253 together.

lowriech:
@wadhwamatic I'm not seeing the error where "subsequent requests are not rendering when data is available even though the response is the expected geojson object".  Can you clarify?  

Merged in range update PR to this PR.

wadhwamatic:
@lowriech - what I meant is:
1. If you request CDI on a date that doesn't have data, you see the error message you created
2. But if you make a second request one another date which does have CDI data, the geojson response is not rendered on the map even though I see in the dev console that the object was returned
3. It seems that the first error prevents a subsequent valid request from rendering

lowriech:
Can you send the environment you're using to look at this?  That's not what I see, I have updates as usual after the error.  @wadhwamatic

wadhwamatic:
@lowriech - I assigned two issues to Paolo to 1) [modify the error message text](https://github.com/WFP-VAM/hip-analysis/issues/92) and 2) to [create an endpoint for available dates](https://github.com/WFP-VAM/hip-service/issues/12) before a CDI layer is loaded. 

What we have so far works, but is less than ideal since a user has no idea what dates are going to be valid. I'm wondering for now if we should just hack this a bit and assume the latest available CDI will be today - 1 month. 

@ericboucher, what do you think about this approach given that we don't have a way to know available dates for CDI as of now. If this sounds good, maybe @gislawill can take on implementation.

frontend/src/utils/fetch-with-timeout.ts
```diff
@@ -29,10 +29,16 @@ export const fetchWithTimeout = async (
...
+      await res.json().then(r => {
+        // eslint-disable-next-line no-console
+        console.log(r);
+        throw new HTTPError(
+          r.detail ??
+            fetchErrorMessage ??
+            `Something went wrong requesting at ${resource}`,
+          res.status,
+        );
+      });
```
gislawill, line 41: Nice, I like that we're now logging the response on errors. I think this response would be useful to include in the HTTPError class. If we add a errorBody property to HTTPError and pass in r as the third parameter when initializing HTTPError here, our error log in the the catch statement below will show the response body log in the same message as the other details of the error. I think that'll be easier to parse when debugging in the future

ericboucher:
@wadhwamatic can you share a case where the data is not available for CDI?

README.md
```diff
@@ -343,7 +343,7 @@ cd frontend
 yarn clean
 yarn install
 yarn setup:common
-REACT_APP_COUNTRY=cambodia yarn start
+REACT_APP_COUNTRY=jordan REACT_APP_USE_LOCAL_HIP_SERVICE=http://localhost:3001/q_multi_geojson docker compose up
```
ericboucher, line 346: 👣

wadhwamatic:
> @wadhwamatic can you share a case where the data is not available for CDI?

@ericboucher - Jordan CDI on 1-June-2024 returns a date error. https://prism-1259.surge.sh/?hazardLayerIds=cdi_v1&date=2024-06-01

Screenshot 2024-06-19 at 08 14 27

gislawill:
@wadhwamatic, @lowriech, @ericboucher — I've made an attempt to improve this and would appreciate your thoughts. The latest available date for analysis _currently_ is May 1st (exactly two months ago). To hack this before WFP-VAM/hip-service#12, I added an "expectedDataLagDays" option to the composite layer that shifts the default date. 

**Currently,** the default date is the latest available date that that's returned from the server as SPI's latest available date.
**This update,** adds an expected lag of 55 days so the latest available date is the latest date returned from the server as _SPI's_ latest available date **and** is more then 55 days in the past
**After WFP-VAM/hip-service#12,** this date should come from the backend and we can remove this logic

frontend/src/utils/useDefaultDate.ts
```diff
@@ -25,10 +28,23 @@ export function useDefaultDate(layerId: LayerKey): number | undefined {
   const { updateHistory } = useUrlHistory();
 
   // TODO - use getPossibleDatesForLayer
+  // Update on 2024-07-01: getPossibleDatesForLayer would give us the first possible layer date, we want the _last_ possible date.
```
gislawill, line 31: Note here, I don't think this to do is still relevant. Either that or we should refactor how composite layers are handled in getPossibleDatesForLayer

wadhwamatic:
@gislawill - thanks for diving into this. The logic you described below all sounds good to me for our temporary solution until hip-service handles date availability.

> **Currently,** the default date is the latest available date that that's returned from the server as SPI's latest available date.
> **This update,** adds an expected lag of 55 days so the latest available date is the latest date returned from the server as _SPI's_ latest available date **and** is more then 55 days in the past
> **After [WFP-VAM/hip-service#12](https://github.com/WFP-VAM/hip-service/issues/12),** this date should come from the backend and we can remove this logic

However when I tested, I see the same result as before, CDI request still uses dates that do not exist. I tried modifying the `expected_data_lag_days` parameter to use a longer lag period but still see the same result. 

Date error message in PRISM:
Screenshot 2024-07-01 at 21 29 34

Request payload:
Screenshot 2024-07-01 at 21 37 23

gislawill:
@wadhwamatic sure thing, I added that update here: https://github.com/WFP-VAM/prism-app/pull/1295

I separated it from this PR because I'm not confident the improvement in user experience is worth the risk of bugs + complications to maintainance we'd be introducing. If https://github.com/WFP-VAM/hip-service/issues/12 is coming soon and this UX improvement, I believe that solution (in #1295) is functional and addresses the issue. It's just pretty unintuitive code so I wouldn't want it in the codebase for now

gislawill:
Heads up, after discussing with Amit, I pulled all of my commits from this branch to https://github.com/WFP-VAM/prism-app/pull/1295

That allows it to be used for demos without adding this date logic complication to the repo

gislawill:
This should be ready for review again @wadhwamatic and @ericboucher

ericboucher:
ok to merge into the other PR but I would prefer to remove the `REACT_APP_USE_LOCAL_HIP_SERVICE` logic before merging to main

PR #1268, opened 2024-06-12
Load all translations in QA mode

ericboucher:
### Description

When creating the QA build, we will now load all shared translations to make testing easier.

Screenshot 2024-06-12 at 6 27 17 PM

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with


gislawill:
Awesome, I'd find this helpful. Thanks for putting this together @ericboucher.

I don't have a ton of context (so take this with a grain of salt) but I question if `REACT_APP_TESTING` is the right environment variable to use for this trigger. It looks like it always changes the behavior of the `authRequired` flag (line 154 of frontend/src/config/index.ts) which would make the test build a poorer representation of actual app behavior. That's intention for jest tests but I'm not sure that we'd want that for QA.

I could see new variable like `REACT_APP_QA_MODE` that's used instead of `REACT_APP_TESTING` as more appropriate here

gislawill (approved):
Thank you!

Issue #1233, opened 2024-05-02
[Feature Request]: Support markdown in layer legend text


wadhwamatic:
### Provide a clear and concise description of what you want to happen.

I want to be able to add links and format text in the legend using markdown syntax. For example, I'd like to be able to create a link using this text:

`"legend_text": "ACLED Conflict analysis 2023 produced by IFPRI. For more details, see [this publication by IFPRI](https://hdl.handle.net/10568/139670)"
`

Which should result in:
ACLED Conflict analysis 2023 produced by IFPRI. For more details, see [this publication by IFPRI](https://hdl.handle.net/10568/139670)

PR #1279, opened 2024-06-18
Adding Markdown support to Legend text

gislawill:
### Description

This addresses feature request #1233 by adding markdown support for the Legend text in the LegendItem component.

I chose to use the [react-markdown library](https://github.com/remarkjs/react-markdown?tab=readme-ov-file#components) due to it's widespread usage and simple API. There are many more capabilities and customizations we can choose to adopt in the future.

This PR adds support for italics, bold, and links. There are many more elements we could add support for (i.e. an underline without link, strikethrough, headers, tables, lists, etc) — we'll just need to ensure the styling works for our needs.

Note: we're not using the latest version of react-markdown (v9) because it requires an updated version of react (v18 compared to our v16). I don't think we're missing out on much in v9 tbh.

~Note 2: by default, links generated by markdown open in the current tab. We can override to open in a new tab if preferred (I do think that's what users would expect so I plan to add this).~  Updated ✅

## How to test the feature:
@wadhwamatic are there any immediate use-cases for this markdown? If so, I can update the json now so we can use real examples for QA.

## Checklist - did you ...

Test your changes with

Example usage:
Screenshot 2024-06-18 at 9 36 12 AM

Generated html:
Screenshot 2024-06-18 at 9 41 18 AM


frontend/src/setupTests.tsx
```diff
@@ -42,6 +43,11 @@ jest.mock('@react-pdf/renderer', () => ({
   Font: { register: () => {} },
 }));
 
+// https://github.com/remarkjs/react-markdown/issues/635
+jest.mock('react-markdown', () => (props: { children: React.ReactNode }) => {
+  return <>{props.children}</>;
+});
```
gislawill, line 49: This is unfortunately, I would have rather rendered this properly in our jest tests. However, the alternative approach seemed more frustrating to maintain than the value it would provide. 

See: https://github.com/remarkjs/react-markdown/issues/635

Example alternative: 
Screenshot 2024-06-18 at 10 00 54 AM

ericboucher:
@gislawill can you add a temp example in one oft the mozambique legends for this PR?

gislawill:
> @gislawill can you add a temp example in one oft the mozambique legends for this PR?

Can do! Just added an example under INAM aggregate rainfall
Screenshot 2024-06-21 at 8 25 31 AM

ericboucher (approved):
Looks great, thanks!

frontend/src/components/MapView/Legends/LegendItem/index.tsx
```diff
@@ -208,10 +209,29 @@ const LegendItem = memo(
...
+                p: ({ children: pChildren }: { children: React.ReactNode }) => (
+                  <Typography
+                    variant="h5"
+                    className={classes.legendTextMarkdown}
+                  >
+                    {pChildren}
+                  </Typography>
+                ),
+              }}
+              allowedElements={['p', 'h5', 'strong', 'em', 'a']}
```
ericboucher, line 225: Is there a reason to limit elements here? If so, it might be useful to reference allowed elements somewhere else. In the readme maybe or next to the the legendText type?

gislawill: Many of the theoretically allowable elements would not have acceptable formatting here (thinking of examples like tables, lists, code, etc). Rather than try to build reasonably acceptable formatting for all of these elements, I decided to limit to the elements I believe this feature request is intended for. 

Good idea, I'll add a note to the readme what exactly is supported

gislawill:
@wadhwamatic, added the following links. The CHIRPS update was applied in every legend text that had the hyperlink
Screenshot 2024-06-25 at 9 39 06 PM
Screenshot 2024-06-25 at 9 39 43 PM

PR #1297, opened 2024-07-03
Use hazard layer legend for relevant analysis layer if found; COUNTRY=rbd 

### LINKED ISSUE #1173: [Feature Request]: Use thresholds in layer legend for analysis results

When an analysis is run, PRISM automatically categorizes the results as Very Low, Low, Medium, High or Very High and visualizes them in shades of red. However, we already have categories and a color scheme defined in the hazard layer's configuration that are not being used here.

This request is to use the configuration of the hazard layer that was analyzed for categorization and visualization. For example, if a user runs an analysis of 10-day rainfall estimate (`rainfall_dekad`), instead of assigning each admin area Very Low to Very High, we can use the thresholds in the legend. In this example, `rainfall_dekad` uses the legend `rfh_16_0_300` in `shared/legends.json` (see below) which can be applied to the statistic that was analyzed.

```
  "rfh_16_0_300": [
    {"value": 0, "label": "0 mm", "color": "#fafafa"},
    {"value": 0.01, "label": "1-2 mm", "color": "#fffadf"},
    {"value": 2, "label": "2-5 mm", "color": "#d3f9d0"},
    {"value": 5, "label": "5-10 mm", "color": "#a9e4a3"},
    {"value": 10, "label": "10-20 mm", "color": "#7cc594"},
    {"value": 20, "label": "20-30 mm", "color": "#5eab91"},
    {"value": 30, "label": "30-40 mm", "color": "#9fffe8"},
    {"value": 40, "label": "40-50 mm", "color": "#90e0ef"},
    {"value": 50, "label": "50-60 mm", "color": "#00b1de"},
    {"value": 60, "label": "60-80 mm", "color": "#0083f3"},
    {"value": 80, "label": "80-100 mm", "color": "#0052cd"},
    {"value": 100, "label": "100-120 mm", "color": "#0000c8"},
    {"value": 120, "label": "120-150 mm", "color": "#6003b8"},
    {"value": 150, "label": "150-200 mm", "color": "#a002fa"},
    {"value": 200, "label": "200-300 mm", "color": "#fa78fa"},
    {"value": 300, "label": ">= 300 mm", "color": "#ffc4ee"}
  ],
```

There are three additional issues to consider with this change:
1. If the hazard layer was already loaded, we might want to hide the raster (png tiles) version of the layer and just show the analysis results
2. In the future, we might want to give user's the option to categorize results based on x buckets. This needs some more thinking and validation of the need though
3. As @ericboucher mentioned in the comment in this issue, this new approach won't apply when we run an area / population exposure analysis.

gislawill:
### Description
This fixes #1173. A surprisingly and delightfully simple update.

| Before | After | 
| - | - |
| Screenshot 2024-07-03 at 12 35 08 PM | Screenshot 2024-07-03 at 12 34 31 PM |

## How to test the feature:
- [ ] Run an analysis (such as RBD's 10-day rainfall estimate)
- [ ] Confirm legend color matches the analysis' associated hazard layer legen

Test your changes with


wadhwamatic:
Nice, glad to hear this was a simple change @gislawill. But, there's a case not covered unfortunately. In some WMS layers, there's a scale and offset factor that is used to convert units from the source to something meaningful for users in PRISM.

An example is lst_daytime. The wcsConfig configuration is used to convert the source (Kelvin) to Celsius. We need to also utilize that factor when we apply the legends in the analysis results layer.

The strange thing is that scaling used to work and the analysis results correctly showed values (in Celsius in this example). I'm not sure when this broke. @ericboucher - any thoughts?

Here's a screenshot of analysis results for LST along with the underlying layer. https://prism-1297.surge.sh/?analysisHazardLayerId=lst_daytime&analysisBaselineLayerId=admin_boundaries&analysisDate=2024-06-21&analysisStatistic=mean&hazardLayerIds=lst_daytime
Screenshot 2024-07-03 at 13 34 10

wadhwamatic:
@ericboucher and @gislawill - I created a new issue for the scale factor problem I mentioned in my earlier comment. Eric, I'm not sure when this issue surfaced or if it somehow has always been there. That seems unlikely, but is not impossible!

https://github.com/WFP-VAM/prism-app/issues/1298

This PR seems completely unrelated and is likely GTG, but I'll let Eric chime in on that.

gislawill:
> But, there's a case not covered unfortunately. In some WMS layers, there's a scale and offset factor that is used to convert units from the source to something meaningful for users in PRISM.

Oh nice catch, I'll look into this @wadhwamatic

wadhwamatic:
> > But, there's a case not covered unfortunately. In some WMS layers, there's a scale and offset factor that is used to convert units from the source to something meaningful for users in PRISM.
> 
> Oh nice catch, I'll look into this @wadhwamatic

Thanks @gislawill. I think this is like on the PRISM API side but might be mistaken. Might be worth having @ericboucher give some guidance before spending too much time digging into this

wadhwamatic:
As far as I can tell, this PR is GTG. The analysis result issue that was noted is unrelated and captured in https://github.com/WFP-VAM/prism-app/issues/1298. 

From what I see, this works as expected when a layer is scaled. See SPI 1-month as an example:
Screenshot 2024-07-10 at 13 25 59

Screenshot 2024-07-10 at 13 24 35

Suggesting we go ahead and merge this. Let me know what you think @gislawill and @ericboucher

wadhwamatic (approved):

gislawill:
Sounds good, thanks @wadhwamatic. It's good to go as far as I'm concerned. I'll wait to press the green button until @ericboucher has a chance to weigh in

frontend/src/context/analysisResultStateSlice.ts
```diff
@@ -737,7 +737,8 @@ export const requestAndStoreAnalysis = createAsyncThunk<
   );
 
   // Create a legend based on statistic data to be used for admin level analsysis.
-  const legend = createLegendFromFeatureArray(features, statistic);
+  const legend =
+    hazardLayer.legend ?? createLegendFromFeatureArray(features, statistic);
```
ericboucher, line 741: In theory this would also depend on the type of `statistic` being run. This makes sense for `min, max, mean, median` but not so much for `sum, std, count`

Maybe add a small note for later? Or a switch on statistics? wdyt?

ericboucher (approved):

Issue #1266, opened 2024-06-11
Decide how to handle overlapping dates


ericboucher:
### Provide a clear and concise description of what you want to happen.

With the new date range vizualiation implemented in #1263, we are facing a few UX issues on how to handle overlapping date ranges for the same layer

wadhwamatic:
@ericboucher - can you remind me, when do we have overlapping dates for the same layer?

wadhwamatic:
@ericboucher - here's how I think this should work for WFP HDC layers:

When a user selects a date, we can safely assume they have selected the **_last dekad in the date range_**. This means that if the HDC layer is for 1-month, the date selected is the dekad 3, which is closest available date noted in the HDC layer's available dates. Then, in this case, we visualize the layer as active for the two previous dekads, and the last one.

A concrete example:
1. Use PR https://github.com/WFP-VAM/prism-app/pull/1271
2. Load Rainfall > Global rainfall products > Rainfall aggregate > 1-month
3. By default, we see the right behavior - the latest available date is activated and the timeline visualizes dekad 3 and forward as valid, and dekads 1 and 2 backwards as valid. 
4. Choose another date - 2024-April-12
5. This should assume that the closest dekad starting point is the 3rd dekad in the 1-month product, which in this case should be 11-April (the 2nd dekad of the month)
6. We should see 11-20 April as valid as the 3rd dekad of this 1-month product. We should also see 21-31 March (1st dekad) and 1-10 April (2nd dekad)
7. In total, the active layer / date should cover the period starting 21-March and until 20-April

The same logic applies regardless of the length of the product. I.e., if it's a 3-month product (which is a total of 9 dekads long), we still assume the user's date selection matches the 9th dekad. 

Hope this is clear!

gislawill:
@ericboucher and @wadhwamatic, I believe this issue can be closed now that #1301 has been merged

Issue #1181, opened 2024-04-03
[Feature Request]: Refine CDI implementation


wadhwamatic:
### Provide a clear and concise description of what you want to happen.

The implementation of CDI needs some refinement both within PRISM and on the hip service side.

1. We should consider how we can improve performance. Can we prefetch the latest CDI?
2. If an input layer for the CDI is not published via HDC's WMS, we can't display it within the layer drop down (see screenshot below). Rather than not show it, we could simply display the layer as a text field without a layer toggle. This would require a `title` in the `input_layers` configuration in PRISM
1. I think we should have two types of CDI layers in PRISM – `monthly` and `seasonal` (TBC before implementing)
a. If seasonal, the configuration in layers.json should require the start of each season and end of each season as parameters. For example, `s1_start` = 1-January; `s1_end` = 31-March; `s2_start` = 1-April; `s2_end` = 30-June; etc.
b. The user’s timeline selection should determine which dates are used in the CDI request. For example, if the user selects CDI for today, 2-April, the seasonal request should include the latest available season, which in this example would be season 1 (Jan-March). 
c. If monthly, we simply have CDI refer to the full calendar month. In the case of a user selecting 2-April, we should show CDI for the month March
3. There’s no map interaction with the CDI layer. Ideally, a user should be able to see the CDI value in the tooltip when they click on a pixel.
4. From what I understand, the aggregation options within each input layer should be implicitly derived on the hip service side instead of expected as parameters in the PRISM configuration. @paololucchino - can you confirm if this is correct and if we can adjust on the hip service?
5. Future to do – I think we should also add an option to aggregate CDI pixel results and get the mean value per admin area. This will require zonal stats to be implemented within the hip service, and for the PRISM frontend to be modified to send a boundary file in the request and to map the response accordingly.

As mentioned in the 2nd bullet point above, the PRISM interface only shows SPI as an `input_layer` because other input layers are not available through HDC
<img width="495" alt="Screenshot 2024-04-02 at 16 59 19" src="https://github.com/WFP-VAM/prism-app/assets/3343536/785aa2de-4fae-4502-bfc9-c40835b35ab1">

lowriech:
Some quick thoughts, mostly driven on my not contributing to PRISM previously:
[...]

PR #1301, opened 2024-07-09
Adding support for seasonal cdi; COUNTRY=jordan

gislawill:
### Description
This addresses [this request in #1253](https://github.com/WFP-VAM/prism-app/pull/1253#issuecomment-2122649243) and points 2 and 3 of #1181.

> 2. If an input layer for the CDI is not published via HDC's WMS, we can't display it within the layer drop down (see screenshot below). Rather than not show it, we could simply display the layer as a text field without a layer toggle. This would require a title in the input_layers configuration in PRISM
> 3. I think we should have two types of CDI layers in PRISM – monthly and seasonal (TBC before implementing)
a. If seasonal, the configuration in layers.json should require the start of each season and end of each season as parameters. For example, s1_start = 1-January; s1_end = 31-March; s2_start = 1-April; s2_end = 30-June; etc.
b. The user’s timeline selection should determine which dates are used in the CDI request. For example, if the user selects CDI for today, 2-April, the seasonal request should include the latest available season, which in this example would be season 1 (Jan-March).

I believe this also fixes issue #1266 

Note: this PR considers seasons as quarters. A PR will follow this that adds support for flexible seasons.

Open questions:
- [x] ~@wadhwamatic is this the UI you had in mind?~ 
  - Answer: incorporate `validity` layer config for better seasonal selection
- [x] ~@lowriech, any idea why the these `q_multi_geojson` requests fail when `began` and `end` are 3 months apart rather than 1 month apart?~ 
  - Answer: End date needs to be the first of the next month, not the last day of the season
- [x] ~I didn't use the s1_start, s2_end - is there a value to sending these instead of the dates that the API is already set to accept?~ 
  - Answer: Yes but we're adding that in a follow-up PR
- [x] ~Should we use SPI 3 months instead of 1 month for seasonal composites?~ 
  - Answer: Yes, 3 month
 
To Do:
- [x] Adjust month validity to be one dekad forward, two back
- [x] Limit Season date availability to first day of season
- [x] Update composite on Mozambique and Cambodia config once finalized

## How to test the feature:
- [ ] Open Combined Drought Index layers in Jordan
- [ ] Use dropdown to toggle monthly vs seasonally

## Checklist - did you ...
Test your changes with

https://www.loom.com/share/e85d16dce7dd4948b4e6a81a0217495e?sid=5d89309c-be1a-4043-9f83-e8629c55a808


frontend/src/components/MapView/LeftPanel/layersPanel/MenuItem/MenuSwitch/SwitchItem/LayerDownloadOptions.tsx
```diff
@@ -254,19 +254,14 @@ function LayerDownloadOptions({
...
-            </>
-          )}
+          layer.baseUrl.includes('api.earthobservation.vam.wfp.org/ows') && [
+            <MenuItem key="download-as-geotiff" onClick={handleDownloadGeoTiff}>
+              {t('Download as GeoTIFF')}
+            </MenuItem>,
+            <MenuItem key="download-style" onClick={handleDownloadQmlStyle}>
+              {t('Download QML Style')}
+            </MenuItem>,
+          ]}
```
gislawill, line 264: This fixed some annoying console errors

frontend/src/components/MapView/LeftPanel/layersPanel/MenuItem/MenuSwitch/SwitchItems.tsx
```diff
@@ -21,12 +22,18 @@ interface SwitchItemsProps extends WithStyles<typeof styles> {
...
+  const { selectedLayers } = useLayers();
   return (
     <>
       {layers.map((layer: LayerType) => {
         const foundNotRenderedLayer = layer.group?.layers.find(layerItem => {
-          return layerItem.id === layer.id && !layerItem.main;
+          return (
+            layerItem.id === layer.id &&
+            !layerItem.main &&
+            !selectedLayers.some(sl => sl.id === layerItem.id)
```
gislawill, line 33: Allows the non-"main" layer to show on initial load if it's the selected layer (relevant for SPI 3 months when 3 months is selected)

gislawill:
I've added "validity" support to the composite layers but now running into issues in the timeline selector when apply more than 1 dekad forward. Seems to be a pre-existing issue but this is the first time (I believe) we're attempting this behavior. I'm digging into and will continue to provide updates.

https://www.loom.com/share/0aacab4149b64b55841b08306dd5e68e?sid=5da3b728-e30e-4e9d-9861-01c82af3496d

wadhwamatic:
@gislawill - yes, that's a known issue. We have a discussion on it here:
https://github.com/WFP-VAM/prism-app/issues/1266

I don't know if @ericboucher has some ideas yet on how to resolve this. Would be worth checking in with him on Monday.

gislawill:
While attempting to add the "validity" support to this feature, I discovered an issue with seasonal date selection:
- The date options in our time selection doesn't match our analysis
  - The time selection allows to select a date every dekad (1st, 11th, 21st days of the month)
    - That is due to using the SPI_1m and SPI_3m layers as the "dateLayer"
  - The CDI analysis always runs from the start of the season to the end of the season
    - This is based on how we developed it in https://github.com/WFP-VAM/prism-app/pull/1253
  - **Impact:** Users can select 12 different dates that perform the same analysis

I believe the solution is finding a way to limit the date selection options to the first day of the season

frontend/src/utils/server-utils.ts
```diff
@@ -395,20 +394,12 @@ export function generateIntermediateDateItemFromValidity(
...
+      return [...acc, ...dateItemsToAdd];
     }, []);
 
   // We sort the defaultDateItems and the dateItemsWithValidity and we order by displayDate to filter the duplicates
   // or the overlapping dates
-  return sortedUniqBy(
-    sortBy(dateItemsWithValidity, 'displayDate'),
-    'displayDate',
-  );
+  return sortBy(dateItemsWithValidity, 'displayDate');
```
gislawill, line 402: Removing these deduping steps does fix #1266. This update essentially increases the number of "available dates" to handle the overlapping periods. Before this change, when a date was found to be within two ranges, it'd be filtered from our list. With this change, that same date would in our list of available dates under all valid ranges.

This does _function_ as we'd desire but it does so by inflating our date list by a factor equaling the dekad in the layer's validity range. For "months", our list of dates to run through increases by 3x. For our seasons, it increases by 9x.

Creating this list doesn't result in any real slowdown of our date initialization. **However**, renders in react are slowed by operations which run through this list (particularly running the `datesAreEqualWithoutTime` function). I'm seeing a slowdown of about half a second in the first few renders on the page.

| Before | After |
| - | - |
| Screenshot 2024-07-13 at 1 06 46 PM | Screenshot 2024-07-13 at 12 59 54 PM |

I'm pursuing two solutions:
- Can we speed up the `datesAreEqualWithoutTime` function by using date-fns so the larger lists aren't an issue?
- Can we base validity time window off of the start and end times of the selected time rather than reading from from the list of available dates? That removes the need to inflate this list

gislawill: I was able to speed up datesAreEqualWithoutTime by 1000x by updating to the function to use math rather than date objects (at least in Chrome browsers).
Screenshot 2024-07-14 at 3 18 26 PM

This brings the performance back up to where things were before. However, I do still wonder if it makes sense to store validity dates in "available dates". I think I'd prefer to keep available dates as our valid dekad options (1st, 11th, 21st) and rely on the start and end dates provide to determine validity at render rather than fill our available dates with multiples of validity dates. I'm going to explore what this refactor would look like

gislawill: Before refactoring validity dates our "available dates", we're already at performance improvements on what existed before. I'm going to hold off on that refactor for now
Screenshot 2024-07-15 at 10 16 57 AM

frontend/src/components/MapView/LeftPanel/layersPanel/RootAccordionItems/index.tsx
```diff
@@ -1,11 +1,16 @@
...
 import useLayers from 'utils/layers-utils';
 import MenuItem from '../MenuItem';
-import { menuList } from '../../utils';
+import { getDynamicMenuList } from '../../utils';
 
 const RootAccordionItems = memo(() => {
   const { adminBoundariesExtent: extent } = useLayers();
+  const { selectedLayers } = useLayers();
+  const menuList = useMemo(
+    () => getDynamicMenuList(selectedLayers),
```
gislawill, line 11: The dynamic menu list allows us to use different layers (monthly vs seasonal) to drive the same menu item with a toggle

frontend/src/config/jordan/layers.json
```diff
@@ -4015,10 +4015,16 @@
     ],
     "legend_text": "Groundwater developed for irrigation (MCM) "
   },
-  "cdi_v1": {
-    "title": "Combined Drought Index (v1)",
+  "cdi_v1_monthly": {
```
gislawill, line 4018: Note: I chose to use different layers for monthly vs seasonal. This was primarily driven by the need for different options in the layers' "available dates" for time selection and querying

frontend/src/utils/date-utils.ts
```diff
@@ -6,13 +6,17 @@ export interface StartEndDate {
...
 ): boolean => {
-  const d1 = new Date(date1).setUTCHours(0, 0, 0, 0);
-  const d2 = new Date(date2).setUTCHours(0, 0, 0, 0);
-  return d1 === d2;
+  const cleanDate1 = date1 instanceof Date ? date1.getTime() : date1;
+  const cleanDate2 = date2 instanceof Date ? date2.getTime() : date2;
+  return (
+    cleanDate1 - (cleanDate1 % millisecondsInADay) ===
+    cleanDate2 - (cleanDate2 % millisecondsInADay)
+  );
```
gislawill, line 19: By avoiding the date manipulation, pretty decent performance boost (200x faster operation)

frontend/src/utils/date-utils.ts
```diff
@@ -172,3 +176,21 @@ export const getFormattedDate = (
...
   new Date(date).getTime();
+
+const SEASON_MAP: [number, number][] = [
+  [0, 2],
+  [3, 5],
+  [6, 8],
+  [9, 11],
+];
+
+export const getSeasonBounds = (date: Date) => {
```
gislawill, line 187: This will be swapped up when we add support for flexible seasons next

frontend/src/utils/server-utils.test.ts
```diff
@@ -91,6 +91,12 @@ describe('Test generateIntermediateDateItemFromValidity', () => {
         startDate: layer.dates[0],
         endDate: new Date('2023-12-11').setHours(12, 0),
       },
+      {
+        displayDate: new Date('2023-12-11').setHours(12, 0),
+        queryDate: layer.dates[0],
+        startDate: layer.dates[0],
+        endDate: new Date('2023-12-11').setHours(12, 0),
+      },
```
gislawill, line 99: These represent the overlapping dates that fixes #1266

wadhwamatic:
@gislawill - sorry, just saw these questions

> * For seasonal ranges, how dynamic do we want/need the ranges to be? They will be configurable through layer.json but should they also be configured by the user?
>   
>   * For example, should we start thinking about building a date range slider of sorts

No, users wouldn't able pick their own ranges on the fly, this is a configuration / setup decision.

> * For the climatology date range, do we want to always send the 19 months from Jan 2020 through July 2021 or something different?
>   
>   * Longer range to better represent climatology of the region?

I am not sure where Jan-2020 - July-2021 came from. The reference period should definitely be longer. I thought we had reference data back to around 2000. This should be part of the individual CDI config IMO otherwise we're creating an arbitrary reference period for all composite layer implementations. But, let's create a separate issue for that and further complicate this PR.

>   * Based the reference range on the date range the user selected?

No, the reference range is moreso driven by the availability of dates in the underlying layers, which someone configuring the CDI will need to know. As of now, that was captured in the hip-analysis package iirc.

wadhwamatic:
@gislawill - another small comment: I fixed an error in the CDI config for Mozambique and tested non-HDC layers. The CDI loads as expected, but the non-HDC layers are not yet shown in the interface. These were expected as text only since they are not configured as PRISM layers. This is another issue we can address separately but mentioning it here as it was noted at the beginning of this PR's discussion.

Screenshot 2024-07-15 at 23 44 03

In this case, `WAPOR_ET0_DEKAD` & `SOIL_MOIST` are non-HDC layers.

```
    "input_layers": [
      {
        "id": "spi_1m",
        "importance": 0.4,
        "key": ["CHIRPS", "R1S_DEKAD"],
        "aggregation": "last_dekad"
      },
      {
        "id": "et0",
        "importance": 0.3,
        "key": ["FAO", "WAPOR_ET0_DEKAD"],
        "aggregation": "last_dekad"
      },
      {
        "id": "rzsm",
        "importance": 0.3,
        "key": ["AGERA5", "SOIL_MOIST"],
        "aggregation": "last_dekad"
      }
    ],
```

frontend/src/components/MapView/DateSelector/TimelineItems/TimelineItem/index.tsx
```diff
@@ -32,7 +32,7 @@ const TimelineItem = memo(
     ): boolean => false;
 
     const isQueryDate = (date: DateItem): boolean =>
-      date.queryDate === date.displayDate;
+      datesAreEqualWithoutTime(date.queryDate, date.displayDate);
```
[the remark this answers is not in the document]
ericboucher: Note - The 12h shift is also used to avoid timezone errors and avoid differences between what would get displayed for two users in opposite timezones. We would need to check that this does not introduce unexpected behaviors.

frontend/src/config/cambodia/layers.json
```diff
@@ -1950,7 +1950,13 @@
   "cdi": {
     "title": "Combined Drought Index",
     "type": "composite",
-    "base_url": "https://to-fill-later",
+    "scale": "monthly",
```
ericboucher, line 1953: The term `scale` feels off to me but not sure what would be better. `period`? or `interval`? Is that a term coming from HIP? Or a standardized notion?

frontend/src/utils/layers-utils.tsx
```diff
@@ -130,7 +130,10 @@ const useLayers = () => {
           }
           if (layer.type === 'composite') {
             // some WMS layer might not have date dimension (i.e. static data)
```
ericboucher, line 132: old 👣

ericboucher:
Great work @gislawill, just a few things to watch out for:

1. timezones and usability in other countries

2. adding validity objects without dedup creates issues with `selectedLayerDatesDupCount` and layer date validators. So we can end up with weird messages like this.

Screenshot 2024-07-17 at 12 42 53 AM

The logic in these files (`frontend/src/utils/layers-utils.tsx` in particular)  can definitely be improved 😅

frontend/src/config/cambodia/layers.json
gislawill: Good call, I like period better. I'll update.

This is all internal to prism and it not used directly by hip-analysis or hip-service

frontend/src/components/MapView/DateSelector/index.tsx
```diff
@@ -97,6 +103,98 @@ const DateSelector = memo(() => {
     [availableDates],
   );
 
+  const AAAvailableDates = useSelector(AAAvailableDatesSelector);
```
gislawill, line 106: Moving these up from TimelineItems to make use of the processed layers that are created

frontend/src/components/MapView/DateSelector/index.tsx
```diff
@@ -278,16 +376,31 @@ const DateSelector = memo(() => {
...
+        .map(dateItem => dateItem.displayDate),
+    );
+
+    // Find the dates that are queriable for all layers
+    const validDates = queriableDates.reduce((acc, currentArray) =>
+      acc.filter(date => currentArray.includes(date)),
+    );
+
+    const selectedIndex = findDateIndex(validDates, dateRange[index].value);
+
```
gislawill, line 394: This feels like a lot of array looping and gave me some pause. I did measure it though for two HDC layer (with differing validity) and found it takes less than a millisecond on my computer. Not cause for concern imo
Screenshot 2024-07-17 at 10 36 13 AM

gislawill:
@wadhwamatic and @ericboucher, I believe I've addressed the immediate issue we identified, they should be deployed to surge shortly. I'm doing a more thorough round of QA now and will set this PR to ready to review once my QA is complete

wadhwamatic:
@gislawill - I see an issue when I load a the daily forecast layer followed by CDI:
Screenshot 2024-07-17 at 12 26 39

I think this is likely due to the `start_date` attribute associated with forecast products. In this case, HDC has date availability for `daily_rainfall_forecast` but we're limiting what is accessible based on the `start_date`. The logic around that is probably not included in the date intersection logic.

gislawill:
> @gislawill - I see an issue when I load a the daily forecast layer followed by CDI: Screenshot 2024-07-17 at 12 26 39
> 
> I think this is likely due to the `start_date` attribute associated with forecast products. In this case, HDC has date availability for `daily_rainfall_forecast` but we're limiting what is accessible based on the `start_date`. The logic around that is probably not included in the date intersection logic.

@wadhwamatic shoot, yea I'm seeing that as well. I'm seeing similar areas on the master branch for these layers as well (but there, the time selection actually drops from the page.

In this case where we do not have any overlapping dates, do you think it's best to either:
- Allow any date to be selected and jump to the nearest date with data for any layer
- Prevent these two layers from being selected at the same time

I'm leaning towards the latter personally

(the situation on the master branch)
Screenshot 2024-07-17 at 3 10 42 PM

wadhwamatic:
@gislawill - yes, I agree we should use the second option:
> Prevent these two layers from being selected at the same time

This logic already exists in PRISM (not sure where though!) where we check to see if overlapping dates exist when a second (or third) layer is activated. If not dates align, the user is given a message to either deselect the current layer or try another date.

gislawill:
> This logic already exists in PRISM (not sure where though!) where we check to see if overlapping dates exist when a second (or third) layer is activated. If not dates align, the user is given a message to either deselect the current layer or try another date.

@wadhwamatic Found it, ok I believe I've mirrored the existing behavior and have a pretty decent UX. Here's a demo (no audio, just display) https://www.loom.com/share/1f0088ffbf604a4b9fda922da91f5421?sid=d52900fe-009a-4408-a945-f0ed1e59134d

frontend/src/components/MapView/DateSelector/TimelineItems/TimelineItem/index.tsx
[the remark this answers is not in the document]
gislawill: Ran this under various timezone on my computer (london, tokyo, manila) and it all looks good 👍 Definitely a good thing check, thanks for the reminder

frontend/src/components/MapView/DateSelector/index.tsx
gislawill, line 1: Reorganizing the callbacks in here to make better use of the no layer matching date alert

gislawill:
Ok @ericboucher and @wadhwamatic, this got quite a big bigger than I intended but I've tested it thoroughly and it's ready for your review. Thanks for your patience with this one.

wadhwamatic:
Thanks @gislawill , but unfortunately I still see some issues.

1. For Jordan, if I select Rolling daily rainfall forecast, followed by CDI I correctly see the message that the date is not valid for all layers, but CDI continues to try and load which then gives the hip service error while the CDI still shows up in the timeline. In this case, CDI shouldn't load
2. For RBD, if I select 10-day rainfall aggregate, followed by Cadre Harmonise > Phase classification > Overall phase classification (layer id `ch_phase`), I incorrectly see the message that dates don't intersect (they do) but the layer loads. Thereafter the timeline date selection is inconsistent. See video below


https://github.com/user-attachments/assets/6cd7a5c0-dfb6-4809-808a-4ae9284405ab

gislawill:
Thanks @wadhwamatic, I'm digging into these issues now as well as an issue @ericboucher pointed out involving the date selection jumping unnecessarily to "queryable" dates rather than remaining on valid dates

gislawill:
Third time's the charm? I've addressed the items you both identified, @wadhwamatic and @ericboucher. Thanks again!

wadhwamatic:
Sorry @gislawill , I'm still seeing the same issue mentioned for Jordan: if I select Rolling daily rainfall forecast, followed by CDI I correctly see the message that the date is not valid for all layers, but CDI continues to try and load. In this case, CDI shouldn't load as it doesn't intersect with the forecast layer

gislawill:
> Sorry @gislawill , I'm still seeing the same issue mentioned for Jordan: if I select Rolling daily rainfall forecast, followed by CDI I correctly see the message that the date is not valid for all layers, but CDI continues to try and load. In this case, CDI shouldn't load as it doesn't intersect with the forecast layer

@wadhwamatic, I might be mis-understanding what you mean by "shouldn't load". I'm seeing that the CDI layer is selected but the error message shows and no request to hip-service is made (see video). Are you seeing something different or expecting a different behavior?

https://www.loom.com/share/0fefb68e757648b88dec04292f81a27e?sid=30673bbc-a9f6-4190-b0e3-fe4fd50f67a2

wadhwamatic:
@gislawill - yes, I'm seeing the same thing. What I mean is that CDI shouldn't show up in the legend and it shouldn't show toggled on as it has no intersection with the forecast layer. 

I thought that is the standard behavior we have implemented: if a second layer is activated but it has no date intersection with the previously selected layer, the user sees a message that no intersection exists, and the layer is not shown in the legend and not toggled on. I might be mistaken though and it's possible that this is a new edge case introduced when we added forecast layers.

The thing to address IMO is to not show CDI in the legend, and to not show it toggled on in this case

frontend/src/utils/layers-utils.tsx
```diff
@@ -389,13 +400,10 @@ const useLayers = () => {
 
   // let users know if the layers selected are not possible to view together.
   useEffect(() => {
-    // TODO: Why is this the case here? maybe we should remove it;
     if (
-      // eslint-disable-next-line no-constant-condition
       selectedLayerDates.length !== 0 ||
       selectedLayersWithDateSupport.length === 0 ||
-      !selectedDate ||
-      1
```
gislawill, line 398: @ericboucher and @echaidemenos, I see that this logic was switched off in #1142 but we definitely do want it running now to handle situations that occur when two layers are selected with no overlapping dates (as is more common now with this composite layers). 

I've added the logic back and pulled up the Anticipatory Action module for Mozambique. Everything looks good to me but I'd love your thoughts. Anything I missed?

https://github.com/WFP-VAM/prism-app/pull/1142/files#diff-655c8639bf9b7a6f6c5aaa1270ad39e55d5500770724783a1eb36a94a8a174ae

gislawill:
A lot of updates have gone in to make sure the date selection and messaging work as expected across different scenarios. A quick list of what I've QA'ed with the latest round:
- [ ] One layer with validity and one without and overlapping dates (Jordan's CDI + SPI)
  - [ ] Selecting a shared queryable date works
  - [ ] Selecting a date that's valid in one but not the other jumps to nearest that's valid for both and displays message
  - [ ] Selecting a date that's valid for neither jumps to nearest that's valid for both and displays message
- [ ] Two layers with overlapping validity (RBD's Rolling daily rainfall forecast + Cadre Harmonize Phase Classification)
  - [ ] Selecting a shared queryable date works
  - [ ] Selecting a date that's valid in one but not the other jumps to nearest that's valid for both and displays message
  - [ ] Selecting a date that's valid for neither jumps to nearest that's valid for both and displays message
- [ ] Two layers with no overlapping selectable dates (Jordan's CDI + Rolling daily rainfall forecast) unselects the prior layer and loads the nearest date of the remaining layer
- [ ] Anticipatory Analysis layers (Mozambique)
- [ ] Dragging the timeline ticks (thanks @ericboucher!)
- [ ] Only actually selectable dates are selectable in the DatePicker (Jordan's CDI + SPI)

ericboucher:
@gislawill This is working pretty well overall, great job.
I still noticed a small issue from time to time where the date picker would jump one day to the right from the date I clicked despite the date clicked being available.

Eg. for RBD http://localhost:3000/?hazardLayerIds=rainfall_dekad&date=2024-05-21&baselineLayerId=ch_phase
Clicking on May 20th always puts me back on May 21st.


https://github.com/user-attachments/assets/163545af-8856-4d0c-beeb-877108b679d9

gislawill:
> This is working pretty well overall, great job. I still noticed a small issue from time to time where the date picker would jump one day to the right from the date I clicked despite the date clicked being available.

Nice catch @ericboucher! I've just pushed up a commit with a fix (replacing a === with `datesAreEqualiWithoutTime` in a function called findDateIndex that called in the quick handler

frontend/src/utils/layers-utils.tsx
gislawill: @ericboucher, friendly nudge here to make sure you see it when you next review (easy to miss in a PR this unruly)

wadhwamatic:
Great work @gislawill! The potential edge cases I've been testing all work well now with one exception for Cambodia.

If load these three layers and then move to another date on the timeline, I get the error message below. Please connect with @ericboucher to get credentials to test the Cambodia flood report layer
Layers:
- EWS 1294 river level data
- Rainfall aggregate 10-day
- Flood report

Error:
Cannot convert undefined or null to object
Screenshot 2024-07-20 at 13 20 20

One minor request: can we change the text that appears when we can't load the second layer?
Screenshot 2024-07-20 at 13 15 18

Remove the word 'previous':
"No dates overlap with the selected layers. Removing layer: Monthly Combined Drought Index (v1)."

ericboucher:
@wadhwamatic I update the error text. However, I can't reproduce the error you're describing for Cambodia.
This seemed to be working well for me and I was able to navigate between dates without any issue. Or maybe a timezone issue with date conversion coming from this layer?

Screenshot 2024-07-22 at 11 18 38 AM

gislawill:
Thanks @ericboucher for updating the copy and for the credentials to view the reports layers 🙏

@wadhwamatic, similar to Eric, I could not reproduce that error. I also ran the page with my computer in Wellington time (New Zealand), Beijing time, London time, and Pacific Daylight time. None of them could reproduced that error when viewing or navigating around those three layers. 

Could I ask you to try again and, if you experience that error again, could you find the details of the error in the JavaScript console (Command+Option+J on a Mac) and send a screenshot? That would help us isolate where the issue is coming from and help us figure out why we're not seeing it.

ericboucher:
@gislawill looking great, one thing I forgot is to test for Anticipatory Action in mozambique.
AAs are a bit particular and we are allowed to select a date that only has 1 out of 2 "window data" available.

We should be able to select May 18th here for example:
Screenshot 2024-07-23 at 1 22 31 PM

gislawill:
@ericboucher I believe the fix for the AA layers is ready in [7f0d825](https://github.com/WFP-VAM/prism-app/pull/1301/commits/7f0d8253985205c8fd3b5f3a71e27ff055f8cd54). I believe this matches the date selection logic/availability that I'm seeing on the master branch. 

Thanks again for your thorough qa on this one. I'm really happy we caught all these issues now before this was merged

ericboucher:
I added a small fix for AA here - https://github.com/WFP-VAM/prism-app/pull/1301/commits/e82ef7af40b3be106ca91e99ce2a08119a6743d1 and a note. It was already off on `master` and displaying august data as "visible" for W2 when May was selected.



Not sure if we should have update that behavior for other layers as well, wdyt @gislawill ? The defaulting to latest date logic is probably handled somewhere else already no? The risk to keep this default here is to display "visible dates" that are out of sync with what users are actually seeing.

ericboucher (approved):

gislawill:
> **I'm realizing now that the logic I mentioned above is no longer functioning as expected — must have regressed in the past few rounds of updates. I'm working on a fix.**

This is fixed in ce6877f

frontend/src/components/MapView/LeftPanel/utils.ts
```diff
@@ -19,20 +20,26 @@ type LayersCategoriesType = LayersCategoryType[];
...
-        const layer = LayerDefinitions[mainLayer?.id as LayerKey];
-        // eslint-disable-next-line fp/no-mutation
-        layer.group = group;
+        const visibleLayer =
+          group.layers.find(gl =>
+            selectedLayers?.some(sl => sl.id === gl.id),
+          ) || group.layers.find(l => l.main);
+
+        const staticLayer = LayerDefinitions[visibleLayer?.id as LayerKey];
+        const layer = { ...staticLayer, group };
```
ericboucher, line 42: @gislawill I think that's where the issue comes from for grouped layers. We end up with multiple layers in the timeline when we only want the `main` one.

Can you explain this change? Could we just do `const layer = staticLayer`

gislawill: This change supported allow the same left panel menu item to support two different layers (in this case, monthly vs seasonal). I see the issue you're referring to but I don't think it stems for this change — the issue is in the timeline view and this change is specifically for the menu items. 

I'm going to investigate what updates were made in the timeline view involving grouped layers.

gislawill: @ericboucher you were right that this function was the source of the error. By using `const layer = { ...staticLayer, group };` instead of `layer.group = group`, I was avoiding the eslint error (fp/no-mutation) without realizing that modifying the layer definition in our global constants was exactly what we were trying to do. 

Because I updated this formatLayersCategories function to run during renders (getDynamicMenuList) rather than just at start up, a new error occurred when writing to our layer definition (they're frozen sometime after initialization but before renders it seems). I put in place a fix for that so everything functions as expected and added comments explaining the fix.

I should note, I don't think that writing to our global layer constants (LayerDefinitions) from within a menu formatter is a great pattern to follow. Rather, we should write the group property when creating the LayerDefinitions. Because this both a pre-existing issue and this PR is already way too large, I'd prefer to that fix that separately. I'd be happy to create an issue if you agree.

ericboucher: Agreed

ericboucher (approved):

PR #1308, opened 2024-07-10
Test translations

ericboucher:
### Description

Add a test script to look for missing translation in the code as well as layers and legends.

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with


ericboucher:
@wadhwamatic @gislawill this adds a new test scripts that pulls all the layer info and tries to see if it was translated:
Screenshot 2024-07-10 at 7 26 56 PM

frontend/src/config/config.test.ts
```diff
@@ -0,0 +1,117 @@
...
+      // const legendLabels = jp.query(layers, '$..legend[*].label');
+      // eslint-disable-next-line fp/no-mutation
+      itemsToTranslate = [
+        ...Array.from(translationKeys),
+        ...categoryKeys,
+        ...categoryGroupTitles,
+        ...layerTitles,
+        ...layerLegendTexts,
+        ...chartLegendLabels,
+      ];
```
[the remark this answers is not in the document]
ericboucher: In particular I did not add legend labels because most of them are `20mm`, `23C`, ... and so on but let me know if you would like to include them and exclude simple labels with some patterns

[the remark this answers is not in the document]
wadhwamatic: @ericboucher I'd rather add the legend labels as it's easy to miss content there that does require translation. Thanks!

frontend/src/config/config.test.ts
```diff
@@ -0,0 +1,121 @@
...
+              missingFields.push(item);
+            }
+          });
+          if (missingFields.length > 0) {
+            // eslint-disable-next-line no-console
+            console.log(
+              `${missingFields.length} missing fields for ${country} - language: ${key}`,
+            );
+            console.error(missingFields);
+          }
```
gislawill, line 114: I think this is a really helpful change and I'm really thankful that you put this step into existing processes.

A minor suggestion, I found the console output a bit hard to read. I think using console.table could help to read the missing fields

```suggestion
          if (missingFields.length > 0) {
            // Create an array to hold the table data
            const tableData: { [key: string]: string }[] = [];

            // Populate the table data with missing fields
            missingFields.forEach(field => {
              const row: { [key: string]: string } = {
                [`Missing Fields: ${country} - ${key}`]: `${field.slice(0, 64)}${field.length > 64 ? '...' : ''}`,
              };
              // eslint-disable-next-line fp/no-mutating-methods
              tableData.push(row);
            });

            // Print the table to the console
            // eslint-disable-next-line no-console
            console.table(tableData);
          }
```

| Before | After |
| - | - |
| Screenshot 2024-07-11 at 4 59 45 PM |  Screenshot 2024-07-11 at 4 58 58 PM |

gislawill:
Note: the list of missing translations is pretty massive (1756 without including legend labels, 4516 when including legend labels). I think we should consider than automated approach to filling in our current gaps. Either a free version of google translate or an LLM powered approach (probably cost somewhere in the ballpark of $10-$50).

frontend/src/config/config.test.ts
wadhwamatic: > A minor suggestion, I found the console output a bit hard to read. I think using console.table could help to read the missing fields

I like this suggestion!

wadhwamatic:
> Note: the list of missing translations is pretty massive (1756 without including legend labels, 4516 when including legend labels). I think we should consider than automated approach to filling in our current gaps. Either a free version of google translate or an LLM powered approach (probably cost somewhere in the ballpark of $10-$50).

Is that based on every possible translation or the specifically configured ones? I.e, I don't need to ensure French translations are available for Cambodia, nor Khmer for RBD.

gislawill:
> Is that based on every possible translation or the specifically configured ones? I.e, I don't need to ensure French translations are available for Cambodia, nor Khmer for RBD.

I believe that's only for translations configured for the country (example below). So for Cambodia, we're only looking for missing kh translations. I pulled those numbers but looking at the total sum of missing fields we're printing out in all the error messages in the "it should have translations with correct depth" test
Screenshot 2024-07-11 at 6 04 34 PM

wadhwamatic:
Thanks for clarifying that @gislawill !

frontend/src/config/config.test.ts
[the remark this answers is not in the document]
ericboucher: @wadhwamatic I added the labels but I feel like it adds quite a bit of noise. Let me know what you think.
Screenshot 2024-07-12 at 11 31 01 PM

ericboucher:
@wadhwamatic should we merge this and see from there?

wadhwamatic (approved):

frontend/src/config/shared/translation/english.json
```diff
@@ -0,0 +1,123 @@
+{
```
gislawill, line 1: 🙌

gislawill (approved):

frontend/src/config/shared/translation/english.json
ericboucher: We still need to figure out exactly how to make this work / use it since it won't have all the legends and other things. Maybe we could load them the same way we do it for tests in dev mode

ericboucher:
@gislawill I re-actived the logs for `en` in dev mode because we are loading all the other languages keys as `en` s if they are missing in `en` it means they are missing in the other language for that country. I removed the warnings though and only kept an info log of the list.

PR #1327, opened 2024-08-08
Adding reprojection step in calculate_stats if geotiffs' and stats borders' crs don't match; COUNTRY=mozambique

### LINKED ISSUE #1298: [Bug]: Analysis results failing for layers with scale factors [correction: layers without EPSG 4326 projection)
### What happened?

This issue surfaced when testing PR https://github.com/WFP-VAM/prism-app/pull/1297 though is not related to the PR.

If I run analysis for a layer with a scale factor set using `wcsConfig` in layers.json, I get three types of errors: either all values are zero, null, or are equal to the offset value. I tested this in Mozambique and RBD and see the same issue in both cases

These are the errors per layer (see screenshots in the next section)
- NDVI. All values rounded to zero
- NDVI anomaly returns all null values
- LST amplitude. All values rounded to zero
- LST daytime, LST nighttime, INAM air temperature. All values equal to offset (-273)
- LST anomaly returns all null values

Steps to reproduce:
1.  REACT_APP_COUNTRY=mozambique yarn start
2. Run analysis with any of the above hazard layers, lowest level admin boundary as baseline (admin 3 for Mozambique), default for all others
3. View table to see values for each admin area

main / mozambique & RBD

NDVI (all zeroes):
Screenshot 2024-07-04 at 14 48 51

LST (all values = offset)
Screenshot 2024-07-04 at 15 00 57

LST anomaly (null values):
Screenshot 2024-07-04 at 14 53 52

LST amplitude (all zeroes):
Screenshot 2024-07-04 at 14 53 00

gislawill:
### Description

This fixes #1298 by reprojecting MODIS rasters to match the coordinates used by our stats borders.

## How to test the feature:
- [ ] Open Mozambique and navigate to analysis tab
- [ ] Select a MODIS derived hazard layer
- [ ] Run analysis
- [ ] Verify the stats match what you'd expected based on the raster

## Checklist - did you ...
Test your changes with

Screenshot 2024-08-07 at 5 36 03 PM


ericboucher:
Nice work @gislawill. I think it's probably worth adding a few comments throughout the code to explain what's happening.

I would say though that I am tempted to do the reprojection in the STAC processing functions of our backend. It feels a bit more homogeneous that the .tif users download from PRISM are always using the same projection, i.e. `ESPG 4326`

ericboucher:
And let's add one or two tests around this if possible :)

api/app/geotiff_from_stac_api.py
```diff
@@ -114,7 +124,11 @@ def get_geotiff(
     """Generate a geotiff and return presigned download url"""
     s3_filename = generate_geotiff_and_upload_to_s3(collection, bbox, date, band)
 
-    s3_client = boto3.client("s3")
+    s3_client = boto3.client(
+        "s3",
+        aws_access_key_id=STAC_AWS_ACCESS_KEY_ID,
+        aws_secret_access_key=STAC_AWS_SECRET_ACCESS_KEY,
+    )
```
gislawill, line 131: I needed to add these for local development. Considering that this shouldn't have negative impacts and eases development, I thought I'd leave it in

gislawill:
> I would say though that I am also tempted to do the reprojection in the STAC processing functions of our backend. It feels a bit more homogeneous that the .tif users download from PRISM are always using the same projection, i.e. ESPG 4326

Good call, I shifted this functionality over to the `generate_geotiff_from_stac_api` method and simplified the functionality. Now, if the raster get from STAC isn't in EPSG:4326, we project it to EPSG:4326 using rioxarray as recommended by @lowriech. rioxarray readily handles the sinusoidal projection with the custom transformation needed for rasterio's reprojection.

> And let's add one or two tests around this if possible :)

Sure thing, I'll add a couple before raising for review

api/pyproject.toml
```diff
@@ -7,7 +7,7 @@ license = "Apache 2.0"
 
 [tool.poetry.dependencies]
 PyJWT = "2.4.0"
-python = "<3.11,>=3.8.0"
+python = ">=3.10,<3.11"
```
gislawill, line 10: rioxarray requires a python version greater than 3.10. This version update was generated by poetry based on all dependencys' requirements

api/app/raster_utils.py
```diff
@@ -114,3 +115,16 @@ def calculate_pixel_area(geotiff_file):
         area = pixel_width * pixel_height / 1e6
 
         return area
+
+
+@timed
+def reproject_raster(raster_file: FilePath, dst_crs: str, out_file: FilePath):
+    """Reproject a raster file to a new CRS."""
+    x_raster = rioxarray.open_rasterio(raster_file, masked=True)
+    x_raster.rio.reproject(dst_crs).rio.to_raster(out_file)
```
gislawill, line 124: This is thankfully pretty fast, only about 1/5th of a second

gislawill:
This is ready for full review @ericboucher and @lowriech

api/app/geotiff_from_stac_api.py
ericboucher: The default AWS account is the one in which the PRISM instance sits, which is (can be) different from the STAC one.

I think we should avoid linking the two if possible. Locally you can use the default boto3 env variables and set them up like this - https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#environment-variables
Or use the `aws configure` command

If this does not work we can figure out a better way to keep the credentials separate.

ericboucher (approved):
Thanks @gislawill for fixing this. Code lgtm. I would just revert the boto3 env changes if possible.

lowriech:
Looks great so far!  Let's talk more about mocking the data, so that we have something static to test on

lowriech (approved):

api/app/geotiff_from_stac_api.py
gislawill: Oh interesting, I didn't realize that. Thanks @ericboucher. I've reverted this change for now and will experiment more with this once I have full access to AWS to test it out

gislawill:
> Looks great so far! Let's talk more about mocking the data, so that we have something static to test on

Thanks @lowriech. I'm curious to hear more of your thoughts on this. I think that we have a pretty static set up currently by using a stored image and mocking all of our s3 calls ([here](https://github.com/WFP-VAM/prism-app/pull/1327/files#diff-3a28e6600eec3a9a6d69588e48202def476fd087a220984daa0810097850ca54)). Curious if you've seen it done differently/better?

PR #1328, opened 2024-08-08
Google Floods Part 1: Creating gauge points and tooltip graphs; COUNTRY=cambodia

### LINKED ISSUE #1317: [Feature Request]: Google Flood Hub integration

Google has provided us with early access to their API for the Flood Hub: https://sites.research.google/floods/. Details on the API endpoints have been shared separately as access is restricted.

The API provides information on flood status by gauge and by area. 
- Areas are polygons which are accessible through an API endpoint, which I believe includes a location lookup.
- Gauges are points with additional data about the river and country where they live. 
- Flood status includes details on severity and forecasted trend (rising, falling, not changing). 

All of the above information is of high value to PRISM deployments. The end goals are:
1. Create new layers in PRISM which allow for visualization and interaction with gauge and area flood status
2. Handle API authentication through the PRISM backend
3. Allow users to see and interaction with layers in the same fashion seen on the FloodHub site including
4. Visualize gauge status using color scheme used by Google
5. On click of a gauge, show a trend and forecast as seen in the screenshot below
6. Visualize inundation areas 

Screenshot 2024-07-18 at 12 23 39

What is not yet clear is if there is the capability to see flood data on a specific date. I have heard that this is forthcoming but not yet available. For now, I think we should just consider floods as valid for 'today' and keep it date-constrained accordingly. I.e., if we're looking at a rainfall layer for a year ago, we should not show today's flood data alongside it.

_No response_

lowriech:
### Description
This fulfills step 1 and 2 of 3 for #1317 ([see plan](https://github.com/WFP-VAM/prism-app/issues/1317#issuecomment-2277125341)).

In this PR, we're fetching flood status by gauge from Google's Floods API for the entire country (10-100 gauges per country). We're displaying these gauges as points on Prism. They're color coded by flood status and provide past and future flooding in a tooltip when clicked.

Forecasts for each gauge were added here: https://github.com/WFP-VAM/prism-app/pull/1330 (merged into this branch for review)

## How to test the feature:
- [ ] Open the Mozambique or cambodia deployment
- [ ] Enable the Google Flood layer
- [ ] Click on a gauge point

## Checklist - did you ...

Test your changes with

Screenshot 2024-08-12 at 2 03 31 PM
382087325-f448a602-a406-4f64-b6c5-861a8bc25c49


api/app/main.py
```diff
@@ -412,3 +413,10 @@ def post_raster_geotiff(raster_geotiff: RasterGeotiffModel):
     return JSONResponse(
         content={"download_url": presigned_download_url}, status_code=200
     )
+
+
+@app.get("/google-floods-gauges")
```
ericboucher, line 418: If we're gonna add a bunch of integrations, it will be worthwile to think about a standardized approach and common architecture. And maybe expose them through a sub endpoint of the API to keep things cleaner. wdyt @lowriech ?

gislawill: I think that's a good question, @ericboucher. Do you see other APIs we're already leveraging fitting into this integration category (kobo, hdc stats, others)? Or are you thinking just for new integrations? 

I'm wary of premature abstraction and trying to determine if I think now is the right time to align on this standardized approach.

frontend/src/config/mozambique/layers.json
```diff
@@ -2954,5 +2954,46 @@
     "start_date": "2020-12-01",
     "end_date": "2021-07-31",
     "opacity": 0.7
+  },
+  "google_flood_gauges": {
+    "title": "Google Flood Gauges",
+    "type": "point_data",
+    "hex_display": false,
+    "data": "http://0.0.0.0:80/google-floods/gauges/?region_code=MZ",
```
gislawill, line 2962: Note: need a better development set up

gislawill:
@lowriech and @ericboucher, there are a couple outstanding issues (noted in the PR description) but this ready for initial review

frontend/src/components/MapView/utils.ts
```diff
@@ -136,24 +166,32 @@ const getData = (
...
-      (obj, item) => ({
+    .filter(prop => keys.includes(prop) && prop !== 'title')
+    .reduce((obj, item) => {
+      const itemProps = featureInfoProps[item] as FeatureInfoProps;
+      if (
+        itemProps.visibility === FeatureInfoVisibility.IfDefined &&
+        !properties[item]
+      ) {
+        return obj;
+      }
```
gislawill, line 177: We have some feature properties that may or may not be defined by the google api (for example: site name, river name). This allows us to hide the property from the pop up if it's not defined

api/app/googleflood.py
```diff
@@ -0,0 +1,106 @@
...
+                data["gaugeLocation"]["latitude"],
+            ],
+        },
+        "properties": {
+            "gaugeId": data["gaugeId"],
+            "issuedTime": data["issuedTime"],
+            "siteName": data["siteName"],
+            "river": (
+                data["river"] if "river" in data and len(data["river"]) > 1 else None
+            ),
```
gislawill, line 34: "river" can come in as a string with just 1 space (`" "`). This ensures there's a real value

frontend/src/config/mozambique/layers.json
```diff
@@ -2954,5 +2954,74 @@
...
+        "visibility": "if-defined"
+      },
+      "river": {
+        "type": "text",
+        "template": "River: %s",
+        "visibility": "if-defined"
+      },
+      "gaugeId": {
+        "type": "text",
+        "template": "Gauge ID: %s",
```
ericboucher, line 2987: Let's think of a way to make this template translatable?

api/app/main.py
lowriech: It seems like low-hanging fruit to abstract once we decide on a pattern.  So yes, agree we should think about it, but not sure about the velocity of adding entirely new providers.  Would vote to wait until we see more pattern around this.

frontend/src/config/mozambique/layers.json
ericboucher: We haven't really done something like that in the past, we would usually have the template be a title and then handle them separately. So instead of template you would just have `title: Gauge ID`.

Another option is to use the translation mapping and have the template be: `Gauge ID: {id}`, then 

          <Trans values={{ id }}>
            <Typography>{t(template)}</Typography>
          </Trans>

gislawill: Thanks @ericboucher! I'm leaning towards the latter option (translation mapping). We used these templates in react-i18next pretty extensively in the distant past (my Hilton Hotels days). Once we started using them, it gave us a ton of flexibility in terms of copy. If you're ok with it, I'll implement it this way.

frontend/src/config/mozambique/layers.json
gislawill: @ericboucher have you all deployed the prism-api out in surge-like deployments before for testing?

gislawill:
Note: I'm adding [pytest-recording](https://github.com/kiwicom/pytest-recording) to store our google flood responses in local tests and use them in CI/CD. That will fix the failing API test in this PR

frontend/src/config/mozambique/layers.json
gislawill: Updated

api/app/googleflood.py
```diff
@@ -0,0 +1,118 @@
...
+            "coordinates": [
+                data["gaugeLocation"]["longitude"],
+                data["gaugeLocation"]["latitude"],
+            ],
+        },
+        "properties": {
+            "gaugeId": data["gaugeId"],
+            "issuedTime": data["issuedTime"],
+            "siteName": data["siteName"],
+            "river": (
```
gislawill, line 33: - [ ] @gislawill update this to riverName

api/app/tests/cassettes/test_google_floods_api/test_get_google_floods_gauges_api.yaml
```diff
@@ -0,0 +1,258 @@
...
+      accept-encoding:
+      - gzip, deflate
+      connection:
+      - keep-alive
+      host:
+      - testserver
+      user-agent:
+      - testclient
+    method: GET
+    uri: http://testserver/google-floods/gauges/?region_codes=BD
```
gislawill, line 227: @gislawill make sure these requests to testserver aren't saved

gislawill:
> In the meantime, in addition to displaying the gauge ID, can we first show the river name? I see that in Flood Hub interface, hopefully it's also exposed via the API. 
> Note, we'd need to translate that text as well. I see that Flood Hub offers several language options including Khmer. Ex: https://sites.research.google/floods/e/l/10.96043262736797/103.63535499532549/8.45328318786621/g/hybas_4121148610?layers&hl=km

Hey @wadhwamatic, the river name is sometimes provided by the API but often it is not. I've currently set this up so we display the site name if it's provided. If not, we display the river name if it's provided. If not, we display the gauge id (which is always provided). How does that logic sound to you? 

Here's how it looks:
| This PR | Next PR |
| - | - |
| Screenshot 2024-10-07 at 3 43 11 PM | Screenshot 2024-10-07 at 3 39 17 PM | 

I tried to find the translated names in the API but unfortunately there's no option in the API documentation and the API fails if the hl parameter is included in the request.
Screenshot 2024-10-07 at 3 49 03 PM

I'll update this title component so that at least "site name:" and "river:" can be translated even in the API provided name won't be.

wadhwamatic:
> Hey @wadhwamatic, the river name is sometimes provided by the API but often it is not. I've currently set this up so we display the site name if it's provided. If not, we display the river name if it's provided. If not, we display the gauge id (which is always provided). How does that logic sound to you?

That logic sounds good - thanks! Is there an example of site name in Cambodia? 

That's too bad about the translation via the API but thanks for noting it regardless.

gislawill:
> That logic sounds good - thanks! Is there an example of site name in Cambodia?

There is not unfortunately. A bunch of rivers but no site names

gislawill:
Heads up @wadhwamatic, the today "Date" is now set on these layers
Screenshot 2024-10-09 at 10 00 54 PM

I'll work with @ericboucher to get a new server deployment out for testing

gislawill:
Server deployment containing the updates is out (thanks @ericboucher for walking me through it). @wadhwamatic you can resume testing

wadhwamatic:
@gislawill - thanks for adding the time dimension here. I tested date intersections a few ways and this is looking good from my side, as does the river gauge info noted in this PR

gislawill:
Remaining todo items:
- [ ] Ensure all parts of tooltip are translate-able

gislawill:
@wadhwamatic and @ericboucher, this PR is ready for review. Note: it contains the updates from #1330 to ease the review. There were tooltip features in this PR originally that were overridden by #1330. Now that they're together, we don't need to review code/ui that'll be immediate removed

api/app/tests/test_google_floods_api.py
```diff
@@ -0,0 +1,126 @@
...
+    filter_query_parameters=["key"],
+)
+def test_get_google_floods_gauges():
+    """
+    This test is not used in the API, but is used to test the get_google_floods_gauges function
+    It is used to ensure that the get_google_floods_gauges function returns a valid GeoJSON object
+    """
+    gauges = get_google_floods_gauges(["BD"])
+    assert gauges["type"] == "FeatureCollection"
+    for feature in gauges["features"]:
```
ericboucher, line 21: Note that this would not test anything if length of gauges["features"] is 0

api/app/tests/test_google_floods_api.py
```diff
@@ -0,0 +1,126 @@
...
+
+@pytest.mark.vcr(
+    match_on=["uri", "method"],
+    filter_headers=["Authorization"],
+    filter_query_parameters=["key"],
+    ignore_hosts=["testserver"],
+)
+def test_get_google_floods_gauges_api():
+    """
+    This test is used to test the API endpoint for getting Google Floods gauges
```
ericboucher, line 40: You can replace "This test is used to test" by "Test ..."

api/app/tests/test_google_floods_api.py
gislawill: Good call, I'll add a length check

api/app/googleflood.py
```diff
@@ -0,0 +1,220 @@
...
+        for future in as_completed(future_to_region):
+            status_response = future.result()
+            if "error" in status_response:
+                logger.error("Error in response: %s", status_response["error"])
+                raise HTTPException(
+                    status_code=500,
+                    detail="Error fetching flood status data from Google API",
+                )
+            flood_statuses.extend(status_response.get("floodStatuses", []))
+
```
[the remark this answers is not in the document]
gislawill: Sure, I can abstract it out

wadhwamatic:
Looks good, thanks Will! I created an issue for the other (pre-existing) bug I mentioned: https://github.com/WFP-VAM/prism-app/issues/1365. If it's simple to address here, please do so otherwise we can address in a separate PR

frontend/src/components/MapView/MapTooltip/index.tsx
```diff
@@ -42,11 +49,13 @@ const useStyles = makeStyles(() =>
       marginBottom: '4px',
     },
     popup: {
+      // Overrides the default maxWidth of 240px set by react-map-gl
+      maxWidth: 'none !important',
```
gislawill, line 53: Heads up @wadhwamatic, this fixes #1365

wadhwamatic:
@gislawill - I noticed today that the charts module has a new bug on this branch. If I go to charts, select an indicator, and then drag the time slider, the chart is not updating to show the new date range even though the axis labels change.

Also, if I choose an indicator that uses a line chart (ex: rainfall anomaly), the thickness of the lines has increased compared to what was previously there.

gislawill:
> If I go to charts, select an indicator, and then drag the time slider, the chart is not updating to show the new date range even though the axis labels change. Also, if I choose an indicator that uses a line chart (ex: rainfall anomaly), the thickness of the lines has increased compared to what was previously there.

Thanks for catching these issues! I'll check it out. 

For the second issue (thickness of the line increasing), that's a result of me purposefully making the line thicker for the flood graphs (easier to interpret imo). @wadhwamatic could you confirm you prefer the thinner line in the charts module? Do you have a stance on the weight of the floods line?

wadhwamatic:
> For the second issue (thickness of the line increasing), that's a result of me purposefully making the line thicker for the flood graphs (easier to interpret imo). @wadhwamatic could you confirm you prefer the thinner line in the charts module? Do you have a stance on the weight of the floods line?

@gislawill - I thought that might be the case. I personally prefer a bit thinner. I added a comment, suggesting borderWidth: 3 instead of 4. Sound / look ok to you?

It looks like this:
Screenshot 2024-11-15 at 18 53 20

Screenshot 2024-11-15 at 18 53 10

frontend/src/components/Common/Chart/index.tsx
```diff
@@ -178,7 +193,7 @@ const Chart = memo(
           fill: config.fill || false,
           backgroundColor: colors[i],
           borderColor: colors[i],
-          borderWidth: 2,
+          borderWidth: 4,
```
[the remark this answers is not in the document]
gislawill: Looks good to me, updated

wadhwamatic:
Thanks @gislawill. The line thickness looks good to me and the time slider issue has been resolved

gislawill:
Thanks @wadhwamatic, is this good to go as far as you're concerned?

Some thoughts on potential remaining steps:
- [ ] How does the copy look?
- [ ] Is the layer set up optimal in each deployment (should this be in it's own layer group as currently implemented or put in an existing layer group)
- [ ] We'll need translations once the copy and layer organization is complete

I'll note for context that the 502 errors from Google in RBD is not an issue on this branch. It seems to only crop up in #1332 (where I've added a cache-based workaround)

wadhwamatic:
@ericboucher - can you make a final pass on this PR?  It's gtg from my side. Let's please remove the google flood layers from prism.json in the demo countries before merging into main though but bring it back subsequently in https://github.com/WFP-VAM/prism-app/pull/1332

api/app/googleflood.py
```diff
@@ -0,0 +1,184 @@
...
+        "type": "Feature",
+        "geometry": {
+            "type": "Point",
+            "coordinates": [
+                data["gaugeLocation"]["longitude"],
+                data["gaugeLocation"]["latitude"],
+            ],
+        },
+        "properties": {
+            "gaugeId": data["gaugeId"],
```
ericboucher, line 32: How confident are we that all the keys will be present? Should ze use a getter with a default None to be safe?
Or make sure that we have proper error handling for missing keys.

api/app/utils.py
```diff
@@ -38,3 +40,29 @@ def filter(self, record):
             self.warning_count += 1
 
         return False
+
+
+def make_request_with_retries(
```
ericboucher, line 45: Do we need something specific, or could we rely on the Retry util directly?
from requests.adapters import HTTPAdapter, Retry

frontend/src/config/rbd/layers.json
```diff
@@ -6531,5 +6531,77 @@
       }
     ],
     "legend_text": "Joint Food Security & Nutrition Hotspot Analysis conducted by WFP, UNICEF, and partners based on SMART surveys and Cadre Harmonisé (CH) & Integrated Food Security Phase Classification (IPC) analyses"
+  },
+  "google_flood_status_at_gauges": {
```
ericboucher, line 6535: Should this layer be shared ans then just override the necessary param?

frontend/src/utils/fetch-with-timeout.ts
```diff
@@ -8,7 +8,7 @@ interface FetchWithTimeoutOptions extends RequestInit {
 
 export const ANALYSIS_REQUEST_TIMEOUT = 60000;
 
-const DEFAULT_REQUEST_TIMEOUT = 15000;
+const DEFAULT_REQUEST_TIMEOUT = 30000;
```
ericboucher, line 11: Still needed? Is it to account for the retries?

frontend/src/components/Common/Chart/index.tsx
```diff
@@ -208,8 +223,21 @@ const Chart = memo(
           fill: false,
         }));
       }
+      if (data.GoogleFloodConfig) {
```
ericboucher, line 226: This PR is introducing a lot of extra complexity to this file. I'll open an issue recommending to split it by functions/services

ericboucher (approved):
Nice work overall. Let's make sure to QA all different tooltips inc. EWS.
I left a few comments in line but this looks gtm overall !

api/app/googleflood.py
gislawill: We're pretty confident that they'll be present but there's no downside to the default none to be safe. I've updated to add a getter for non-essential values and error handling for missing essential values

api/app/utils.py
gislawill: The retry utility has some useful functionality that this approach didn't have (back-off timing and status filtering for retries specifically). I've updated this function to wrap the retries utility. This `make_request_with_retries` function is still useful to reduce redundancy

frontend/src/config/rbd/layers.json
gislawill: I haven't implemented a shared layer before and I'm still trying to understand how it works. I see in the README:
> Since many layers are common across multiple countries, we created shared configuration files that any deployment can access. These layers are generated by WFP globally and made available through the Humanitarian Data Cube.

Is the HDC part meaningful? Based on a quick scan, the current shared layers do appear to all be coming from HDC but I don't see why they _need_ to (technologically speaking that is).

Also, @ericboucher, could you point me to examples of shared layers with overrides? I see examples of shared layers with duplicates in countries but I'm not seeing overrides.

frontend/src/utils/fetch-with-timeout.ts
gislawill: It's still needed. Not necessarily for retries but for loading all of the data needed for RBD. The server needs to make a lot of separate requests to get all of the various country data and, even with parallelization, it can take about 20 seconds.

wadhwamatic: @gislawill - I'm still seeing "Impossible to get point data dates for forecast_river_discharge_at_gauges" error on the surge deployment and locally. Any thoughts on this?

gislawill: @wadhwamatic there are two issues right now:
- The current deployment of our production PRISM server no longer has the functionality from this branch
  - This is resulting in 404 errors because the route to get the flood dates isn't found on the server
  - @ericboucher, any objection to me running a deployment from this branch to production? It will have the latest on main merged in
- The google floods API seems to be down again. I'm seeing 502 "Bad gateway" errors when hitting these endpoints locally. I'll reach out to Aviel about it.

gislawill:
> Nice work overall. Let's make sure to QA all different tooltips inc. EWS. I left a few comments in line but this looks gtm overall !

Confirmed EWS tooltips look good
Screenshot 2025-01-05 at 6 29 50 PM

frontend/src/utils/fetch-with-timeout.ts
gislawill: @wadhwamatic redeploying the server fixed it 
Screenshot 2025-01-06 at 9 08 46 PM

frontend/src/config/rbd/layers.json
gislawill: @ericboucher, adding this layer to our shared layer config caused it to load available dates for every deployment — regardless of whether or not the deployment used this layer. I think it's better to only add this layer to deployments that are using it until we're more selective about which layers we're loading data for

ericboucher:
@gislawill are we good to merge?

gislawill:
I believe we are good to go. I'd love to have @wadhwamatic's approval before merging.

wadhwamatic:
sorry @gislawill, missed the fix for the png download. yes, gtg now. thanks!

PR #1332, opened 2024-08-22
Google Floods Part 3: Adding Inundation Maps; COUNTRY=mozambique

lowriech:
This adds inundation polygons from the Google Floods API.

https://www.loom.com/share/5b430b093ee546139f5ccbe3299c60c9?sid=049273de-18ff-4b6a-8718-b1a164c82b52


api/app/googleflood.py
```diff
@@ -55,3 +61,45 @@ def get_google_floods_gauges(
...
+    level: str
+    serializedPolygonId: str
+
+class InundationMapSet(BaseModel):
+    inundationMaps: List[InundationMap]
+
+
+def get_google_floods_inundations(
+    inundationMapSet: InundationMapSet,
+) -> gpd.GeoDataFrame:
```
gislawill, line 77: Follow up from conversation, we're going to fetch the gauge by region here then fetch the maps — all backend

lowriech:
@gislawill this is updated to call `gauges`, rather than rely on a previous call passing its attributes

gislawill:
Seeing kind of a weird bug while doing QA, the flood layer is loading in with various shades for the same data. It's very intermittent — changes with each page reload. I'm digging into figure it out. @ericboucher, have you see this in the past with any other layers?

| Darkest (correct) | Medium | Light |
|-|-|-|
| Screenshot 2024-10-07 at 2 41 10 PM | Screenshot 2024-10-07 at 2 38 18 PM | Screenshot 2024-10-07 at 2 38 47 PM |

ericboucher:
@gislawill I have not seen this before. Could this be a bug linked to the opacity?

gislawill:
> @gislawill I have not seen this before. Could this be a bug linked to the opacity?

I don't believe so - the opacity is always coming through as .9 (tested it at 1 just to be sure too). I think it may have to do with layer order. The theory being that if it's rendered under 1 admin level its somewhat lighter, 2 admin layers and it's much lighter. The layer order in the canvas doesn't necessary match the layer order in jsx so I think I'll have to play around with "beforeId" prop to test it out

wadhwamatic:
@gislawill - I've not been able to access the flood inundation layer through the surge deployment, but it works locally. Here's the error I see on surge:
Screenshot 2024-10-10 at 13 46 55


Things look good locally. I've not seen the shading error you mentioned.

gislawill:
> @gislawill - I've not been able to access the flood inundation layer through the surge deployment, but it works locally. 

Oh weird, it looks like this https request to the prism-api is getting redirected to a http request _by_ the prism-api. I'm not sure why that would be but I'll look into it

api/app/tests/cassettes/test_google_floods_api/test_get_google_floods_inundation_api.yaml
gislawill, line 1: These two files are each about 4 MB — which is unfortunately how large the response from Google is at this endpoint.

gislawill:
Note: on 11/7/24, I received this error message intermittently but very frequently. Hoping the google floods has resolved it. Or will need to reach out to them

Request failed at url post https://floodforecasting.googleapis.com/v1/floodStatus:searchLatestFloodStatusByArea?key={GOOGLE_FLOODS_API_KEY}: {'regionCode': 'ML'}
Response status code: 502
Response content: (abbreviated)
```
    <title>Error 502 (Server Error)!!1</title>
    <p>The server encountered a temporary error and could not complete your request.
    <p>Please try again in 30 seconds.  <ins>That's all we know.</ins>
```

wadhwamatic:
looks good from my side Will. some surprising discrepancies in Cambodia between NASA SERVIR flood extent and the Google model, but we've also known there to be overestimation by the SERVIR tool.

gislawill:
The 502 errors from Google Floods have continued and seem to be a result of us hitting the searchLatestFloodStatusByArea endpoint too frequently. I've just emailed the google floods team about this issue.

In the meantime, I've added google floods response caching which gets around the issue for now. I've set the cache timeout to only 10 minutes as we know this data can change quickly. 

@ericboucher, I'm curious for your thoughts on this caching ([a45de24](https://github.com/WFP-VAM/prism-app/pull/1332/commits/a45de247ac10af8109541046cb12b02255b423ba). Specifically,
- I'm not seeing any cache clearing so far. These new cached files could bloat the EC2 instance
- Would it be helpful to include the cache timeout logic in the caching utilities for easier use elsewhere?

api/app/caching.py
```diff
@@ -98,13 +98,13 @@ def cache_file(url: str, prefix: str, extension: str = "cache") -> FilePath:
 
 
 @timed
-def cache_geojson(geojson: GeoJSON, prefix: str) -> FilePath:
+def cache_geojson(geojson: GeoJSON, prefix: str, cache_key: str) -> FilePath:
```
ericboucher, line 101: Please explicit the inputs here, esp. the cache_key and how it should be used.

api/app/googleflood.py
```diff
@@ -30,13 +38,31 @@ class InundationMapSet(BaseModel):
...
+        prefix="google_floods", data=cache_key, extension="json"
+    )
+
+    # Check if the cached file is valid and not expired
+    if is_file_valid(cache_filepath):
+        cached_data = get_json_file(cache_filepath)
+        cache_age = get_cache_age(cache_filepath)
+        if cache_age < CACHE_TIMEOUT:
+            logger.debug("Returning cached data.")
+            return cached_data
```
ericboucher, line 63: You could return `get_json_file(cache_filepath)` on the last line and remove the step above `cache_age = ` ... (i.e. remove L59)

ericboucher:
@gislawill thanks for this. Yes I think it could be useful to add the cache timeout mechanism for other requests as well.

And for the cache clearing, I think it would be worth implementing but not sure if there is an easy way. We could do something simple like delete any file that's more than 30 days old or something like this. @wadhwamatic any thoughts on that point?

api/app/googleflood.py
gislawill: Updated

api/app/caching.py
gislawill: Added some add comments and updated the params to be more explicit

wadhwamatic:
> And for the cache clearing, I think it would be worth implementing but not sure if there is an easy way. We could do something simple like delete any file that's more than 30 days old or something like this. @wadhwamatic any thoughts on that point?

@ericboucher - yes, we should delete older files in this case and 30 days sounds like a good initial period to try

gislawill:
@wadhwamatic and @ericboucher, we have many different options for clearing this cache. I think whichever option we go with, it should be easy to trigger the cache clearing automatically and manually. We'll also want to be able to easily adjust the "minimum age to clear" to allow for experimentation with this cache age and for clearing caches after a bug fix rollout.

The two good options would be a shell script and an endpoint:
- **Script** 
    - A Python script (could be Bash too) that loops through the cache directory and clears files based on age
    - The age would default to 30 days but could be provided as an environment variable
        - A provided age of 0 would clear all cached files
    - This could be triggered manually by SSH-ing into the container and running it
- **Endpoint**
    - A FastAPI endpoint that loops through the cache directory and clears files based on age
    - The age would default to 30 days but could be provided in the request body
    - This could be triggered manually by any request client (i.e. curl)
        - Very easy, though not secure (but can't think of any real risks associated with it)

Two options for automated triggers:
    - A cron daemon in the Docker image to keep simple
    - An ECS Scheduled Task which is a slightly more complicated set up but easier to monitor and more resilient (retries, alerting)

I'm leaning towards a script that's triggered by an ECS Scheduled Task. Curious for your thoughts. Also curious if either of you would have a preference between adding this cache clearing in this PR vs a follow-up?

ericboucher:
@gislawill the endpoint option seems to be convenient and allows any of us to trigger it manually when necessary. We could "secure" it with a token or something simple, that should be enough to test things out.

I'd prefer to have this added in a follow-up PR as it would make it a bit easier to discuss implementation and triggers. If we go with the endpoint option, we could use a simple GitHub action to run daily and call this endpoint.

gislawill:
Sounds good to me @ericboucher. @wadhwamatic I've created issue https://github.com/WFP-VAM/prism-app/issues/1371 for this work. 

This PR should be ready for review. It's blocked on rollout by the copy + translations questions here: https://github.com/WFP-VAM/prism-app/pull/1328#issuecomment-2483487510

ericboucher:
@gislawill can you resolve the conflicts when you get a chace? :)

wadhwamatic:
@gislawill - I'm still getting the errors mentioned on Slack when I click on a gauge:
Screenshot 2025-02-20 at 18 09 24

The red flood extent improves visibility I think but there's also the issue of the river following a boundary and becoming difficult to see. I'm not sure what else we can do about that though. The same lack of visibility when zoomed out issue is on Google Flood Hub. Not something we need to work on now, but maybe drawing a polygon on the map of a flood area when zoomed out could be useful. Thoughts on level of effort to do something like this?
Screenshot 2025-02-20 at 18 10 54

gislawill:
Apologies for the delay @wadhwamatic, responses below:
> I'm still getting the errors mentioned on Slack when I click on a gauge

That turned out to be a deployment issue (the server deployment from this PR had been replaced). I've redeployed this branch and it's working once again. 



> There's also the issue of the river following a boundary and becoming difficult to see. I'm not sure what else we can do about that though.

I was able to ensure that this layer is rendered above the boundary layer. The latest surge deployment (in about 10 minutes) will show the flood extents on top of the boundary layer.
Screenshot 2025-03-04 at 10 40 24 AM



> Maybe drawing a polygon on the map of a flood area when zoomed out could be useful. Thoughts on level of effort to do something like this?

That's certainly possible. The approach would look like the following:
- Fetch the KMLs from Google Floods API (already doing)
  - We get many vectors in regions with flooding
- Group them based on proximately (spatial clustering) (new)
- Calculate the bounding box for each cluster of flood maps (new)
- Send a geojson representation of the group's bounding box to PRISM (new)
- Add a new layer type for the bounding box in PRISM (new)

The backend work sounds more complicated but would be pretty straightforward to implement (1-2 hours?). The frontend work ("a new layer type for the bounding box") is easy but time consuming (updating many different files) (2-4 hours?).

gislawill:
Heads up @wadhwamatic, this server deployment has been superseded so the gauge details errors will occur on this branch again. Let me know when you're available to test and I'll make sure the server can support this branch for testing

wadhwamatic:
Thanks for addressing the conflicts yesterday @gislawill. I saw some flood events in Angola on Flood Hub today and updated the Mozambique config for this branch (just locally) and was able to see this in action. 

It's looking good from my side.

config changes (not going to commit them) were simply to use the Angola ISO code in place of Mozambique

Screenshot 2025-03-28 at 14 48 09

Here are the gauges and flood event:
Screenshot 2025-03-28 at 14 44 40

Map export worked well too:
Angola flooding_2025_03_28

api/README.md
```diff
@@ -141,19 +141,21 @@ This should open playwright in debug mode, with a browser window and a debugging
...
-```
+To deploy, ssh into the EC2 instance:
+- Get the private key and copy it to `~/.ssh/{some name}.pem`
+- Add your IP address to the EC2 instance's whitelist in the AWS console
+- Run `ssh -i ~/.ssh/{some name}.pem ubuntu@{Public IPv4 DNS}`
+- Navigate to the api directory
+- Confirm you're on the right branch and the branch is up to date
+- Run `make deploy`
 
 There are a few known issues happening from time to time
```
ericboucher, line 160: @gislawill maybe add the docker/volume issue in here for future reference?

api/app/caching.py
```diff
@@ -98,13 +98,30 @@ def cache_file(url: str, prefix: str, extension: str = "cache") -> FilePath:
...
+def cache_geojson(
+    geojson: GeoJSON, prefix: str, cache_key: str | None = None
+) -> FilePath:
+    """
+    Locally cache geojson for future use.
+
+    Args:
+        geojson (GeoJSON): The GeoJSON object to be cached.
+        prefix (str): A prefix to be used in the cache file's name.
+        cache_key (str | None, optional): A unique key to identify the cache entry.
```
ericboucher, line 110: Could you briefly explain why we need the cache_key?

btw I've been tempted for a while do de-hash part of the file name to make looing at the cache and debugging a bit easier.

api/app/googleflood.py
```diff
@@ -1,22 +1,81 @@
...
+
+# This cache helps avoid making repeated requests to the Google Floods API
+CACHE_TIMEOUT = 600  # 10 minutes
+GOOGLE_FLOODS_API_KEY = os.getenv("GOOGLE_FLOODS_API_KEY", "")
 if GOOGLE_FLOODS_API_KEY == "":
     logger.warning("Missing backend parameter: GOOGLE_FLOODS_API_KEY")
 
 
+def make_google_floods_request(url, method="get", data=None, retries=1, timeout=10):
+    """Make a request with retries and error handling, caching the result for 5 minutes."""
```
ericboucher, line 44: Timeout seems to be 10min? Maybe no need to specify the time here, you could just say "CACHE_TIMEOUT"

api/app/googleflood.py
```diff
@@ -47,37 +106,60 @@ def format_gauge_to_geojson(data):
...
     """Fetch flood statuses concurrently for a list of region codes."""
     flood_statuses = []
-    with ThreadPoolExecutor() as executor:
-        future_to_region = {
-            executor.submit(fetch_flood_status, code): code for code in region_codes
-        }
-        for future in as_completed(future_to_region):
-            status_response = future.result()
+    if run_sequentially:
+        # Run synchronously
```
ericboucher, line 115: I think the use of "synchronously" is a bit misleading here. The boolean is self explanatory, no need for the comment I think

api/app/googleflood.py
```diff
@@ -47,37 +106,60 @@ def format_gauge_to_geojson(data):
...
+) -> list[dict]:
     """Fetch flood statuses concurrently for a list of region codes."""
     flood_statuses = []
-    with ThreadPoolExecutor() as executor:
-        future_to_region = {
-            executor.submit(fetch_flood_status, code): code for code in region_codes
-        }
-        for future in as_completed(future_to_region):
-            status_response = future.result()
+    if run_sequentially:
```
ericboucher, line 114: When would we prefer one vs the other?

api/app/utils.py
```diff
@@ -6,6 +6,8 @@
 
 logger = logging.getLogger(__name__)
 
+logger = logging.getLogger(__name__)
```
ericboucher, line 9: Why is this here twice?

frontend/src/components/MapView/Layers/GeojsonDataLayer/index.tsx
```diff
@@ -0,0 +1,65 @@
+import { memo, useEffect } from 'react';
```
ericboucher, line 1: Would it make sense to dedup and merge with the Point layer? since it's also a geojson techncally?
No need to do it in this PR but maybe open an issue if you believe it's a good idea @gislawill

ericboucher (approved):
Nice work! Tested with AO local changes as well.

ericboucher (approved):
Nice work! Tested with AO local changes as well.

api/app/caching.py
gislawill: The idea between the cache_key was to have a declarative string to store the cache under an unhashed file name for easier readability. In practice, I inadvertently hashed it as well as so it didn't make a difference. I'm going to update these arguments to be a bit more declarative about what's going on.

I don't love the name `data` here — it's sometimes used for actual json data and other times used as string (like url). Instead, I'm updating so we have one argument: `cache_lookup` that can handle both strings and objects. I'll add a new argument called `hash_lookup` that defaults to false.

api/README.md
gislawill: Good call, updated 👍

api/app/googleflood.py
gislawill: Agreed, updated

api/app/googleflood.py
gislawill: Added some notes to the code (thanks for flagging). It's very useful for debugging but slower.

api/app/utils.py
gislawill: Oops, yep! Thanks

frontend/src/components/MapView/Layers/GeojsonDataLayer/index.tsx
gislawill: I do like the idea of merging them considering the returned rendered jsx is so similar. There's enough different in the data fetching/preparation that I'm hesitant to add too many more if statements to that component.

A half formed idea: how would you feel about experimenting with separating out the logic of preparing the data from the logic of the jsx rendering? This would involve moving a lot of the following into a parent component of sorts that's more specific to prior "point data" data sources + applications. This part wouldn't be shared with the GeoJSON data layer but they would share a render component?
```
// Move to Parent Component
const PointDataLayer = memo(({ layer, before }: LayersProps) => {
  const layerId = getLayerMapId(layer.id);

  const selectedDate = useDefaultDate(layer.id);
  const serverAvailableDates = useSelector(availableDatesSelector);
  const layerAvailableDates = serverAvailableDates[layer.id];
  const userAuth = useSelector(userAuthSelector);

  useMapCallback('click', layerId, layer, onClick);

  const queryDate = getRequestDate(layerAvailableDates, selectedDate);
  const validateLayerDate = !layer.dateUrl || queryDate;

  const layerData = useSelector(layerDataSelector(layer.id, queryDate)) as
    | LayerData<PointDataLayerProps>
    | undefined;
  const { updateHistory, removeKeyFromUrl, removeLayerFromUrl } =
    useUrlHistory();

  const { data } = layerData || {};

  useEffect(() => {
    if (layer.authRequired && !userAuth) {
      return;
    }

    if (!data && validateLayerDate) {
      dispatch(loadLayerData({ layer, date: queryDate, userAuth }));
    }
  }, [
    data,
    dispatch,
    userAuth,
    layer,
    queryDate,
    layerAvailableDates,
    validateLayerDate,
  ]);

  useEffect(() => {
    if (data && !data.features && layer.authRequired) {
      dispatch(
        addNotification({
          message: 'Invalid credentials',
          type: 'error',
        }),
      );

      dispatch(removeLayerData(layer));
      dispatch(clearUserAuthGlobal());
      return;
    }

    if (
      data &&
      (data as PointLayerData).features?.length === 0 &&
      layer.authRequired
    ) {
      dispatch(
        addNotification({
          message: `Data not found for provided date: ${getFormattedDate(
            selectedDate,
            'default',
          )}`,
          type: 'warning',
        }),
      );
    }
  }, [
    dispatch,
    layer,
    data,
    selectedDate,
    userAuth,
    removeKeyFromUrl,
    removeLayerFromUrl,
    updateHistory,
  ]);

  if (!data || !validateLayerDate) {
    return null;
  }
```

ericboucher: Yep why not, I think it would also make sense to think about these changes with "uploading geojson files". cc @wadhwamatic

api/app/googleflood.py
ericboucher: ? sorry my point was to highlight that the description says  `caching the result for 5 minutes` but uses `CACHE_TIMEOUT` which is set before to 10min, no?

ericboucher: no biggy though

wadhwamatic:
@gislawill & @ericboucher - is this ready to merge?

api/app/googleflood.py
gislawill: Ah, I misunderstood your comment, thought you were referring to the timeout on line 43 😅

I've updated the comment

PR #1341, opened 2024-09-12
layer config updates; COUNTRY=jordan

wadhwamatic:
### Description

This PR reconfigures seasons for CDI layers and adds missing Arabic translations (WIP) 

Note that seasons don't work correctly as the current season doesn't have sufficient data. This can be addressed in our [temporary date selection hack](https://github.com/WFP-VAM/prism-app/pull/1326) but it seems we should also better handle the lack of data for the full period without the hack.

## How to test the feature:

- [ ] Toggle language to Arabic and confirm translations are accounted for
- [ ] Select seasonal CDI and see that we have two seasons in a year
- [ ] Note that the data lag issue in hip-analysis will cause the initial request to fail. Select a date before 1-May-2024 to see seasonal CDI load correctly

## Checklist - did you ...

Test your changes with


gislawill (approved):
Looks good to me, thanks!

PR #1346, opened 2024-09-19
Updating tests that failed in CI

gislawill:
### Description
These are failing in CI due to an error from the OWS. I think we'll need @valpesendorfer help figuring out why they're failing in CI but not when the test is run locally. Skipping until then.

The tests are failing due to a 500 from https://api.earthobservation.vam.wfp.org/ows/?service=WCS&request=GetCoverage&version=2.0.0&coverageId=wp_pop_cicunadj&subset=Long(95.71,96.68)&subset=Lat(19.42,20.33)

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with


ericboucher:
@gislawill I think it would be better to update this test instead to not use a WCS request

gislawill:
Thanks Eric. This update looks good to me. One thought, could we use the server in our test suite (the locally running test server) rather than the production server? Would rather rely on a the functionality already in our test suite than introducing a production dependency. 

Also, I'm currently in the middle of a move but can test that suggestion out later this week

PR #1370, opened 2024-11-22
[DRAFT]: Divided AA into drought and storm layers

Doniaab:
### Description

- [Create another type of anticipatory action : Anticipatory Action Storm]
- [ Update code to handle multiple anticipatory actions]
- [Update nav items to handle childrens, add menu selection item on click]
- [Separates AAStorm panel and AADrought panel]

## How to test the feature:
- [Click on the anticipatory action menu to switch between AAStorm and AADrought layers]
- [Panels should be differents]

## Checklist - did you ...

Test your changes with


frontend/src/components/MapView/LeftPanel/index.tsx
```diff
@@ -124,31 +135,45 @@ const LeftPanel = memo(() => {
...
+    const layer = LayerDefinitions[selectedLayerId];
+    console.log('layer', layer);
+    console.log('AALayerInUrl', AALayerInUrl);
     // Add to url when getting to AA tab
-    if (AALayerInUrl !== undefined || !layer) {
+    if (!layer) {
       return;
     }
 
+    // Change AA layer in url
```
Doniaab, line 164: TODO: Should rework AA layers switch in url, currently not working correctly

Doniaab: switch between AA in url is now working but date query no longer appear in url after switch

frontend/src/components/MapView/DateSelector/index.tsx
```diff
@@ -51,8 +57,10 @@ const POINTER_ID = 'datePointerSelector';
 const calculateStartAndEndDates = (startDate: Date, selectedTab: string) => {
   const year =
     startDate.getFullYear() -
-    (selectedTab === 'anticipatory_action' && startDate.getMonth() < 3 ? 1 : 0);
-  const startMonth = selectedTab === 'anticipatory_action' ? 3 : 0; // April for anticipatory_action, January otherwise
+    (isAnticipatoryActionLayer(selectedTab) && startDate.getMonth() < 3
```
ericboucher, line 60: Actually I believe this will only apply for the drought AA, since we don't have the same concept of "windows 1 and 2".

frontend/src/components/MapView/LeftPanel/AnticipatoryActionPanel/DistrictView/index.test.tsx
```diff
@@ -3,7 +3,7 @@ import { Provider } from 'react-redux';
 import { BrowserRouter } from 'react-router-dom';
```
ericboucher, line 3: I think the views (District, Forecast, Hometable, Timeline...) will be different for AA storm so they should probably live under AA Drought? What do you think?

ericboucher: I guess we can move it later, I just want to be careful when copying components like this not to carry over extra complexity and avoidable dependence

frontend/src/components/MapView/DateSelector/index.tsx
Doniaab: Yes, i haven't yet distinguished the content (panel, timeline, etc.), that's what I'm going to do next. that's include logic for window 1 and 2 that I saw in certain parts.

frontend/src/components/MapView/LeftPanel/AnticipatoryActionPanel/DistrictView/index.test.tsx
Doniaab: yes, I have not yet touched on the distinction between the view elements specific to each AA. Also i wanted to test if everything was loaded correctly when i switch AA.

I'm going to see what can be put in common for the two AA types, perhaps extracting common parts. 
But I understand that the current view elements are specific to AA drought and must be separated from AA storm

Max-Z80: I have moved the AA drought specific components to AADroughtPanel folder

frontend/src/components/MapView/DateSelector/index.tsx
Max-Z80: The component `AnticipatoryActionDroughtPanel` has been cleaned

frontend/src/utils/keep-layer-utils.ts
```diff
@@ -37,7 +38,7 @@ export function keepLayer(layer: LayerType, newLayer: LayerType) {
     return false;
   }
 
-  if (newLayer.type === layer.type && layer.type === 'anticipatory_action') {
+  if (newLayer.type === layer.type && isAnticipatoryActionLayer(layer.type)) {
```
ericboucher, line 41: I believe this will only be needed for aa_drought

ericboucher (approved):

frontend/src/utils/keep-layer-utils.ts
Max-Z80: Yes, this is only relevant for the aa_drought because it handles the fact that aa_drought can display the layer of each 'window' simultanously on the map. There is no concept of window in the aa_storm layer, as far as I know. I will fix that.

frontend/src/components/MapView/LeftPanel/index.tsx
Doniaab: resolved

PR #1378, opened 2024-12-06
Fix TS analysis; COUNTRY=mozambique

### LINKED ISSUE #1380: [Bug]: population exposure analysis for tropical storms fails
### What happened?

Running a population exposure analysis for tropical storms no longer functions. 

Steps to reproduce:
1. country=mozambique
2. select layer: tropical storms
3. click the exposure analysis icon in the layer navigation panel on the left
4. see the error 'Cannot read properties of undefined (reading 'length')' 

From what I can tell, the error happens before a server request is made

mozambique on main branch

Screenshot 2024-12-06 at 16 42 12

ericboucher:
### Description

Fix #1380 

Make sure to run `yarn setup:common` locally to apply the changes before deploying.

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -96,9 +96,9 @@ jobs:
             **/node_modules
             **/.eslintcache
             ${{ steps.yarn-cache-dir-path.outputs.dir }}
-          key: ${{ runner.os }}-yarn-2024-7-${{ hashFiles('**/frontend/yarn.lock') }}-${{ hashFiles('**/common/yarn.lock') }}
+          key: ${{ runner.os }}-yarn-2024-12-${{ hashFiles('**/frontend/yarn.lock') }}-${{ hashFiles('**/common/yarn.lock') }}
           restore-keys: |
-            ${{ runner.os }}-yarn-2024-7
+            ${{ runner.os }}-yarn-2024-12
 
       - name: Install dependencies if needed.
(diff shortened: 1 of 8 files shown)


wadhwamatic:
Thanks @ericboucher. Pop exposure analysis now runs. I haven't looked at this in a while but am now seeing a few other issues:
1. After getting the results, the download report feature doesn't work (500 error from the API). This occurs locally and on a deployed version. If there's no quick fix, can we remove the report button for now?
2. I see some results that don't make sense. For example, use TS date 2019-04-25 and then run pop exposure analysis. Then sort the table results on the 120km/h column. Click on the first row (Cabo Delgado, Macomia, Macomia). The tooltip shows 'Sum: 11,688' and 'label: 90 km/h' while the table shows 11,688 in the 120 km/h column (screenshot below) 
3. Running an analysis through the analysis tab on TS layers doesn't give any results (empty table). Again, if no quick fix, can we hide TS layers from the analysis tab for now?

Screenshot 2024-12-07 at 18 17 54

ericboucher:
1. Fixed part of it but not completely, we really need to revamp this report thingy... Hid it for now and opened #1383 
2. investigating, probably because of the way the data get's aggregated
3. Seems to be working for me. Did you change the dates to have a storm that touches the country?

PR #1389, opened 2024-12-11
Anticipatory Action - Storm - COUNTRY=mozambique

ericboucher:


frontend/src/context/anticipatoryAction/AAStormStateSlice/utils.ts
```diff
@@ -0,0 +1,208 @@
...
+} from './parsedStormDataTypes';
+import { StormDataResponseBody, WindState } from './rawStormDataTypes';
+
+const districtNameMapping: { [key: string]: string } = {
+  Maganja_Da_Costa: 'Maganja Da Costa',
+  Cidade_Da_Beira: 'Cidade Da Beira',
+};
+
+const watchedDistricts: { [key in AACategory]: string[] } = {
+  [AACategory.Severe]: [
```
ericboucher, line 19: We should probably move away from the Severe/Moderate nomenclature used for AADrought since the meaning doesn't really match in this case.

PR #1405, opened 2025-01-09
remove useless aa storm state items

Max-Z80:
### Description

- clean anticipatory action state
- handle TODO - create a file to make the icons easier to access
- handle // TODO - setAAFilters is mostly used for the viewType. Maybe rename and make sure it's used for the viewType only?

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with

### DIFF
--- a/frontend/src/components/MapView/Layers/AnticipatoryActionStormLayer/index.tsx
+++ b/frontend/src/components/MapView/Layers/AnticipatoryActionStormLayer/index.tsx
@@ -309,12 +309,15 @@ const AnticipatoryActionStormLayer = React.memo(
       return null;
     }
 
+    // Create a report id based on the reference time of the report
+    const reportId = stormData.forecastDetails?.reference_time || '';
+
     return (
       <>
         {/* First render all fill layers */}
         {coloredDistrictsLayer && (
           <Source
-            key="storm-districts"
+            key={`storm-districts-${reportId}`}
             id="storm-districts"
             type="geojson"
             data={coloredDistrictsLayer}
@@ -342,7 +345,7 @@ const AnticipatoryActionStormLayer = React.memo(
         <>
           {stormData.activeDistricts?.Moderate?.polygon && (
             <Source
-              key="exposed-area-48kt"
+              key={`exposed-area-48kt-${reportId}`}
               type="geojson"
               data={stormData.activeDistricts?.Moderate?.polygon}
             >
@@ -361,7 +364,7 @@ const AnticipatoryActionStormLayer = React.memo(
           )}
           {stormData.activeDistricts?.Severe?.polygon && (
             <Source
-              key="exposed-area-64kt"
+              key={`exposed-area-64kt-${reportId}`}
               type="geojson"
               data={stormData.activeDistricts?.Severe?.polygon}
             >
@@ -383,7 +386,7 @@ const AnticipatoryActionStormLayer = React.memo(
         {/* Uncertainty cone */}
         {stormData.uncertaintyCone && (
           <Source
-            key="uncertainty-cone-map"
+            key={`uncertainty-cone-map-${reportId}`}
             type="geojson"
             data={stormData.uncertaintyCone}
           >
--- a/frontend/src/components/MapView/Layers/AnticipatoryActionStormLayer/index.tsx
+++ b/frontend/src/components/MapView/Layers/AnticipatoryActionStormLayer/index.tsx
@@ -15,7 +15,6 @@ import { LayerData } from 'context/layers/layer-data';
 import { getBoundaryLayersByAdminLevel } from 'config/utils';
 import {
   AADataSelector,
-  AAFiltersSelector,
   AALoadingSelector,
   loadStormReport,
 } from 'context/anticipatoryAction/AAStormStateSlice';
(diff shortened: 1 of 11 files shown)


Max-Z80:
@ericboucher j'ai clean le state, il reste quelques todo a traiter

frontend/src/components/MapView/Layers/AnticipatoryActionStormLayer/index.tsx
```diff
@@ -327,7 +322,7 @@ const AnticipatoryActionStormLayer = React.memo(
         {/* First render all fill layers */}
         {coloredDistrictsLayer && (
           <Source
-            key={`storm-districts${dateId}`}
+            key="storm-districts"
```
ericboucher, line 325: Why did we remove the dateId? It is used to make sure we are not displaying the wrong layers inadvertently. This could happen in some edge cases when switching between dates. You can use the date contained in the stormData instead as id

Max-Z80: it has been removed because it is always an empty string. `selectedDateTime` is always undefined. I have removed all the filters from the state because unused.
Ok, i will use the date you mentionned.

Max-Z80: fixed

Max-Z80:
@ericboucher : I am done with the handling of the todos. 
There is one last todo which is about handling the case where there are several storms on one day. I have asked gabriela and marc on figma about that. I suspect that it will take some time and may imply several modfications so I suggest to handle this point in a dedicated branch once the specs are mature. This enables to not block this PR.

frontend/src/components/MapView/Layers/AnticipatoryActionStormLayer/index.tsx
```diff
@@ -47,18 +37,19 @@ const boundaryLayer = getBoundaryLayersByAdminLevel(2);
 
 // Add this mapping object at the top of the file with other imports
 const WIND_TYPE_TO_ICON_MAP: Record<string, string> = {
```
ericboucher, line 39: This could also live in the icon file

frontend/src/components/MapView/Layers/AnticipatoryActionStormLayer/index.tsx
```diff
@@ -47,18 +37,19 @@ const boundaryLayer = getBoundaryLayersByAdminLevel(2);
...
-  'severe-tropical-storm': severeTropicalStorm,
-  'tropical-cyclone': tropicalCyclone,
-  'intense-tropical-cyclone': intenseTropicalCyclone,
-  'very-intensive-tropical-cyclone': veryIntensiveCyclone,
-  inland,
-  dissipating,
-  default: defaultIcon,
+  disturbance: anticipatoryActionIcons.disturbance,
+  'tropical-disturbance': anticipatoryActionIcons.disturbance,
+  low: anticipatoryActionIcons.disturbance,
```
ericboucher, line 42: need to check if this was not a mistake

ericboucher: Looks like this was done by @wadhwamatic so let's remove the commented lines. We can revisit colors later on

frontend/src/components/MapView/LeftPanel/AnticipatoryActionPanel/AAStormLegend/index.tsx
```diff
@@ -4,51 +4,42 @@ import {
...
-import tropicalCyclone from '../../../../../../public/images/anticipatory-action-storm/tropical-cyclone.png';
-import intenseTropicalCyclone from '../../../../../../public/images/anticipatory-action-storm/intense-tropical-cyclone.png';
-import veryIntensiveCyclone from '../../../../../../public/images/anticipatory-action-storm/very-intensive-tropical-cyclone.png';
 
 const phases = [
   {
-    icon: disturbance,
+    icon: anticipatoryActionIcons.disturbance,
     label: 'Weak low pressure system',
     speed: '< 51 km/h',
```
ericboucher, line 14: This could potentially be moved to the mapping as well, wdyt?

frontend/src/components/MapView/Layers/AnticipatoryActionStormLayer/index.tsx
Max-Z80: okay, done

frontend/src/components/MapView/LeftPanel/AnticipatoryActionPanel/AAStormLegend/index.tsx
Max-Z80: this is only used in this file. So I would not do it

frontend/src/components/MapView/Layers/AnticipatoryActionStormLayer/index.tsx
Max-Z80: could be. I did not find the `low` occurence looking at few reports. I have found all others occurences.  But according to Val it can happen https://ovio-public.slack.com/archives/C01PFEJ0NJD/p1734174315413249?thread_ts=1734172468.085679&cid=C01PFEJ0NJD

wadhwamatic: Hi yes that was done quickly by me. I also didn't see low. I think we can keep default as it was as I've replaced the icon for it which should cover the need here

ericboucher: Ok let's remove the comments for now then to clean things up. And then this should be ready to merge

ericboucher (approved):

PR #1406, opened 2025-01-09
AA Storm map screenshot for alert mail

Doniaab:
### Description

This fixes #issueNumber.

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with

### DIFF
--- a/frontend/src/components/MapView/Layers/AnticipatoryActionStormLayer/index.tsx
+++ b/frontend/src/components/MapView/Layers/AnticipatoryActionStormLayer/index.tsx
@@ -23,6 +23,7 @@ import { AACategory } from 'context/anticipatoryAction/AAStormStateSlice/types';
 import { updateDateRange } from 'context/mapStateSlice';
 import { useWindStatesByTime } from 'components/MapView/DateSelector/TimelineItems/hooks';
 import { getAAColor } from 'components/MapView/LeftPanel/AnticipatoryActionPanel/AnticipatoryActionStormPanel/utils';
+import { sendMailWithImage } from 'context/anticipatoryAction/AAStormStateSlice/utils';
 import AAStormDatePopup from './AAStormDatePopup';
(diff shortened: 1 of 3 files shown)


ericboucher:
@Doniaab can you run me through the architecture you have in mind here?
Let's plan things ahead a bit to make sure the functions we develop match. For example, I am not sure the emails would be sent by the frontend, most likely they would be sent from a backend process. Either a version of the one we already have for alerting or a new one.

Doniaab:
@ericboucher
I have in mind that we need to use the alert API to send the email (or do something similar specific to storm alerts). The comments refer to the API call for sending the email. For now, I’ve been focusing on capturing the map with the correct layers.

I don’t fully understand the specificities of the current alert handler, but I think we could specify an alert type (either in the body of the request or as a route parameter, e.g., alerts/:type -> alerts/storm). Based on this alert type, the backend could parse the data structure and inject it into an appropriate email template.

The alerts are stored in the database, so we need to ensure that we can store the storm-specific data without causing inconsistencies with the existing model. I’m currently reviewing this aspect. 

Also i now understand that we can generate a WMS URL from the map with the appropriate layers, which the backend can use to retrieve the map image dynamically.

ericboucher:
Ok, I would recommend writing a mini tech spec to clarify what you need and what we are trying to achieve.
For example, triggering alert emails cannot be the responsibility of the frontend since it needs to happen without any user interaction

Doniaab:
1. Cron Job
Create a new cron job for storm alerts.
The cron job will run the alert-worker with a type parameter "storm".

2. Alert Worker
In the alert-worker, check if the type is "storm".
If the type is "storm", fetch the latest data to check for new activated districts.
If new activated districts are found, create a new alert in the database.

3. Email
Create an email template and inject the districts activation data.
Capture an image of the map with the activated districts using WMS GetMap.

PR #1426, opened 2025-01-31
Fix composite input layer invert bug; COUNTRY=jordan

wadhwamatic:
### Description

This fixes https://github.com/WFP-VAM/prism-app/issues/1425

## How to test the feature:

- [ ] check CDI config in layers.json, look for invert attribute of input layer
- [ ] load the same CDI layer in prism
- [ ] verify that the correct invert attribute is sent to hip-analysis

## Checklist - did you ...

Test your changes with

### DIFF
--- a/frontend/src/context/layers/composite_data.ts
+++ b/frontend/src/context/layers/composite_data.ts
@@ -84,7 +84,7 @@ export const fetchCompositeLayerData: LazyLoader<CompositeLayerProps> =
         key,
         aggregation,
         importance,
-        invert: Boolean(invert),
+        invert,
       })),
       ...aggregationMaskParam,
       ...aggregateByParam,


frontend/src/context/layers/composite_data.ts
```diff
@@ -84,7 +84,7 @@ export const fetchCompositeLayerData: LazyLoader<CompositeLayerProps> =
         key,
         aggregation,
         importance,
-        invert: Boolean(invert),
```
ericboucher, line 87: This "should" be doing the same thing. Can you explain the bug you were seeing?

ericboucher: The error is actually on the JSON side.

"invert": "True" should be "invert": true

And

"invert": "False" should be "invert": false

If you prefer your option and it works well, that's ok, we would simply need to update the type:

    invert?: boolean | undefined; ->    invert?: 'False' | 'True' | undefined;

wadhwamatic: ahh thanks @ericboucher. fixing the error in the config is definitely better. will close this PR

PR #1430, opened 2025-02-05
Use backend report status directly for "readiness" and "activation"

Doniaab:
### Description

This PR is a change to use the report status to display readiness in the left panel and activated districts.
What this means is that if the status is ready, even though some districts are listed as activated, we don't show them as active since the landfall lead time is above 72 hours.

In the future we might display activated regions in that case, when they will be available from the backend.

image

## How to test the feature:

- You can go to this url and select 12pm at this date on the timeline : https://prism-1430.surge.sh/?hazardLayerIds=anticipatory_action_storm&date=2025-02-03
- The report at this date and time is in status "ready" and one watched district "Maganja Da Costa" is in the list exposed_area_48kt: We shouldn't see this district in the activated list on the left panel because the status is "ready".
- We should display readiness on the left panel.

## Checklist - did you ...

Test your changes with


frontend/src/context/anticipatoryAction/AAStormStateSlice/utils.ts
```diff
@@ -104,53 +104,80 @@ export function parseAndTransformAA(data: StormDataResponseBody): ResultType {
...
+
+  // Process the active and inactive districts
+  const [activeDistricts, naDistricts] = (
+    Object.values(AACategoryKey) as AACategoryKey[]
+  ).reduce(
+    ([activeResult, naResult], categoryKey) => {
+      const category = AACategoryKeyToCategoryMap[categoryKey];
+
+      // If the storm status is not active, all watched districts should be marked as inactive
+      if (!isActivated) {
```
ericboucher, line 124: Is this check necessary? won't the sets be empty anyway further down?

Doniaab: Yes, this check is necessary because if isActivated is false, we need all watched districts to be explicitly added to naDistricts. Otherwise, they wouldn't be included anywhere.

Also we can be in ready status and have watched districts in the affected districts list. In this case the status already take in account the 'landfall leaditme > 72 hours' and we dont have to put those district in the activated list.

ericboucher:
Ok, is this ready for review by @wadhwamatic ?

frontend/src/context/anticipatoryAction/AAStormStateSlice/utils.ts
Max-Z80: I have a doubt about whether we should display districts in the NA>XXX boxes in the leftPanel when the report status is `monitoring` or `ready` because on figma, there is no district displayed when the status is ready / monitoring
image

Since, the spec have changed a bit in the meantime (and the figma template did not follow on changes), it could well be that the current implementation is as expected (ie display NAxxx boxes when status is monitoring/ready)

frontend/src/context/anticipatoryAction/AAStormStateSlice/utils.ts
```diff
@@ -104,53 +104,80 @@ export function parseAndTransformAA(data: StormDataResponseBody): ResultType {
...
-                ...naResult,
-                [category]: {
-                  districtNames: notActive,
-                  polygon: {},
-                },
-              },
-            ];
-          }
-          return [activeResult, naResult];
+  // Determine if watched districts are active based on storm status
```
Max-Z80, line 107: It would be nice to add a test on parseAndTransformAA

frontend/src/context/anticipatoryAction/AAStormStateSlice/utils.ts
ericboucher: @wadhwamatic any preference on this?

ericboucher: I'm merging like that for now but we can easily change that behavior later

PR #1437, opened 2025-02-19
Add pmtile support for analysis server; COUNTRY=global

gislawill:
### Description

This PR adds the ability to reference S3 stored parquets in our analysis feature. It works by adding a function to our zonal_stats module that either loads geojson data directly (standard path) or uses DuckDB to create a temporary geojson representation. 

We then take the zonal stats stats response and create a new layer (powered by the same PMTiles) that renders our color coded analysis map.

This work currently doesn't support running analysis for admin level 2 (aka county-level) due to the performance of our zonal stats implementation. Fixing this will require a fairly large refactor of our zonal stats. 

## Known Issues (outside scope of this PR)
- ~App start up is incredibly slow — I'm looking into why that is~
    - This was addressed by refactoring in [frontend/src/components/Common/BoundaryDropdown/utils.ts](https://github.com/WFP-VAM/prism-app/pull/1437/files#diff-c30263f71eb62b8f99853b86b66ba76b99fb00404298106a4262953e7d39004f)
- Initial viewport is very wide to allow us run zonal stats on all regions — I'm looking into ways of separating this
- Our admin bounds are cut off at 50°N and 50°S — we just need new regions to get around that

## How to test the feature:
- [ ] Load the global deployment
- [ ] Select the analysis layer
- [ ] Run an analysis 

Test your changes with

Screenshot 2025-04-15 at 6 23 42 PM


api/app/zonal_stats.py
```diff
@@ -68,22 +69,89 @@ def get_wfs_response(wfs_params: WfsParamsModel) -> WfsResponse:
...
 
 
-def _extract_features_properties(zones_filename: FilePath) -> list:
-    with open(zones_filename) as json_file:
-        zones = load(json_file)
+def _read_zones(
+    zones_filepath: FilePath,
+    admin_level: Optional[int] = None,
+    bbox: Optional[tuple[float, float, float, float]] = None,
+) -> GeoJSON:
```
gislawill, line 76: This is the core of the update: a method that selectively reads from a cached geojson (geojson-path) or uses duckdb to read from a parquet and generate a geojson dictionary of the admin level data

api/app/zonal_stats.py
```diff
@@ -68,22 +69,86 @@ def get_wfs_response(wfs_params: WfsParamsModel) -> WfsResponse:
...
+            WITH (FORMAT GDAL, DRIVER 'GeoJSON', LAYER_CREATION_OPTIONS 'WRITE_BBOX=YES')
+            """
+        )
+
+        with open(temp_geojson, "r") as f:
+            geojson_data = load(f)
+
+        con.close()
+        os.remove(temp_geojson)
+        return geojson_data
```
gislawill, line 126: Here's the interesting part, we use duckdb to selectively query the parquet and convert to geojson so we don't need to change the rest of the function

gislawill: Not the most efficient but avoids a full rewrite of zonal stats

frontend/src/context/analysisResultStateSlice.ts
```diff
@@ -717,6 +722,8 @@ export const requestAndStoreAnalysis = createAsyncThunk<
   const loadedAndCheckedBaselineData: BaselineLayerData =
     await getCheckedBaselineData();
 
+  // TODO: This merges with PMTiles data, not GeoJSON and breaks the flow.
+  // Bypassing works... is it needed?
   const features = generateFeaturesFromApiData(
```
gislawill, line 727: @ericboucher, I'm curious for your thoughts here. The current implement looks to be phasing out `geojson_out` but it's actually super useful for pmtiles layers. In this implement we're using geojson_out and bypassing the merge with admin boundary steps. Obviously this needs logic to maintain backwards compatibility but, before then, I'm curious for your thoughts on this approach? Any qualms with relying on geojson_out?

ericboucher: The main concern with using geojson_out is sending big files in/out of the app. When we are looking at admin 3 or a global deployment this can end up being a huge file.

That's part of the reason why the frontend is sending a link to a geojson file and not posting the shape each time.

Does this help?

gislawill: Yea, I hear you there. We're already disabling the ability use the analysis layer for admin level 2 & 3 in the global deployment. Admin level 2 is actually pretty functional, just kind of slow (30 seconds?). By disabling these levels and using `ST_,simplify` to reduce geometry size, I don't think we need to be too worried about file size for the time being.

frontend/src/context/analysisResultStateSlice.ts
```diff
@@ -436,11 +439,13 @@ async function createAPIRequestParams(
...
     ...maskParams,
     // TODO - remove the need for the geojson_out parameters. See TODO in zonal_stats.py.
+    // TODO - Add back logic
     geojson_out: Boolean(geojsonOut),
     intersect_comparison:
       exposureValue?.operator && exposureValue.value
         ? `${exposureValue?.operator}${exposureValue?.value}`
         : undefined,
+    adminLevel,
   };
```
gislawill, line 449: TODO: Allow passing of simplify tolerance

frontend/src/config/global/prism.json
```diff
@@ -14,7 +14,7 @@
     "wms": ["https://api.earthobservation.vam.wfp.org/ows/wms"]
   },
   "map": {
-    "boundingBox": [-100, -35, 150, 50]
+    "boundingBox": [-130, -50, 177, 50]
```
gislawill, line 17: @ericboucher — this effectively controls both the initial viewport and the area with which we will run analysis on. I'll separate these in a later PR

frontend/src/components/MapView/Layers/AnalysisLayer/index.tsx
```diff
@@ -70,31 +70,70 @@ const onClick =
...
+      }) as Record<
+        string,
+        { data: string | number | null; coordinates: GeoJSON.Position }
+      >;
+
+    const layerTitle = (
+      analysisData as BaselineLayerResult | ExposedPopulationResult
+    ).getLayerTitle(t);
+
+    if (
```
ericboucher, line 101: Add a comment here for consistency maybe?

ericboucher (approved):

PR #1494, opened 2025-07-01
AA Drought - centralize config

ericboucher:
### Description

This fixes #issueNumber.

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with


frontend/src/config/malawi/index.ts
```diff
@@ -3,7 +3,14 @@ import rawLayers from './layers.json';
 
 const rawTables = {};
 const rawReports = {};
-const translation = {};
+
+// Temp: hack for Malawi window labels
+const translation = {
+  en: {
+    'Window 1': 'NDJ',
```
ericboucher, line 10: @wadhwamatic can we anticipate having more than 2 windows? or just 1?

wadhwamatic: I don't expect more than two windows, but a single monitoring window is definitely a possibility. Let me double check with Gabriela.

wadhwamatic: @ericboucher - confirmed that we will have a single monitoring window in some countries. No plans for more than 2, and doubtful we would in the future

wadhwamatic:
This looks good from my side @ericboucher. Thanks for the chart fix too.

frontend/src/config/malawi/index.ts
ericboucher: Ok, let's wait until this happens to adapt

wadhwamatic:
@ericboucher - Gabriela noted an issue here.

4 districts are in set for JFM while still also listed as ready for the same period. If we get to set stage for a specific period e.g. JFM, that is what we should display. 

image

ericboucher:
> @ericboucher - the `seasonStartMonth` for Malawi is correctly set to July. We have new forecast data in the CSV file ([data.earthobservation.vam.wfp.org/public-share/aa/aa_probabilities_triggers_mwi.csv](https://data.earthobservation.vam.wfp.org/public-share/aa/aa_probabilities_triggers_mwi.csv)) covering July but when you load the AA module, it advances to Nov 2024 by default. You can get to July 2025 in the timeline / calendar. I tried changing the `seasonStartMonth` to June, but that didn't affect the behavior

Nice find, you uncovered a bug where we didn't reorder the combined AA dates and assumed window2 would have the latest date. Which is not the case of course at the beginning of the season.

Note that seasonStartMonth is not used for that. seasonStartMonth is used to offset the date selector visually and to reset the "set" states between years as we can't use the actual year YYYY

Issue #1499, opened 2025-07-09
[Bug]: schemathesis tests were skipped and silently failling


laurentS:
### What happened?

As part of the python api dependencies upgrade, I removed the code for `schemathesis` tests, to limit loading time overhead.

I think a `uvx schemathesis run http://localhost:80/openapi.json` (after a `make api`) is actually more useful and would be worth running every now and then, maybe as a github scheduled job or maybe CI if an extra 3-4 minutes of runtime isn't an issue. It also does not require any dependencies or specific code, which makes maintenance much easier.

A quick run on my machine found 28 failures in about 3 minutes. This ticket is about fixing these issues which cover a variety of problems such as:
- Undocumented HTTP status code
- API rejected schema-compliant request
- Server error (500) on a valid request (valid as per the openapi schema we produce)

### Which country / deployment are you running?

any

PR #1500, opened 2025-07-09
WIP: Upgrade rasterstats, fiona, fastapi...

laurentS:
### Description

This PR upgrades some python dependencies:
- rasterstats
- fiona
- fastapi
- pytest

`vcrpy` is pinned to `6.0.1` because something breaks when updating to `7.0.0`.

I took the liberty to remove the `schemathesis` test code as it was skipped by pytest anyway, and it started adding massive startup time overhead (3-4 minutes!) on the test suite. Instead see #1499 

TODO: 4 of the tests are still broken and need fixing. I don't understand the functional logic enough to have an immediate fix for the problem. @ericboucher if you have any ideas about why the `test_calculate_stats_*` suddenly break. 

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with


ericboucher:
What worries me is that the failing tests suggest that the stats function got broken when updating fiona

laurentS:
> What worries me is that the failing tests suggest that the stats function gets broken when updating fiona

Yes, upon further investigation, I think we hit [this bug](https://github.com/Toblerity/Fiona/issues/1455). I've just pushed a fix which I don't find very nice, but it seems to get tests to :green_circle: I'm not sure if there are cases where we would pass an actual geojson string to `calculate_stats`. The type annotations seem to suggest that it's not a problem...

api/pyproject.toml
```diff
@@ -4,42 +4,47 @@ version = "1.1.1"
...
 psycopg2 = "^2.9.9"
-rasterstats = "0.19.0"
+rasterstats = "0.20.0"
 boto3 = "^1.28.61"
 cachetools = "^5.3.1"
 odc-geo = "^0.4.1"
-odc-stac = "^0.3.11"
-dask = "2024.10.0" # temporary fix for https://github.com/opendatacube/odc-stac/issues/199
+odc-stac = "^0.4.0"
+dask = "^2025.5.0"
```
ericboucher, line 27: We might not need to set this explicitely anymore based on the comments you removed?

ericboucher (approved):

PR #1501, opened 2025-07-10
Add ability to disableDownload for layers

### LINKED ISSUE #1482: [Feature Request]: make layer data download optional

Some government partners have requested that we disable the option to download data from PRISM.

I'd like to have this as an optional flag within layers.json. By default, the layer should be downloadable, but a flag, something like 'download=false' would disable it. 

_No response_

VishalV1807:
### Description

This fixes #1482.

- This solution adds a download flag to all the layers of layers.json of each country. 
- The default value of the flag is true. To change the flag to false, run an add_download_flag.py script in the root directory `python frontend/scripts/add_download_flag.py frontend/src/config` and to toggle it to true, run `python frontend/scripts/add_download_flag.py frontend/src/config --default-true`
- It also removes the download button if the flag is set to false
## How to test the feature:

- [ ] Run the script mentioned above and set the download flag to false
- [ ] Do the download buttons appear for any layer?
- [ ] Set the flag to true again
- [ ] Do the download buttons reappear?

## Checklist - did you ...

Test your changes with

# Flag set to false. Country: mozambique

image

# Flag set to true. Country: mozambique

image

### DIFF
--- a/frontend/src/config/somalia/layers.json
+++ b/frontend/src/config/somalia/layers.json
@@ -422,7 +422,6 @@
     "legend_text": "This dataset contains population estimates generated through the 2014 Population Estimation Survey (PESS), conducted by the Somalia National Bureau of Statistics (SNBS) with support from the United Nations and international partners. Siyrce: PESS 2014"
   },
   "pop_2024": {
-    "disableDownload": true,
     "title": "Population (UN 2024)",
     "type": "admin_level_data",
     "boundary": "admin2_boundaries",


VishalV1807:
That makes sense, I will only add the flag when downloads have to be disabled. But, should I add this flag to each layer of every layers.json file? Or would you want this in the form of a 'global' flag that can be accessed to check if downloading is allowed or not?

wadhwamatic:
> For clarity, let's use `disableDownload` as the flag name. And only add it to one Mozambique layer for testing purposes for now.
> 
> @wadhwamatic do you already have layer/country pairs you would like to disable?

The request came from Somalia but it wasn't for a specific layer yet. For now, let's apply it to the `pop_2024` layer in Somalia and I will demonstrate this to them

ericboucher:
@VishalV1807 will you be able to make the change?

VishalV1807:
@ericboucher @wadhwamatic I made the changes. The Somalia pop_2024 layer is now not downloadable when disabledownload flag is set to true. Here is the screenshot for the same:

image

ericboucher (approved):

PR #1506, opened 2025-07-16
Speed up app loading by delaying computation of available dates

### LINKED ISSUE #1461: [Bug]: PRISM is slow to load when we have multiple layers with validity ranges
### What happened?

I created PR https://github.com/WFP-VAM/prism-app/pull/1457 to show our team in Mozambique how the validity date feature can clarify date ranges for selected indicators. In the PR, I added validity ranges for all relevant indicators in shared/layers.json as well as locally relevant layers for Mozambique.

The initial load time is significantly longer than usual after that change, both locally and when I deployed it on Amplify https://dev2.dkr9o3t8h8nvn.amplifyapp.com/

I know the available date array is heavy with this approach. Any quick wins you can think of to improve this? 

Mozambique / https://github.com/WFP-VAM/prism-app/pull/1457 

_No response_

laurentS:
## Description

This will fix #1461 (PR on top of #1457)

This PR changes the way `serverState` is computed, instead of building the entire state slice when the app starts (and computing available dates for all layers in one go), it delays this calculation to layer activation (via the switch in the left toolbar).

When profiling the original code, most of the time was spent waiting for network requests. These have been moved to run in the background and in parallel when the app starts. No rendering should be conditional on the results anymore (except of course if a layer is activated before data is received, ie. in the first few seconds after the app becomes responsive, or via the url). The data loaded from the network is not processed, and is stored as-is in a new state slice (`serverPreloadState`).

When a layer is activated, the rest of the calculation is done, based on preloaded data. This should happen within <300ms or so (the slowest layer seems to be SPI rainfall 5 years).

To fix some typing issues with redux thunks, I ended up upgrading `react-redux`, which is probably a good thing, but not sure if doing it as part of this PR is ideal...

~~Note: there is still a bug with (at least) SPI layers:~~
~~- open the app for Mozambique `REACT_APP_COUNTRY=mozambique yarn start`~~
~~- activate SPI with the toggle (Rainfall/ INAM rainfall data / SPI). This turns on the default SPI 1 month layer.~~
~~- switch to another SPI layer, eg. 6 months. The date calculation function is now run more than 10 times for some obscure reason.~~
~~- ultimately, the dates appear to load correctly, but they take an order of magnitude longer than needed.~~

While trying to debug this, I noticed a variety of `useMemo` functions that appear to be run dozens of times, despite the fact that their inputs do not change values. `redux-dev-tools` are so sluggish that they're unusable given the size of the state tree. I will take a fresh look tomorrow, but any ideas about the cause of these multiple calculations (or how to investigate) are welcome!

## How to test the feature:

I've only tested with Mozambique so far, and only very basically, but it should work for any country.
- [ ] start the app, notice that it starts much faster than before
- [ ] activate some layers, this should now be fast enough not to be a problem
- [ ] hopefully functionality is not affected

## Checklist - did you ...

not yet

Test your changes with


frontend/src/components/MapView/LeftPanel/layersPanel/MenuItem/MenuSwitch/SwitchItem/index.tsx
```diff
@@ -48,6 +51,7 @@ const SwitchItem = memo(
     const { t } = useSafeTranslation();
     const selectedLayers = useAppSelector(layersSelector);
     const serverAvailableDates = useAppSelector(availableDatesSelector);
+    const layersLoadingDates = useAppSelector(layersLoading);
```
ericboucher, line 54: The name sounds a bit weird, can you describe what its' holding?

wadhwamatic:
Thanks @laurentS for all of the work on this! I tested the firebase auto-deployment for Mozambique and the performance is very good. 

I tried using Chrome's device simulation settings and throttled CPU 3.3x. When I did that, the metrics look pretty good, but **boundaries did not load**.  Then, if I try to toggle between a couple of rainfall layers, I see Interaction to Next Paint (INP) 'needs improvement' at 440ms. 

In general, some loading indication on the map and timeline would be helpful, especially for slower machines. 

I'd also like to test this with date intersections between admin level and wms layers, when country=RBD. If you or Eric have time to deploy that, that would be appreciated.

laurentS:
I've updated the PR with:
- removing the gray overlay so it's not there at all anymore, as it was not actually used at the start of the app. If you think it's better with it during initial loading, I'll bring it back
- `LoadingBar` now spins while dates are being calculated, as you requested
- there was a bug mentioned in the PR description which is gone. This substantially speeds up switching between layer variants (like `SPI` 1 month to SPI 5 years), because the calcultion is done only once instead of 14 times (I have not figured out why 14).
- my last commit adds caching for `preprocessed-layer-dates.json`, which seems redundant, but appears to shave off 70-100ms on layer activation.
- I did a bit of profiling again with the new code around layer activations. The slowest layer I've found for Mozambique is `spi_blended_5y` and the date calculation for that one takes around 300ms. The UI gets frozen for a little more than that, but most of it seems to be the map itself loading tiles and such (the flame graph dives into `Maplibre...` and `reactDom` functions). I changed the final `sortBy` call for a faster native one, this saves anywhere between 0 and 50-60ms per layer for Mozambique (more on slower machines).

Regarding @wadhwamatic's questions above, I have not been able to reproduce the problem with boundaries not loading. Do you have a step-by-step that leads to it, so I can investigate further?

The INP seems a lot longer than the time spent on calculating dates. I find it hard to understand where time is spent in detail, but it looks like it's react updating the DOM.

One question: this was not modified by this PR, I checked with the base branch, but there are about 109k `DateItem`s coming out of `generateIntermediateDateItemFromValidity` for `spi_blended_5y`. At the end of that function, there is a comment about deduplicating, but that seems to be gone in the code itself (in the sense that there are more than 30 items with the same `displayDate`, but the other members of the `DateItem` are different). Is that number of items expected, or is there something that got lost in the code?

laurentS:
There was another bug with Analysis where selecting a layer that had not previously been activated in the main view left it with no dates. This is fixed.

Are there any other parts of the app that I missed?

If not, I think this PR is pretty much ready to go. Sorry it took so long, async work is always a bit messy :face_with_spiral_eyes:

ericboucher:
This looks good to me, one small UX issue remaining : the first layer we activate should not complain about not having data available for “today” since technically no date has been set yet. It does not complain on `master` for example.

wadhwamatic:
Thanks @laurentS and @ericboucher. I want to spend some more time testing this one to make sure we've got all of the date cases covered. I'll follow up early next week.

frontend/src/config/index.ts
```diff
@@ -87,6 +87,10 @@ const safeCountry =
     ? (COUNTRY.toLocaleLowerCase() as Country)
     : DEFAULT;
 
+// list countries that have a preprocessed-layer-dates.json file
+// to avoid a failed network call on each layer activation
+const countriesWithPreprocessedDates = ['cameroon', 'nigeria', 'rbd'];
```
ericboucher, line 92: I like the idea, but is there a way to be smarter about that? And maybe create this list dynamically? Is theere a flag in prism.json we could use?

laurentS: yep, good call. I've added a few lines in the script that produces the preprocessed dates to also generate that list, so they should be in sync without extra action.

PR #1507, opened 2025-07-17
Dashboard shell

conor-jh:
### Description

This sets up some foundational pieces of the dashboard/reports* feature. Highlights:

- Add `ConfiguredReport` and associated types
- Display Dashboard in nav bar if it exists for a given deployment config
- Allow a user to choose a configured report

<details><summary>*<strong>Dashboard context</strong></summary>
<p>
Just a bit of background on the dashboard/reports feature…

This is the very beginning phase of a larger effort to provide configurable reports to Prism users. Deployment configs can specify a “dashboard” comprising a map and one or more content blocks (e.g. text, a chart, or a table). This inaugural PR is to specify the overall structure of the config shape and initialize a navigable placeholder dashboard.

</p>
</details> 

I’ve left a few inline comments throughout to provide context and pose a few questions concerning next steps.

## How to test the feature:

1. Start the Mozambique deployment
2. Observe “Dashboard” in the main navigation
3. Select “Dashboard” and the single report it contains
4. Observe the dashboard placeholder content

### Additional testing notes
* The route will update as users navigate to and from the dashboard tab — try navigating to the dashboard view directly via your browser’s address bar and observe the route changing when selecting dashboard vs. non-dashboard tabs
* If you change the config title, you should see that reflected in the UI

## Checklist - did you ...

<!-- If any of the following items aren't relevant to your contribution,
     please still tick them so we know you've gone through the checklist. →

Test your changes with

### DIFF
--- a/frontend/src/components/NavBar/index.tsx
+++ b/frontend/src/components/NavBar/index.tsx
@@ -23,6 +23,7 @@ import {
   TableChartOutlined,
   TimerOutlined,
   Notifications,
+  SpeedOutlined,
 } from '@material-ui/icons';
 import { useDispatch, useSelector } from 'react-redux';
 import {
@@ -31,7 +32,11 @@ import {
 } from 'context/leftPanelStateSlice';
 import GoToBoundaryDropdown from 'components/Common/BoundaryDropdown/goto';
 import Legends from 'components/MapView/Legends';
-import { areChartLayersAvailable } from 'config/utils';
+import {
+  areChartLayersAvailable,
+  areDashboardsAvailable,
+  getConfiguredReports,
+} from 'config/utils';
 import {
   areTablesAvailable,
   isAnticipatoryActionDroughtAvailable,
@@ -46,10 +51,30 @@ import PanelButton from './PanelButton';
 
 const { alertFormActive, header } = appConfig;
 
+const getAvailableDashboards = (): PanelItem[] => {
+  const configuredReports = getConfiguredReports();
+  return configuredReports.map((report, index) => ({
+    panel: Panel.Dashboard,
+    label: report.title,
+    icon: <SpeedOutlined />,
+    reportIndex: index,
+  }));
+};
+
 const panels: PanelItem[] = [
   { panel: Panel.Layers, label: 'Layers', icon: <LayersOutlined /> },
   ...(areChartLayersAvailable
     ? [{ panel: Panel.Charts, label: 'Charts', icon: <BarChartOutlined /> }]
+    : []),
+  ...(areDashboardsAvailable()
+    ? [
+        {
+          panel: Panel.Dashboard,
+          label: 'Dashboard',
+          icon: <SpeedOutlined />,
+          children: getAvailableDashboards(),
+        },
+      ]
     : []),
   {
     panel: Panel.Analysis,
@@ -147,7 +172,11 @@ function NavBar() {
       [panel.label]: child,
     });
     handleMenuClose(panel.label);
-    handlePanelClick(child.panel);
+    if (panel.panel === Panel.Dashboard && child.reportIndex !== undefined) {
+      handlePanelClick(Panel.Dashboard);
+    } else {
+      handlePanelClick(child.panel);
+    }
   };
 
   const { title, subtitle, logo } = header || {
@@ -194,9 +223,10 @@ function NavBar() {
                   (panel.children &&
                     panel.children.some(child => tabValue === child.panel));
 
-                const buttonText = selectedChild[panel.label]
-                  ? selectedChild[panel.label].label
-                  : t(panel.label);
+                const buttonText =
+                  selectedChild[panel.label] && panel.panel !== Panel.Dashboard
+                    ? selectedChild[panel.label].label
+                    : t(panel.label);
 
                 return (
                   <React.Fragment key={panel.panel}>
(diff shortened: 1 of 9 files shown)


frontend/src/components/NavBar/index.tsx
```diff
@@ -46,11 +51,31 @@ import PanelButton from './PanelButton';
 
 const { alertFormActive, header } = appConfig;
 
+const getAvailableDashboards = (): PanelItem[] => {
+  const configuredReports = getConfiguredReports();
+  return configuredReports.map((report, index) => ({
+    panel: Panel.Dashboard,
+    label: report.title,
+    icon: <SpeedOutlined />,
```
conor-jh, line 59: Looks like the icon in the Figma mockups isn’t available in this version of Material icons

frontend/src/config/types.ts
```diff
@@ -998,3 +1000,52 @@ export class AnticipatoryActionLayerProps extends CommonLayerProps {
...
+export enum DashboardElementType {
+  CHART = 'CHART',
+  TEXT = 'TEXT',
+  TABLE = 'TABLE',
+}
+
+export interface DashboardChartConfig {
+  type: DashboardElementType.CHART;
+  startDate: string;
+  endDate: string;
```
conor-jh, line 1018: Typing dates as strings as I imagine that’s how they’ll have to come in via JSON configs — it’ll probably be good to document a particular format unless there’s already an established pattern I’m not aware of.

conor-jh (commented):
One other big thing—you’ll notice the view doesn’t really update after selecting the Dashboard panel. I’m curious if there’s any feedback about how we want to set up the initial layout view.

`<MapView />` is always rendered in the main app, so we could modify it to support dashboards… but it almost feels like it could be cleaner to create a new view altogether.

Then we could render `<MapView />` OR `<DashboardView />` depending on the selected panel, giving us a clean slate for a “bounded” map component and additional panel elements. Thoughts?

frontend/src/config/types.ts
ericboucher: We usually go for YYYY-MM-DD

lowriech:
Couple comments, generally good start. Quick UI thing here:
https://www.loom.com/share/8c414ca5a0a94175a2a0885601db4577

frontend/src/config/types.ts
```diff
@@ -998,3 +1000,52 @@ export class AnticipatoryActionLayerProps extends CommonLayerProps {
...
+  hazardLayerId: string;
+  baselineLayerId: string;
+  stat: AggregationOperations;
+}
+
+export interface ConfiguredReport {
+  title: string;
+  selectedDate: string;
```
gislawill, line 1041: Let's make `selectedDate` optional — the majority will be "latest available dates" (aka selectedDate = undefined)

frontend/src/config/types.ts
gislawill: String here sounds fine to me. I suppose we could do something fancy like:
```
type TwoDigit = `${number}${number}`;
type ISODateString = `${number}${TwoDigit}-${TwoDigit}-${TwoDigit}`;

const valid: ISODateString = "2025-07-18"; // passes
const invalid: ISODateString = "July 18, 2025"; // type error
```
But that'd probably be overkill and would lend more confidence than it truly should (there's no runtime validation for these types and no static validation either — the joys and pains of JSON). I suppose we _could_ add runtime validation format for date configurations but the benefits seem too small to be worth it imo

gislawill (commented):
Looking good so far! I think we should use this PR to include a new route for dashboards in frontend/src/components/App/index.tsx and the routing functionality to get to and from the dashboard. The dashboard itself can be blank. 

If you'd prefer to handle that as a follow-up PR, that's fine with me too.

gislawill:
Regarding the UI note from @lowriech's loom
> Quick UI thing here

Consider enforcing similar behavior to the anticipatory actions actions. If we update one, we should update both. When in doubt, we can give options in the slack and get thoughts from Eric, Amit, and Marc

conor-jh:
> Couple comments, generally good start. Quick UI thing here: https://www.loom.com/share/8c414ca5a0a94175a2a0885601db4577

Think I’ve addressed this—as for the menu position, I played around with the `anchorPosition` props but there seem to be some issues… not sure if it’s due to our MUI version or something else but it feels more appropriate to address any concerns on that separately (the anticipatory action dropdown has the same issue).

gislawill:
One nit but otherwise looking good to me. Thanks Conor! Once we have passing CI, let's open for review.

frontend/src/components/NavBar/index.tsx
```diff
@@ -140,14 +177,25 @@ function NavBar() {
 
   const handlePanelClick = (panel: Panel) => {
     dispatch(setTabValue(panel));
+    if (panel === Panel.Dashboard) {
+      history.push('/dashboard');
+    } else if (location.pathname !== '/') {
+      history.push('/');
+    }
   };
```
ericboucher, line 185: Can we think of a more future proof approach? What happens if we have another page? eg. /alerts or /batch-print ?

frontend/src/config/mozambique/prism.json
```diff
@@ -464,5 +464,15 @@
...
-  }
+  },
+  "configuredReports": [
+    {
+      "title": "Population Exposure",
+      "selectedDate": "2025-01-01",
+      "mapPosition": "left",
+      "minMapBounds": [31, -25, 40, -11],
+      "mapLayers": [],
+      "flexElements": []
```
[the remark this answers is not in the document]
conor-jh: Yes, I’m very open to that—I think the best shape for the config will naturally evolve as we actually build out the layout engine. I don’t anticipate this is the final state.

PR #1522, opened 2025-08-12
Setup e2e testing with cypress (replaced)

laurentS:
This PR is replaced by #1524 but is kept for the history of changes and discussions.

### Description

This PR builds on #1506 (keeping it separate until we're happy with the attempt in here).

This sets up testing using [cypress](https://cypress.io). We had a quick chat with @ericboucher about this. `playwright` is powerful, but also very clunky to setup, and has resulted in very flaky tests. To do frontend testing, we would have to either install playwright in javascript (currently we use the python flavour for testing), which would double the setup time/number of dependencies, or test the frontend from the backend codebase, which feels wrong.

Eric suggested we give cypress a shot, as we use it elsewhere.

This PR is a quick attempt to setup end to end testing with it.

Impressions so far:
- initial setup is much easier than playwright. I had the cypress UI running locally in 2 minutes, most of which was downloading binaries.
- I suspect we might hit some flakiness in tests as well, particularly because cypress seems to run within the same JS thread as our code.
- like playwright, you write tests by selecting DOM elements and clicking on them, and then asserting various things. I did not manage to change the language for Cambodia for instance, cypress seemed to go into an endless refresh loop.
- other than that, it should be reasonably easy to write tests with this, as it does not require any setup beyond picking a URL as a starting point.

## How to test the feature:

- [ ] `npx cypress run` or
- [ ] `yarn cypress open` for the UI, then click the `e2e` tests, then click on the `spec` file, and watch tests run

## Checklist - did you ...

Test your changes with


ericboucher:
The windows tests are mostly meant to make sure that country offices using windows can run a full build on their own. I'd be fine to simply run an install+build for one country

frontend/cypress/support/component-index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
```
ericboucher, line 1: Is this file needed?

laurentS:
Closing this PR as it was replaced by #1524

PR #1525, opened 2025-08-24
Refactor map state

conor-jh:
### Description
This PR sets the foundation for supporting multiple map instances. See [Proposal for Layer/Date Selection that Supports Multiple Maps](https://www.notion.so/whitecap/Proposal-for-layer-date-selection-that-supports-multiple-maps-24c0c175c4d88063b023c371be07499a) for additional details.

The primary goal at this stage is to decouple map-related UI components from hard-coded selectors and actions so that future map instances can read/write to arbitrary slices within the global Redux store.

To that end, `useMapState` is a new hook that serves as a sort of migration/abstraction layer. Its main responsibility is to provide map-related selectors and actions to children within a particular map _context_. For now, that will always be the primary map—so the returned selectors and actions will work with the existing `mapStateSlice`. 

Key pattern —  replacing direct Redux selectors/actions:
```typescript
// Before
const selectedLayers = useSelector(layersSelector);
dispatch(addLayer(layer));

// After  
const mapState = useMapState();
const selectedLayers = mapState.layers;
mapState.actions.addLayer(layer);
```

This PR also introduces some scaffolding for a `MapInstanceContext` provider, which will wrap future map instances and link them to a particular slice of state.

### Scope of migration
* Focus in this PR is on migrating the core map-related UI components, particularly for layer and date selection. 
* I tried to keep other UI components that use `mapStateSlice` selectors out of scope for now (e.g. analysis panel, anticipatory action, etc.)

## How to test the feature:

Ensure core map functionality works—users can select layers, dates, etc. and all layers load as expected following existing behavior. 

> [!NOTE]
> I did merge `setup-e2e-testing-with-cypress` locally and Cypress tests are passing with these changes

## Checklist - did you ...

Test your changes with

- [] Add / update necessary tests?


frontend/src/components/MapView/LeftPanel/layersPanel/MenuItem/MenuSwitch/SwitchItem/index.tsx
```diff
@@ -168,21 +169,22 @@ const SwitchItem = memo(
...
-        refreshBoundaries(map, dispatch);
+        refreshBoundaries(map, mapState.actions);
       },
       [
         appendLayerToUrl,
         dispatch,
         group,
         layer,
         map,
+        mapState.actions,
```
gislawill, line 187: Out of curiosity, do you have a sense of how often the pointer for mapState.actions updates? I see that it's memoized so I assume not much. Just hoping to confirm we're not introducing a new and unnecessary trigger for useEffect calls across the app

gislawill:
Honestly, a much smaller diff than I was expecting. That's great.

ericboucher:
Looks great!

laurentS:
Nice! Like Will, I expected a much bigger diff. My only concern going forward is the size of the `mapState`. See [my notes here](https://github.com/WFP-VAM/prism-app/pull/1526#issuecomment-3236477804). The map slice is already huge, so it might be good to dig a bit further to see what we can break out to avoid replicating entire chunks of state across the various maps. (I'm of the opinion that a lot of the data that won't change during a session like admin geometries should be moved out of the redux tree entirely and just have flags (`loading/loaded`...) in the state, but maybe that would make some stuff harder than needed?)

gislawill (approved):
This is looking good to go to me @conor-jh, thanks! I tested it out across a few deployments and across as many features as I could think of — no issues! Nice work.

ericboucher (approved):

PR #1537, opened 2025-09-25
Dashboards analytics tables

gislawill:
### Description

This PR adds support for the analysis results (zonal stats) inside our dashboards. We support both the preview and edit mode in this PR. The edit mode allows you to defined the parameters for the analysis results. The preview mode just shows the results in the table and overlays it on the map. 

## How to test the feature:
- [ ] Open Mozambique and load the Spring Dashboard
- [ ] Change over to edit and edit some parameters
- [ ] Re-run the analysis results

## Checklist - did you ...

Test your changes with

Screenshot 2025-09-26 at 5 18 18 PM

### DIFF
--- a/frontend/src/components/DashboardView/TableBlock.tsx
+++ b/frontend/src/components/DashboardView/TableBlock.tsx
@@ -264,7 +264,7 @@ function TableBlock({
   return (
     <Box className={classes.grayCard}>
       <Typography variant="h3" className={classes.blockTitle}>
-        Block #{index + 1}
+        {t('Table Block')} #{index + 1}
       </Typography>
 
       <Box className={classes.formContainer}>


frontend/src/components/Common/AnalysisFormComponents/AdminLevelSelector.tsx
gislawill, line 1: The components in this folder all came from the AnalysisPanel form. I pulled them into a common component source to reduce duplication between these features.

frontend/src/utils/analysis-hooks.ts
gislawill, line 1: Abstracting out the analysis form business logic to hooks that are used both in dashboards and analysis panel

frontend/src/utils/analysis-hooks.ts
```diff
@@ -0,0 +1,449 @@
...
+          baselineLayer.boundary
+        ] as BoundaryLayerProps;
+        // Remove displayed boundaries
+        getDisplayBoundaryLayers().forEach(l => {
+          if (l.id !== boundaryLayer.id) {
+            safeDispatchRemoveLayer(map, l, dispatch);
+          }
+        });
+
+        safeDispatchAddLayer(
```
conor-jh, line 317: I’m noticing a few instances in the analysis hooks where we’re explicitly dispatching map state slice actions or using map state slice selectors. Is that intentional or do we want to swap out with actions/selectors derived from `useMapState`?

frontend/src/components/DashboardView/TableBlock.tsx
```diff
@@ -0,0 +1,449 @@
...
+          </Box>
+        )}
+      </Box>
+    );
+  }
+
+  return (
+    <Box className={classes.grayCard}>
+      <Typography variant="h3" className={classes.blockTitle}>
+        Block #{index + 1}
```
conor-jh, line 265: Do we have any interest in specifying the block types between text, charts, and analysis?

conor-jh: Also can you wrap strings in a `t` function throughout this component?

conor-jh (commented):
I think this is looking great, and I appreciate the abstractions around the components and new analysis hooks. Just left a few questions re: map state

frontend/src/utils/analysis-hooks.ts
gislawill: That's a really good question. I'm not sure. I don't think the analysis tab or the table (analysis) block are wrapped in map contexts. Though it seems to be working well without it?

Do you understand why the analysis results are being put on the dashboard map despite explicitly dispatching map state slice actions?

frontend/src/components/DashboardView/TableBlock.tsx
gislawill: Ha! Good catch 😅 Will add the translation support. Regarding the block types, do you mean just specifying it in the copy?

ericboucher (approved):
This looks great, thanks for abstracting the analytics part.

Not super happy with the block disposition but we can iterate on design/UX later.

Let's also think about how to improve caching mechanisms of the API to speed up the loading time of analysis in dashboards

frontend/src/components/DashboardView/TableBlock.tsx
conor-jh: Yeah exactly… might be more of an Amit question

frontend/src/utils/analysis-hooks.ts
conor-jh: Oh… good point about the context. What’s the desired behavior here—do we want selections in an analysis block to impact the dashboard map(s)?

I’m frankly a bit stumped as to why the dashboard map is currently displaying analysis-driven layers. My best guess is that some call in the analysis logic is invoking an `addLayer` action outside a map context, but I can’t really explain why or how the map instance would then read _from_ the global map store :/

This is a good example of where it would be super helpful to have working Redux dev tools, we’d be able to trace all the actions and selector calls and more easily narrow it down.

gislawill: >  What’s the desired behavior here—do we want selections in an analysis block to impact the dashboard map(s)?

Yes, we do. I think there might be some updates coming where we configure if this is added to the map or not. TBD based on conversations with Amit. I'll side-step this for now but I'm confident we'll come back to it soon.

PR #1538, opened 2025-09-26
AA Storm Fixes

ericboucher:
### Description

- Switch to 119k/h
- Monitor all disctricts in coastal provinces


wadhwamatic:
@ericboucher - not related to these changes, but can we default to the closest to landfall storm where there is more than one storm? for example, on 10-March, we should show Jude instead of Ivone 

https://staging-prism-frontend--prism-1538-skor17r3.web.app/?hazardLayerIds=anticipatory_action_storm&date=2025-03-10

wadhwamatic:
@ericboucher - we can change the term 'pilot districts' to 'monitored districts' in the legend

alerting/src/aa-storm-alert/alert.ts
```diff
@@ -138,34 +139,35 @@ export function transformReportsToLastProcessed(
   return lastProcessedReports;
 }
 
+// TODO - make sure spelling is exactly the same for districts as in the report
+// or add a mapping below if needed
+const districtNameMapping: { [key: string]: string } = {
```
ericboucher, line 144: Check with Valentin once the data has been updated on their end.
cc @valpesendorfer

PR #1540, opened 2025-09-30
Place flex elements below maps in multi-map dashboards

conor-jh:
### Description

This introduces a layout change to place flex elements in multi-map dashboards below each map. See inline comments.

## How to test the feature:

1. Run app using `REACT_APP_COUNTRY=mozambique yarn start`
2. Select “Precipitation Comparison Dashboard” from the dashboards menu in the nav bar
3. Observe text boxes under each map

## Checklist - did you ...

<!-- If any of the following items aren't relevant to your contribution,
     please still tick them so we know you've gone through the checklist. →

Test your changes with

### DIFF
--- a/frontend/src/components/DashboardView/index.tsx
+++ b/frontend/src/components/DashboardView/index.tsx
@@ -26,6 +26,15 @@ function DashboardView() {
   const [mode, setMode] = useState<DashboardMode>('preview');
   const { t } = useSafeTranslation();
 
+  const isTwoMapLayout = dashboardMaps.length === 2;
+
+  const getFlexElementsForMap = (mapIndex: number) => {
+    if (!isTwoMapLayout) {
+      return [];
+    }
+    return dashboardFlexElements.filter((_, index) => index % 2 === mapIndex);
+  };
+
   const handlePreviewClick = () => {
     setMode('preview');
   };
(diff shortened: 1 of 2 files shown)


frontend/src/components/DashboardView/index.tsx
```diff
@@ -77,37 +86,43 @@ function DashboardView() {
...
-                  className={classes.titleBarInput}
-                  placeholder={t('Enter dashboard title')}
-                  value={dashboardTitle}
-                  onChange={e => dispatch(setTitle(e.target.value))}
-                  name="dashboard-title"
-                />
-              </label>
-            </Box>
-          )}
+      {isTwoMapLayout ? (
```
conor-jh, line 89: I’m not tied to this solution at all… in fact I don’t think it’s likely to be sustainable. I can give it a bit more thought and maybe we combine this work with other layout-specific concerns. 

My general POV is that we’ll probably need to nail down what layout config options are available vs. what should we infer? 

For instance, when multiple maps are included in a deployment’s config, we treat the first one as the “left” side and the second as the “right side”. But is that what we want?

How can a user configure a dashboard with a text box only for the second map?

I think we’ll need to really nail down some business logic about what constitutes a valid config, what to do with invalid configs (e.g. what happens if a consumer adds 6+ flex elements?) and if there are any other layout options we want to expose. 

From there I would think we can create some sort of `<Layout />` abstraction that takes care of how to arrange all the blocks given various configs.

gislawill (commented):
Thanks @conor-jh! Overall, it's looking great! Nice work, especially with the map syncing. A few thoughts here:
* I see what you mean about how complicated the render statement of DsahboardView is getting. I feel like our separation of maps from flex elements is part of the problem. They're all flex at this point. Curious if you think merging into one component list would make it simpler or more complex?
    * Somewhat related, do you know if [this view](https://www.figma.com/design/5GX8jVxBCfLrcfkTniRvFi/WFP-Prism?node-id=2150-869&m=dev&t=aM99KhbOxw5Cdkas-1) is currently supported? 
* Could we change the cursor when the map is fixed? We're showing a grab cursor but there's no available grab interaction.
* How would you feel about one sync maps toggle that is just to the right of the dashboard title rather than a part of both map blocks? Feels like slightly cleaner design to me, but it's a pretty loosely held opinion. 
* Let's continue to leave the title as full width and both maps beneath it. Current:
    Image

ericboucher:
Could it make sense to have the config menus as popup cards instead of elements that shift the layout?

gislawill:
> Could it make sense to have the config menus as popup cards instead of elements that shift the layout? And get closer to edit/preview mode layout looking the same?

This suggestion honestly makes a lot of sense. My main reaction is that I wish we had thought of it earlier. 

I suppose I still think that the map edit mode is currently useful (compared to a simplified config and a pop-up card) — it's simpler to just reuse the existing components that are designed to be alongside the map. But I do agree that the edit modes for tables, charts, and text boxes would make more sense as a config pop-up card than a whole separate page. 

We're going to run a walkthrough with Amit of the whole feature on Thursday, and we'll bring up this idea. 

> I don't feel like "Dashboard Title" and some of the other title elements are needed.

Why don't you think these are needed and do you think they detract from the user experience?

PR #1542, opened 2025-10-01
Dashboard exports to PNG/PDF

gislawill:
## How to test the feature:
- [ ]
- [ ]

## Checklist - did you ...

Test your changes with


wadhwamatic:
Hey @gislawill. I started testing this PR locally and came across a few issues:
1. Any changes made to the map (zoom / pan) are not retained when I go to the print page
2. Admin area selection in the print view doesn't function
3. Map labels toggle in print view doesn't work
4. Two map / print function moves very slowly and frequently crashes
8. Scrollable content is cutoff when printed (ex: spring dashboard)

gislawill:
This PR has gotten far too large. I'm spinning out the remaining the issues into follow-up PR:
- Browser crashing when many maps are initialized (publish is entered and exited multiple times)
- Handling content overflow
- Locking the map view in edit across to preview and publish

@conor-jh could I ask for your code review? Please approve or request changes

ericboucher:
I think we should aim to resize. Just calculate the max height/width and scale to A4, keeping the aspect ratio intact.

- Let's confirm with @wadhwamatic but I would recommend **defaulting to A4 landscape**, and even keeping A4 only.
- The ratios are way too different between what I see on the dashboard before printing and what I see in the print mode. We should either fix that, or make the map controllable in print mode. Otherwise it feels impossible to know what part of the map will get selected in the print mode.
- Wouldnt it make more sense to simply select the format when clicking "download" instead of having a format selector?
- Maybe hide the overflow to make it more obvious what actually gets printed?

IMHO, the dashboard view itself should be capped to something that looks like an A4 landscape in terms of aspect ratio, that would make a lot of what we are doing here simpler. And if the end goal is to easily display it as a static page and print, I believe it would make more sense too.

For the purpose of **this** PR, and to avoid conflicts:
- let's fix the map zoom/location behavior
- let's hide the overflow
- merge :)

Hope this is helpful, and let's continue discussing the rest.

gislawill:
@ericboucher, when you say "let's fix the map zoom/location behavior", I believe you're referring to this point:
> The ratios are way too different between what I see on the dashboard before printing and what I see in the print mode. We should either fix that, or make the map controllable in print mode. Otherwise it feels impossible to know what part of the map will get selected in the print mode.

How would you feel about the fix being storing the viewport the user sets in edit mode and applying that as the min bounds for the print mode map? Alternatively, we could make the map zoom and pan-able in print mode, but we've discussed with @wadhwamatic that might open up too much customization & flexibility in frequently referenced dashboards (where consistency is valued).

conor-jh (approved):
Looks good to me—I am definitely anticipating some major conflicts with the new “layout engine” but I’d rather get this merged in and deal with any fallout on that branch separately. Once this is in I can rebase the layout changes on top of `dashboards-phase-one` or just a cut a clean branch and cherry-pick changes over.

Issue #1453, opened 2025-03-14
[Feature Request]: Create a dashboard / report view


wadhwamatic:
### Provide a clear and concise description of what you want to happen.

Allow users to create a customized view that includes a map, chart(s) and free text area that can be exported as a PDF

WIP mockup (only for communicating the concept, not a final design): https://www.figma.com/design/5ZTNgCyn3S0GliMgV8oWIN/PRISM---Dashboards?node-id=0-1&p=f&t=aMkJXKpBVBDIq67e-0

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

Outstanding question: Should we persist the once it's been configured?

PR #1543, opened 2025-10-02
Dashboards phase one

gislawill:
### Description

This feature branch addresses #1453 by creating configurable reports.

All code and functionality has been previously reviewed. Here it is, all together, for the first time.

Screenshot 2025-11-15 at 4 46 42 PM


wadhwamatic:
@gislawill & @conor-jh. Great work on this. I spent a good amount of time with this feature today and have a lot of feedback. I'll break it down into issues that seem to be bugs, and improvements to consider.

**Potential bugs / required changes**
1. The SPI dashboard map is configured for 21-March, but the text above the map shows March 20. Guessing this is a timezone issue?
2. In the SPI dashboard, the analysis table is configured for 21-March, but in the dashboard the table shows 21-Sept
3. If I switch between the SPI dashboard and either of the others while an analysis is being run, the analysis results incorrectly get applied to the map in the other dashboard
4. The chart config uses the parameter `wmsLayerId` but I believe it refers to the PRISM `layerId`. We should just call it `layerId`. Am I correct in assuming this grabs the chart configuration from the associated layer? It seems correct, but I didn’t verify yet.
5. There's another chart config param called `adminUnitLevel` that doesn't seem to have an impact. If it’s intended to allow us to query the stats API at different levels, it should also require an `adminId` parameter. By default we should use the `countryAdmin0Id` set in prism.json if `adminUnitLevel`=0. If `adminUnitLevel`=1 or 2, it should require an `adminId` parameter
6. In the context of dashboards, adding an analysis table shouldn’t require adding the output as a layer in the map. Please add the option to disable the analysis result layer result from the map view

**Suggested improvements**
1. It seems inevitable that users will need to be able to adjust map views (panning and zoom) based on their screen resolution. How difficult is it to make this part of the general dashboard view (not just edit view)
2. It would be helpful to be able to control the legend size in the config, and to optionally hide the legend as well.
3. I didn’t try to add translations for the dashboard text yet. Please make sure that text is translatable.
4. I’d like to be able to add a default sort column and direction (ascending vs descending) in the analysis table
5. When content spills over in the browser view, can we make each component scrollable and give equal relative height within a column? In this screenshot, I'd suggest the text and table panels each become the same height (50% of map height minus padding) and both become scrollable
Screenshot 2025-10-28 at 14 33 20

gislawill:
Thanks for the detailed review on this @wadhwamatic! Overall, I think these updates are very do-able. I'll respond to these issues one by one but **tl;dr: I believe we can make all of these required updates swiftly.**

**Questions for you @wadhwamatic are in bold.** 

I'll also keep this checklist up to date as we add fixes: 

> **Potential bugs / required changes**
> 
> - [ ] The SPI dashboard map is configured for 21-March, but the text above the map shows March 20. Guessing this is a timezone issue?

   - Yeah, almost certainly a time zone issue. We'll get a quick fix in for this. 

> - [ ] In the SPI dashboard, the analysis table is configured for 21-March, but in the dashboard the table shows 21-Sept

   - Thank you for catching this. We are not using the correct date when passed in by the configuration for the analysis results. Should be an easy fix thankfully but thank you for catching it.

> - [ ] If I switch between the SPI dashboard and either of the others while an analysis is being run, the analysis results incorrectly get applied to the map in the other dashboard
   
   - Oh shoot, good catch! I'll make sure that this fetch is canceled if we change maps. 
   
> - [ ] The chart config uses the parameter wmsLayerId but I believe it refers to the PRISM layerId. We should just call it layerId. Am I correct in assuming this grabs the chart configuration from the associated layer? It seems correct, but I didn’t verify yet.

   - We can switch that over to just the layer ID. And yes, that's exactly right. It pulls it from the chart configuration from the layer.json. 

> - [ ] There's another chart config param called adminUnitLevel that doesn't seem to have an impact. If it’s intended to allow us to query the stats API at different levels, it should also require an adminId parameter. By default we should use the countryAdmin0Id set in prism.json if adminUnitLevel=0. If adminUnitLevel=1 or 2, it should require an adminId parameter

   - Ah yes, this set up is incomplete. We'll fix this configuration option.
   - **Is this a feature (a chart for a sub-country admin unit) you'd like to include in the demo?** 

> - [ ] In the context of dashboards, adding an analysis table shouldn’t require adding the output as a layer in the map. Please add the option to disable the analysis result layer result from the map view.

   - Will do.
> 
> **Suggested improvements**
> 
> - [ ] It seems inevitable that users will need to be able to adjust map views (panning and zoom) based on their screen resolution. How difficult is it to make this part of the general dashboard view (not just edit view)

   - It's very easy – it's the default behavior. We removed it after our discussion on Oct 9th during which we decided that we did not want to give users the option to change the map view on maps they didn't have edit access for.
   - The map view is determined by the most zoomed-in state that can support the defined map bounds. For our A4 exports, that's always the exact same image. When viewing in a browser, though, that leads to different sizes. 
   - **We could enable panning and zooming on preview, but still lock it down in exports?** 

> - [ ] It would be helpful to be able to control the legend size in the config, and to optionally hide the legend as well.

   - We do have that option in the print configuration modal 
   - If we enable panning and zooming, is that still important in the preview? 

> - [ ] I didn’t try to add translations for the dashboard text yet. Please make sure that text is translatable.
 
   - All the text that is rendered by our React code is translatable. 
   - **Do you expect the text that's written in the prism.json to also need to be translatable?** 

> - [ ] I’d like to be able to add a default sort column and direction (ascending vs descending) in the analysis table

   - Can do. 

> - [ ] When content spills over in the browser view, can we make each component scrollable and give equal relative height within a column? In this screenshot, I'd suggest the text and table panels each become the same height (50% of map height minus padding) and both become scrollable

   - We can set max heights on our components containers and allow content to scroll if it's too long. 
   - **How do you think we should handle dashboards with three items in a column in this set up?** 
   - **In the screenshot, how would you suggest handling the map height? Would that also be cut off at the edge of the viewport?** 
   - Happy to continue this discussion on a call or also do some rounds with Marc to test out different dashboards on various screen sizes.

wadhwamatic:
@gislawill - I'm sharing feedback I received from a demo today with our primary stakeholder in WFP HQ. First and foremost, this is very well received and will be a really useful feature. Well done!

None of these for action yet, but to consider as we start planning phase two and priorities going forward for this feature.
1. It would be very useful if we could add the ability to create a collection / folder of dashboards which can then be shown as sub-navigation sections in the Dashboards menu. For example, I want to be able to create a collection of dashboards showing 1-month SPI map, chart and text for every month between Oct to March
2. Related to the above, I'd like to be able to pre-define dashboards for entire season (Oct to March) and have the newest dashboard automatically get added to the list when new data becomes available. For example, I'd like to setup dashboards for each month now (Oct) but have the Nov - March dashboards appear once data for those months is available
3. Add the ability to navigate between dashboards within a collection using a previous / next arrow in the UI
1. Support dynamic text content that can be updated externally and without a build process
2. The deployment model in general needs a deeper think through. I'd like to consider a separate dashboard config file that can be updated without requiring a build. Though addressing points 2 and 4 above may make this less of an issue
3. There's unused white space at the top of the dashboard UI that could be removed
4. If we are to have full window vertical scrolling, we should keep the top nav sticky

gislawill:
Ok @wadhwamatic, I'm excited to say that this big feature branch is ready for review 🙌

PR #1550, opened 2025-10-14
Fix/layer removal from legends panel

conor-jh:
### Description

This fixes the issue where users couldn’t remove a layer from the legends panel in dashboard maps.

Two questions/issues to answer:
- How do we feel about not re-running analysis results automatically? Without f0dbf9f1979b6bb220733eda6b3ce5e561abc1c0, upon removing a layer from the legends panel, the table will re-run and add the layer back. But I’m not sure this is an optimal solution
- Do we want to allow users to remove layers in both edit and preview/publish mode? Right now, the implementation allows for either, but that could change here or in the ticket that introduces config-driven editing permissions

## How to test the feature:

- [ ] Run the Mozambique deployment locally (`REACT_APP_COUNTRY=mozambique yarn start`)
- [ ] Open any dashboard page
- [ ] Select the **X** icon in any legends panel
- [ ] Observe the layer getting removed from the map instance

## Checklist - did you ...

Test your changes with


conor-jh:
> Hey @conor-jh could you say more about what you mean by "upon removing a layer from the legends panel, the table will re-run and add the layer back". In my brief tests, I was seeing the layer be removed as well as the table data and not re-added. Is this what you're referring to?
> 
> On your second question, I believe we should only allow users to remove layers this way in the edit view (not preview or print), but could you confirm @wadhwamatic?

Just to confirm, did you test that with [b6b828b](https://github.com/WFP-VAM/prism-app/pull/1550/commits/b6b828b2c680caa3d88b408756c7078df3caa426) reverted?

gislawill:
Hey @ericboucher, flagging this for your review. It enables the legend component to show the opacity & remove layer options while in "Edit Mode". It also uses enums for dashboard constants.

The bulk of the massive diff is a just snapshot test update.

PR #1551, opened 2025-10-14
Dashboard deeplink support

conor-jh:
### Description

This PR adds a route parameter to `dashboards/*` allowing users to deep link to specific, pre-configured dashboards. Users can supply a `path` in the app config—if a value isn’t provided, a path is automatically generated from the title. The URL syncs with the selected dashboard, so users can easily copy the URL for a particular report.

Note: this did require updating some relative resource paths.

## How to test the feature:

- [ ] Run the Mozambique deployment
- [ ] Navigate to any dashboard and observe the URL
- [ ] Navigate directly to a URL such as http://localhost:3000/dashboard/spring-dashboard and observe that the relevant dashboard loads correctly

## Checklist - did you ...

Test your changes with

### DIFF
--- a/frontend/src/config/mozambique/prism.json
+++ b/frontend/src/config/mozambique/prism.json
@@ -504,6 +504,7 @@
     {
       "title": "Summer Dashboard",
       "path": "summer-temperature-report",
+      "isEditable": true,
       "maps": [
         {
           "defaultDate": "2025-06-01",
--- a/frontend/src/components/DashboardView/index.tsx
+++ b/frontend/src/components/DashboardView/index.tsx
@@ -20,10 +20,8 @@ import {
 
 import MapBlock from './MapBlock';
 import {
-  dashboardTitleSelector,
+  dashboardConfigSelector,
   setTitle,
-  dashboardFlexElementsSelector,
-  dashboardMapsSelector,
   dashboardSyncEnabledSelector,
   toggleMapSync,
   dashboardModeSelector,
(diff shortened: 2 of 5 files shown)


frontend/src/config/mozambique/prism.json
```diff
@@ -503,6 +503,7 @@
     },
     {
       "title": "Summer Dashboard",
```
ericboucher: Should we simply use a kebab-case version of the title? wdyt?

ericboucher: I agree that having a path makes sense right now, but my worry is that we're in an in-between state :
- it's not super useful if we only have preconfigured dashboards that have a somewhat fixed config.title (regardless of what users end up changing it to)
- if we end up moving towards a more complicated approach where users can create their own dashboards and publish them, paths would become obsolote and we would likely have to move to unique ids anyway.

I'm fine either way, I just feel like it's a not super needed complexity given our current "in-between" state :)

ericboucher (approved):

frontend/src/config/utils.ts
```diff
@@ -66,16 +67,35 @@ export function deepCamelCaseKeys(obj: any): any {
   return obj;
 }
 
+// Helper function to ensure data paths are absolute
```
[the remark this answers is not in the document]
conor-jh: To be honest, I don’t think it’s a matter of paths beginning with a `/` it’s just based on how static assets are served from `/data/` and how that conflicts with the dynamic param.

PR #1557, opened 2025-10-21
Memory leak fix: moving boundary data out of redux

gislawill:
### Description

Inspired by @laurentS's concerns about the size of the map state, this PR addresses the sluggish performance, especially within the dashboard module, by moving our boundaries data out of Redux and into a global cache that we load once and then can pull from. 

**After this update, the site's JS memory allocation when showing dashboards with two maps dropped to just 10% of memory allocation before the update (2.2 GB => 256MB).**

In my tests, this makes the entire Prism tool feel snappier and makes dashboards much snappier. Plus, it hasn't crashed once this update. 

The heap snapshots are on my computer and I'm happy to share with anyone interested. Compressed, they're about 50 MB and 500 MB. 

## How to test the feature:
- [ ] Load up the deployment and open some dashboards. 
- [ ] Jump back and forth between dashboards in the print menu and the edit menu.

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,12 +339,11 @@ Use the `useBoundaryData` hook to access boundary data:
 
 ```typescript
 import { useBoundaryData } from 'utils/useBoundaryData';
-import { mapSelector } from 'context/mapStateSlice/selectors';
-import { useSelector } from 'react-redux';
+import { useMapState } from 'utils/useMapState';
 
 function MyComponent() {
-  const map = useSelector(mapSelector);
-  const { data, loading, error } = useBoundaryData('admin_boundaries', map);
+  const { maplibreMap } = useMapState();
+  const { data, loading, error } = useBoundaryData('admin_boundaries', maplibreMap);
   
   if (loading) return <div>Loading boundaries...</div>;
   if (error) return <div>Error: {error}</div>;


frontend/src/components/MapView/Layers/BoundaryDropdown/BoundaryDropdownOptions.tsx
```diff
@@ -87,39 +83,43 @@ const BoundaryDropdownOptions = React.forwardRef(
...
+          {} as Record<string, BoundaryLayerData | undefined>,
+        ),
+      [],
     );
 
     // Combine the data from all layers
     const combinedData = useMemo(() => {
-      const layerData = Object.values(allBoundaryLayerData).filter(Boolean);
+      const layerData = Object.entries(allBoundaryLayerData)
+        .filter(([, data]) => data !== undefined)
```
laurentS: `null` vs `undefined` here again.

frontend/src/utils/analysis-hooks.ts
```diff
@@ -150,9 +148,19 @@ export const useAnalysisForm = (
...
+    useMemo(() => {
+      if (!boundaryDataResult.data || !adminLevelLayer) {
+        return undefined;
+      }
+      return {
+        layer: adminLevelLayer,
+        data: boundaryDataResult.data,
+        date: Date.now(),
+      };
+    }, [boundaryDataResult.data, adminLevelLayer]);
```
laurentS: The function passed to `useMemo` should be pure, so it shouldn't really access the current time. But more generally, I don't think `useMemo` is needed at all here, because the calculation you're running isn't actually expensive.
I would try to go with the following instead, what do you think?
```suggestion
  const adminLevelLayerData: LayerData<BoundaryLayerProps> | undefined =
    boundaryDataResult.data && adminLevelLayer
      ? {
          layer: adminLevelLayer,
          data: boundaryDataResult.data,
          date: Date.now(),
        }
      : undefined;
```
(I think there are lots of places in the codebase where we could do the same thing, but that's a separate discussion)

ericboucher (approved):

PR #1564, opened 2025-10-26
Chart heights config

conor-jh:
### Description

Allows users to select from three predefined chart heights to control overall dashboard spacing. 

This ended up being less trivial than I expected—I’m wondering if there’s a different approach we can take, but the docs kind of make it seem like we have to trade-off some of Chart.js’s built-in height logic to control the total height more explicitly. 

The biggest risk is that charts can become too short where the y-axis labels overlap. To get around this, I bumped up the shortest and tallest options slightly, but am very open to other ideas here. 

## How to test the feature:

- [ ] Run the Mozambique deployment and make the spring dashboard editable (add `isEditable: true` to the config)
- [ ] In EDIT mode, choose a different chart height, then switch to PREVIEW and EXPORT modes.

## Checklist - did you ...

Test your changes with

### DIFF
--- a/frontend/src/components/DashboardView/ChartBlock.tsx
+++ b/frontend/src/components/DashboardView/ChartBlock.tsx
@@ -353,6 +353,20 @@ function ChartBlock({
               );
             }}
           />
+          <FormControl variant="outlined" className={classes.formControl}>
+            <InputLabel>{t('Chart Height')}</InputLabel>
+            <Select
+              value={chartHeightOption}
+              onChange={e =>
+                setChartHeightOption(e.target.value as ChartHeight)
+              }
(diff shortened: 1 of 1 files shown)


gislawill (commented):
@conor-jh Seems like the kind of thing we'll want to configure inside the dashboard configuration. Any objections to adding it there as well?

gislawill (approved):
Looks good to me! Thanks Conor!

@ericboucher and @wadhwamatic, ready for your review

wadhwamatic (approved):
looks good, thanks @conor-jh. I like that I was able to add `chartHeight` as a config parameter in prism.json too

PR #1606, opened 2025-11-09
Make MapBlock legends configurable

gislawill:
### Description

This adds the ability to configure each map's legend visibility and position. Both within the edit mode and prism.json

Reference Slack conversation: https://ovio-public.slack.com/archives/C0958FADX2M/p1762362361109979

Screenshot 2025-11-09 at 5 37 33 PM

### DIFF
--- a/frontend/src/components/DashboardView/DashboardLegends.tsx
+++ b/frontend/src/components/DashboardView/DashboardLegends.tsx
@@ -22,13 +22,7 @@ const DashboardLegends = memo(
       exportConfig?.toggles?.legendVisibility ?? legendVisible ?? true;
     // Convert position: exportConfig uses 0=left/1=right, map state uses 'left'/'right'
     const exportPosition = exportConfig?.legendPosition;
-    const position =
-      // eslint-disable-next-line no-nested-ternary
-      exportPosition !== undefined
-        ? exportPosition
-        : legendPosition === 'left'
-          ? 0
-          : 1; // default right (1)
+    const position = exportPosition ?? (legendPosition === 'left' ? 0 : 1); // default right (1)
     const scale = exportConfig?.legendScale ?? 0; // default 100% (0 means no reduction)
 
     if (!isVisible) {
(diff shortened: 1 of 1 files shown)


frontend/src/components/DashboardView/DashboardLegends.tsx
```diff
@@ -8,34 +8,48 @@ import type { ExportConfig } from './DashboardContent';
...
+const DashboardLegends = memo(
+  ({ exportConfig, legendVisible, legendPosition }: DashboardLegendsProps) => {
+    const classes = useStyles();
+    const mode = useSelector(dashboardModeSelector);
 
-  if (!isVisible) {
-    return null;
-  }
+    // Priority: exportConfig takes precedence if present, otherwise use map state values
+    const isVisible =
```
conor-jh: If `exportConfig` takes precedence, is there any expectation that modifying legend visibility in export view should impact the “normal” dashboard view? Or is the idea that you can just override legend visibility _within_ the export view?

frontend/src/components/DashboardView/DashboardLegends.tsx
```diff
@@ -8,34 +8,48 @@ import type { ExportConfig } from './DashboardContent';
...
+      exportConfig?.toggles?.legendVisibility ?? legendVisible ?? true;
+    // Convert position: exportConfig uses 0=left/1=right, map state uses 'left'/'right'
+    const exportPosition = exportConfig?.legendPosition;
+    const position =
+      // eslint-disable-next-line no-nested-ternary
+      exportPosition !== undefined
+        ? exportPosition
+        : legendPosition === 'left'
+          ? 0
+          : 1; // default right (1)
```
conor-jh: You could avoid the nested ternary here with a nullish coalescing operator 
```suggestion
    const position = exportPosition ?? (legendPosition === 'left' ? 0 : 1);
```

frontend/src/components/DashboardView/DashboardLegends.tsx
```diff
@@ -8,34 +8,48 @@ import type { ExportConfig } from './DashboardContent';
...
+    if (!isVisible) {
+      return null;
+    }
+
+    return (
+      <aside
+        className={classes.container}
+        style={{
+          left: position % 2 === 0 ? '8px' : 'auto',
+          right: position % 2 === 0 ? 'auto' : '24px',
```
conor-jh: What’s the reason for the difference in values between left/right configurations?

frontend/src/components/DashboardView/DashboardLegends.tsx
gislawill: Good call, updating

Issue #1452, opened 2025-03-14
[Feature Request]: Batch produce maps


wadhwamatic:
### Provide a clear and concise description of what you want to happen.

Allow users to publish multiple maps from a single interface to avoid manually repeating the process.

We should also consider capturing print map setup as a template that can be reused. This would allow a user to position the map once, and then apply for consistent map production.

Stretch goal: Add more than one indicator to the map input screen so that a batch of maps for each indicator can be produced. This would save time for users.

Future iteration would include a scheduled approach

Issue #1611, opened 2025-11-13
[Feature Request]: Add date coverage window to legend and print footer


wadhwamatic:
### Provide a clear and concise description of what you want to happen.

PR https://github.com/WFP-VAM/prism-app/pull/1572 will introduce date coverage as a concept. This will better describe the date range that a selected layer's data covers. This will be shown in the tooltip when hovering over the timeline.

In addition to showing it in the tooltip, it would be useful to see this in the legend and in the footer when printing. When coverage is not provided in the configuration, this could display the validity period.

See examples below.

**Legend**
![Image](https://github.com/user-attachments/assets/8d622ece-59d3-4793-bdc7-b5e65217f8a9)

**Footer in print screen**
![Image](https://github.com/user-attachments/assets/cc2ff207-b855-4923-970e-539f6db56f2d)

PR #1633, opened 2025-11-30
Batch Maps Feature Branch

gislawill:
### Description

This PR adds batch map export functionality (#1452), allowing users to generate multiple maps across a date range and download them as a merged PDF or ZIP of PNGs. The feature uses server-side rendering (playwright) to build and package PRISM rendered maps.

**Architecture**
The frontend's Print Dialog constructs URLs for each date in the selected range and sends them to a new /export API endpoint. The backend uses Playwright to render each URL via a dedicated /export frontend route (which reads all config from URL params), waits for window.PRISM_READY, then captures and packages the results.

```
Print Dialog (PRISM Frontend - image.tsx)
       │
       ▼ downloadBatch() constructs URLs
POST /export (PRISM API Server)
       │
       ▼ Playwright renders each URL
/export route (PRISM Frontend - ExportView)
       │
       ▼ Sets window.PRISM_READY when loaded
Returns merged PDF or ZIP
```

**Aspect Ratio Update**
This branch now includes the aspect ratio configuration feature (merged from #1648), which replaces the previous "Map Width" percentage slider with a more intuitive aspect ratio selector. Users can now choose from preset ratios, A4 print formats, a custom ratio input (w:h), or "Auto" mode which allows the map to resize based on the browser window. This approach provides both flexibility and consistency for **both** single and batch map exports.

## How to test the feature:
1. Start the app with a country that has date-supported layers: `REACT_APP_COUNTRY=mozambique yarn start`
4. Select a layer with date support (e.g., rainfall forecast)
5. Open Print Dialog (camera icon in navbar)
6. Configure appearance (title, legend position, etc.)
7. Enable batch maps: Toggle "Create a sequence of maps"
8. Select date range using the date pickers
9. Verify map count displays correctly
10. Download:
    * Click Download → "Download maps as PDF" → verify merged PDF with correct page count
    * Click Download → "Download maps as PNGs" → verify ZIP with correctly named PNG files

## Checklist - did you ...

Test your changes with

Screenshot 2025-12-20 at 2 53 50 PM

### DIFF
--- a/frontend/src/components/MapExport/MapExportLayout.tsx
+++ b/frontend/src/components/MapExport/MapExportLayout.tsx
@@ -444,7 +444,7 @@ function MapExportLayout({
 
   // The map content (title, legend, footer, map itself)
   const mapContent = (
-    <div className={classes.printContainer}>
+    <div ref={printRef} className={classes.printContainer}>
       {toggles.bottomLogoVisibility && bottomLogo && (
         <img
           style={{
@@ -649,11 +649,10 @@ function MapExportLayout({
       >
         {mapDimensions.width > 0 && mapDimensions.height > 0 && (
           <div
-            ref={printRef}
             style={{
               position: 'relative',
               zIndex: 3,
-              border: '1px solid #9E9E9E',
+              border: signalExportReady ? 'none' : '1px solid #9E9E9E',
               boxSizing: 'border-box',
               width: `${mapDimensions.width}px`,
               height: `${mapDimensions.height}px`,
--- a/frontend/src/components/NavBar/PrintImage/image.tsx
+++ b/frontend/src/components/NavBar/PrintImage/image.tsx
@@ -54,6 +54,34 @@ const debounceCallback = debounce((callback: any, ...args: any[]) => {
 
 const boundaryLayer = getBoundaryLayerSingleton();
 
+const UNSAFE_FILENAME_CHARS = new Set([
+  '<',
+  '>',
+  ':',
+  '"',
+  '/',
+  '\\',
+  '|',
+  '?',
+  '*',
+]);
+
+function sanitizeFilenamePart(input: string): string {
+  // Avoid control-character regexes (ESLint `no-control-regex`) by checking char codes.
+  const sanitized = input
+    .trim()
+    .split('')
+    .map(ch => {
+      const code = ch.charCodeAt(0);
+      const isControl = code < 32 || code === 127;
+      return isControl || UNSAFE_FILENAME_CHARS.has(ch) ? '_' : ch;
+    })
+    .join('');
+
+  // Collapse multiple underscores into a single underscore (e.g., "___" -> "_").
+  return sanitized.replace(/_+/g, '_').replace(/^_+|_+$/g, '');
+}
+
 function DownloadImage({ open, handleClose }: DownloadImageProps) {
   const { country, header } = appConfig;
   const logo = header?.logo;
@@ -376,7 +404,8 @@ function DownloadImage({ open, handleClose }: DownloadImageProps) {
         /(\s*-\s*)?\{(date|date_coverage)\}/gi,
         '',
       );
-      const filename = `${cleanedTitle}_${startDateStr}_to_${endDateStr}`;
+      const safeTitle = sanitizeFilenamePart(cleanedTitle);
+      const filename = `${safeTitle}_${startDateStr}_to_${endDateStr}`;
       // Server returns ZIP file when format is 'png'
       const contentType =
         format === 'pdf' ? 'application/pdf' : 'application/zip';


gislawill:
@ericboucher could we get your review on the architecture and server-code for this feature? We'll need to deploy the server updates for the preview build to work and actually demonstrate the functionality. But before I do that, I'd love to get your thoughts.

I'll note that one item that's still in flux is how we manage the aspect ratio for the generated map. That is all being handled in #1648 based on this conversation in Slack https://ovio-public.slack.com/archives/C0958FADX2M/p1765328289971239

api/prism_app/export_maps.py
```diff
@@ -0,0 +1,405 @@
...
+from playwright.async_api import Browser, Playwright
+from playwright.async_api import TimeoutError as PlaywrightTimeoutError
+from playwright.async_api import async_playwright
+from pypdf import PdfReader, PdfWriter
+
+from .models import AspectRatio, ExportFormat
+
+logger = logging.getLogger(__name__)
+
+# Timeouts
```
ericboucher: Should/could the playwright settings be shared with the report exports?

gislawill: I don't have a great sense of how the report exports work. Are they triggered by something in the browser or only in CI? I think the BROWSER_LAUNCH_ARGS should likely be shared. Not sure about the rest of the set up

api/prism_app/export_maps.py
```diff
@@ -0,0 +1,405 @@
...
+        return
+    if not EXPORT_ALLOWED_DOMAINS:
+        raise ValueError(
+            "No allowed domains configured. Contact administrator to add domains to EXPORT_ALLOWED_DOMAINS constant."
+        )
+
+    hostname_lower = hostname.lower()
+    is_allowed = False
+
+    for domain in EXPORT_ALLOWED_DOMAINS:
```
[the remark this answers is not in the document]
gislawill: Yea, good call. I'll see if I can find library that does this

wadhwamatic:
> > UI tweaks :
> > How about using a dropdown for the aspect ratio? I think it's taking too much space atm. Same for the sequence map, how about moving the explanatory text to an info tooltip or similar?
> 
> I can make those changes… in my view the tradeoff we’d make for less space is that any aspect ratio change will take two clicks instead of one, and we’d lose the ability to visually show the ratios themselves and group options — @wadhwamatic do you have a preference on a dropdown vs. the toggle groups?

I do see the value in having a visual representation of the aspect ratios. We could tighten up the layout to reduce its footprint. We could have auto, a4 portrait, a4 landscape and custom as a single row of buttons. The text 'Map resizes based on browser window' could be a tooltip on hover over auto.  I think that alone would make a big difference. The sequence text could also be a tooltip over the batch toggle.

gislawill:
@wadhwamatic, thanks for the thorough QA! I'll add some thoughts across the different areas of feedback

1. Data not loading before printing
We added tracking to prevent this in [895b7f5](https://github.com/WFP-VAM/prism-app/pull/1633/commits/895b7f58e9065f90f836e5573f535690d75b4a81). Frustratingly, this is consistently preventing the issue in Moz but not our other deployments. I'll dig into why that is and update.
    - [ ] Fixed (will check when done)

2. Varying types of layers (not just WMS) with date support
I support limiting v1 to just WMS layers. FWIW I believe Cadre Harmonise should be trivial to support (may already be supported). I'm not exactly sure about the issue with the kobo set-up - the username and password in aws (created Oct 2023) isn't working unfortunately. Would you be able to send me some test credentials in Slack? I'm guessing we need to pass the credentials back to our browser rendered version for batch maps - a good follow-up feature.

3. Date selection 
Conor is getting to work on those updates. All of those updates/fixes seem very do-able.
    - [ ] Updated (will check when done)

4. Date clarity
When you say:
    > This feature highlights the need to better display the selected date in the map view. What if we added date as a variable in the title?

    I'm assuming you're referring to the generated map export (but correct me if not). Yes, I agree we could add it to the title. For UX, I see a couple of options:
     * Automatically add date to whatever title the user provides `<h1>{title} - {date.toLocaleDateString()}`
     * Give user a date substitution variable, like...
        Screenshot 2025-12-23 at 1 32 10 PM

    For date formatting, we could use also use either YYYY-MM-DD or ([toLocaleDateString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString)). We could also support something more complex like [this templating language](https://day.js.org/docs/en/display/format). 

    Estimated level of effort for #1611: a couple hours. It seems pretty straightforward. Just let us know if you want us to pick this up.

wadhwamatic:
> * I support limiting v1 to just WMS layers. FWIW I believe Cadre Harmonise should be trivial to support (may already be supported). I'm not exactly sure about the issue with the kobo set-up - the username and password in aws (created Oct 2023) isn't working unfortunately. Would you be able to send me some test credentials in Slack? I'm guessing we need to pass the credentials back to our browser rendered version for batch maps - a good follow-up feature.

@gislawill - that's totally fine. let's limit to WMS layer for now as this is the real use case we have. Let's just disable the option for non-WMS layers. 

> I'm assuming you're referring to the generated map export (but correct me if not). Yes, I agree we could add it to the title. For UX, I see a couple of options:

Yes, that's what I meant, and I think that giving users a date substitution variable is the best approach. Let's use the existing date to locale functionality we have in date-utils.ts

> Estimated level of effort for https://github.com/WFP-VAM/prism-app/issues/1611: a couple hours. It seems pretty straightforward. Just let us know if you want us to pick this up.


Great, let's please address https://github.com/WFP-VAM/prism-app/issues/1611 while we're working on this feature.

api/prism_app/main.py
```diff
@@ -544,3 +546,54 @@ def get_google_floods_inundations_api(
...
+    Export maps for multiple dates using server-side rendering.
+
+    Accepts a URL with map parameters and a list of dates, renders maps using
+    Playwright, and returns either a merged PDF or ZIP archive of PNGs.
+    """
+    try:
+        dates = extract_dates_from_urls(export_request.urls)
+        file_bytes, content_type = await export_maps(
+            urls=export_request.urls,
+            viewport_width=export_request.viewportWidth,
```
ericboucher: Maybe worth having defaults for viewport_width / viewport_height (standard full screen size?) so that we can easily request a map export with just a list of urls?

How do we control zoom / movements done in the frontend ?

gislawill:
> There is a grey line around the maps when batch downloading. This line is not present when using the standard download feature.

Good catch @ericboucher! I'm finding I actually prefer the small gray border on the current batch maps over the borderless single map. I'm going to make sure that the single map has the same border for consistency. Curious for @wadhwamatic's thoughts on whether there's a better design path here?

| Current batch map border | Current single map border (lack of) |
| - | - |
| Screenshot 2026-01-12 at 10 05 44 AM | Screenshot 2026-01-12 at 10 05 53 AM |

frontend/src/components/NavBar/PrintImage/mapDimensionsUtils.ts
```diff
@@ -0,0 +1,118 @@
...
+import {
+  AspectRatio,
+  isAutoRatio,
+  isCustomRatio,
+} from 'components/MapExport/types';
+
+/**
+ * All available aspect ratios with their decimal values
+ */
+const ASPECT_RATIOS: {
```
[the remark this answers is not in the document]
gislawill: Good call, I've just created a new constants file that has a single source of truth for values + types

ericboucher:
@gislawill I am specifically talking about the border of the document, not the map itself

gislawill:
Removed document borders per thread in Slack (cc: @wadhwamatic & @ericboucher)

PR #1644, opened 2025-12-03
Batch maps submit

conor-jh:
### Description

This PR puts the basic pieces together so users can download a sequence of maps. There are a few things going on here:

- The export payload structure has been simplified: now, the UI simply passes a list of URLs with each `date` in the sequence encoded as a query string parameter
- Implemented a `PRISM_READY` global to ensure the map is loaded before Playwright captures it

## How to test the feature:

- [ ] Start the frontend app using `REACT_APP_LOCAL=true yarn start`
- [ ] Spin up the API server using `make api` (if you don’t have KOBO credentials, comment those lines out of `docker-compose.develop.yml`
- [ ] Select any date-enabled layer, choose a range of dates in the print config, and select **Download maps as PDF** and/or **Download maps as PNGs**

## Checklist - did you ...

Test your changes with

https://www.loom.com/share/6088a328befe49faa61b9c829a0c4f5b

### DIFF
--- a/api/prism_app/export_maps.py
+++ b/api/prism_app/export_maps.py
@@ -9,7 +9,9 @@ from datetime import datetime
 from typing import Final, Tuple
 from urllib.parse import parse_qs, urlparse
 
-from playwright.async_api import Browser, TimeoutError as PlaywrightTimeoutError, async_playwright
+from playwright.async_api import Browser
+from playwright.async_api import TimeoutError as PlaywrightTimeoutError
+from playwright.async_api import async_playwright
 from pypdf import PdfReader, PdfWriter
 
 from .models import AspectRatio, ExportFormat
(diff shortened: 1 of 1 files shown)


api/prism_app/export_maps.py
```diff
@@ -177,6 +197,7 @@ async def render_single_map(
             {"width": viewport_width, "height": viewport_height}
         )
         await page.emulate_media(media="screen")
+        url = url.replace("localhost:3000", "host.docker.internal:3000")
```
conor-jh: I couldn’t get things to work locally without this, but not sure it’s something we want to merge in…

frontend/src/components/MapView/Map/index.tsx
```diff
@@ -231,6 +231,11 @@ const MapComponent = memo(
           watchBoundaryChange(map);
         }
         trackLoadingLayers(map);
+
+        map.once('idle', () => {
+          // eslint-disable-next-line fp/no-mutation
+          (window as any).PRISM_READY = true;
+        });
```
gislawill: I don't think we need/want to set this in MapView, just ExportView

frontend/src/components/ExportView/index.tsx
```diff
@@ -60,6 +60,10 @@ const ExportView = memo(() => {
       setMeasuredFooterHeight(entry.contentRect.height),
     [],
   );
+  const onMapLoad = useCallback(() => {
+    // eslint-disable-next-line fp/no-mutation
+    (window as any).PRISM_READY = true;
+  }, []);
```
gislawill: I could be wrong, but I'm worried this will fire before all of our WMS tiles have loaded. I'm curious in your testing if it ever took a snapshot too early. We might need to wait for on idle rather than on load.

gislawill: Seems to work in my testing!

conor-jh:
Should be ready for re-review. I’m not quite sure what’s going on with the API linting failure—looks like a non-ignored vulnerability but AFAIK no recent changes update any dependencies…

PR #1657, opened 2025-12-16
Updating server browser management & ready state & fixing error message

gislawill:
### Description

This needs some cleanup but fixes the screenshot before all tiles loaded and improves server performance when loading a longer list.


gislawill:
CI failing due to security audit that Eric's already addressed.

PR #1697, opened 2026-01-14
Adding amplify to list allowed domains

gislawill:
### Description

This fixes @wadhwamatic issues with batch maps on amplify.
- [x] Allows amplify urls for batch map jobs
- [x] Extends timeout to accommodate slower page loads (not sure why they're slower)


gislawill:
CI should pass after #1691

PR #1699, opened 2026-01-15
Hide coverage in print view if AA flood or storm; COUNTRY=mozambique

wadhwamatic:
### Description

This fixes https://github.com/WFP-VAM/prism-app/issues/1698

Hides coverage from print view if either AA storm or AA flood layer is active. AA drought is ok as is.

## How to test the feature:

- [ ] REACT_APP_COUNTRY=mozambique
- [ ] Select AA flood or AA storm
- [ ] Print map
- [ ] Verify that coverage does not appear in the footer
- [ ] Select a layer with coverage (ex: 10-day rainfall anomaly)
- [ ] Print map
- [ ] Verify that coverage appears in the footer

## Checklist - did you ...

Test your changes with

### DIFF
--- a/api/poetry.lock
+++ b/api/poetry.lock
@@ -898,14 +898,14 @@ standard-no-fastapi-cloud-cli = ["email-validator (>=2.0.0)", "fastapi-cli[stand
 
 [[package]]
 name = "filelock"
-version = "3.20.1"
+version = "3.20.3"
 description = "A platform independent file lock."
 optional = false
 python-versions = ">=3.10"
 groups = ["main", "dev"]
 files = [
-    {file = "filelock-3.20.1-py3-none-any.whl", hash = "sha256:15d9e9a67306188a44baa72f569d2bfd803076269365fdea0934385da4dc361a"},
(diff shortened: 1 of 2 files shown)


gislawill:
@wadhwamatic could you foresee a scenario in which we have both AA flood or storm layers AND any additional layer with date? **If so,** I think we may show the incorrect coverage text: PRISM would show coverage text _for the additional layer_ though it may appear it's also the coverage text for the active AA layer. In this case, we'd want to either prevent the coverage text from showing or ensure that the coverage window text is prefixed with additional layer name.

wadhwamatic:
@gislawill - we don't show other layers alongside AA layers. I believe this was made explicit here https://github.com/WFP-VAM/prism-app/pull/1627. This might change in the future but not planned as of now. 

AA storm and AA flood are also unique in that they are forecasts and their coverage depends on a few factors. For storms, it depends on the issuer of the forecast and the lead time of the storm. AA floods in Mozambique currently describes the average flood risk over a 3 to 5 day window. But that window change by country. This complexity means that we'll probably keep these independent of other layers for the foreseeable future.

PR #1773, opened 2026-03-11
Initial batch maps cadence control

conor-jh:
### Description

This PR enhances the batch maps feature by giving users more fine-grained control over the interval at which maps are generated. Cadence options are determined by the selected layer's `coverage_window`:

* **10-day layers**: Every N dekads option only
* **Monthly layers**: Monthly option only
* **Quarterly layers**: Monthly and Quarterly options

#### Implementation notes
* Monthly and quarterly dates use the 21st of the month (last dekad), inferred from the layer's `coverage_window`
* Months/quarters with incomplete coverage windows are excluded from the generated sequence
* Quarterly labels display as e.g. `Oct - Dec 2025`
* Batch maps is disabled when no time-sequenced layer is selected, when a layer lacks `coverage_window`/`validity` config, or when more than one time-sequenced layer is active
* Date formatting is consistently localized across map title and footer
* Toast error messaging is scoped to the print dialog context only

#### Security fixes
* Bumped `pypdf` (6.9.1 → 6.9.2), `requests` (2.32.5 → 2.33.1), and `pygments` (2.19.2 → 2.20.0) in the API to resolve pip-audit CVEs
* Bumped `brace-expansion` (2.0.2 → 2.0.3 and 5.0.4 → 5.0.5) in alerting to resolve yarn audit CVE-2026-33750

## How to test the feature:

- [ ] Run the app locally for any deployment (e.g. `REACT_APP_COUNTRY=cambodia yarn start`)
- [ ] Open the print dialog, toggle the **Create a sequence of maps** option, and observe the cadence selection UI
- [ ] Verify that a 10-day layer only shows "Every N dekads", a monthly layer shows "Monthly", and a 3-month layer shows "Monthly" and "Quarterly"
- [ ] Verify that months/quarters without full coverage are excluded from the generated sequence
- [ ] Verify that the error toast does not appear when switching layers outside of the print dialog

> **NOTE:** Actually downloading the maps will fail unless you have the API server running, but export URLs should update accordingly

## Checklist - did you ...

Test your changes with


wadhwamatic:
Thanks for those updates @conor-jh. 

I need to have a think and ask some colleagues about what should happen when a user chooses to print monthly for a 10-day layer. Should it by default use the first dekad of the month? Is there a use case to use the 2nd or 3rd dekad?  Maybe the most recent available dekad available after the start date should determine which dekad to print for monthly? I'll let you know what I hear back. 

In the meantime, this didn't seem right as a different dekad is printed for Jan vs Feb and March. 
Screenshot 2026-03-18 at 21 20 45

When a user chooses to print monthly, we should just display the month-year to reduce confusion. And we should only include months where the entire coverage period is available and within the start/end range. In the example below, it should show: December 2025, January 2026, February 2026. And the underlying dates selected should be Dec 21 2025, January 21 2026, February 21 2026. The same logic would apply to quarterly. Hope that is clear!
Screenshot 2026-03-18 at 21 32 24

conor-jh:
Hey @wadhwamatic just pushed some changes that should address that feedback. Let me know if you find any additional issues. 

One callout:
> There are also layers without a time dimension, like pop_proj_2025 in Mozambique. Batch maps should not apply here

Batch maps won’t apply once the logic is uncommented. But I did update that commented out logic to exclude the other mentioned cases (lacking coverage window or > 2 time-sequenced layers).

Only to-do remaining is to add a tooltip for that 2-layer case.

conor-jh:
> @wadhwamatic the latest commits should address the date coverage issues, quarterly cadence labeling, and the exclusion of partially-covered dates.
> 
> One note: the map _preview_ is tied to the selected date prior to entering the print config. I’m not sure what users might expect here… architecturally it’s difficult to update since the date picker and map context are tightly coupled. Curious for your thoughts.

After working on https://github.com/WFP-VAM/prism-app/pull/1793, I’m thinking we might be able to provide a better UX here. What would you think about allowing users to actually click/select the query dates that we display based on the cadence?

That way a user could explicitly see the preview for say, Oct 2025, instead of being stuck with the latest date. Might provide a nice way to spot check what will actually get exported—thoughts? @wadhwamatic

ericboucher:
Screenshot 2026-04-01 at 9 17 23 PM

Nice work conor, mini question, could this fit on one line? Or maybe some texts are longer and wouldnt fit?

Issue #1760, opened 2026-03-10
[Feature Request]: Targeted Date Selection in Batch Map Range


gislawill:
### Provide a clear and concise description of what you want to happen.

When generating batch maps over a date range, users currently have no way to control the density of dates included — every available date is processed. For large date ranges this produces an unmanageable number of maps and makes it difficult to generate representative samples efficiently.

**Proposed approach:** Add a date interval selector to the batch map range configuration, allowing users to specify how frequently dates are sampled — e.g., once a month, once a quarter, or every N dekads.**Technical implementation details:**
- Interval options should include at minimum: monthly, quarterly, and every N dekads
- Should work in conjunction with the existing date range selection, not replace it

PR #1793, opened 2026-03-25
Batch maps layer selection

### LINKED ISSUE #1759: [Feature Request]: Batch Map Layer Selection in Print Menu

The batch map print menu currently lacks layer selection, meaning users cannot control which layers appear in generated maps. This limits the flexibility of batch outputs compared to the main chart interface, which already supports layer options.

**Proposed approach:** Add a layer selection dropdown to the print configuration menu, mirroring the existing layer options available in charts.**Technical implementation details:**
- Dropdown should follow the same UI pattern as the existing layer selection in chart configuration

_No response_

conor-jh:
### Description

This resolves #1759. The implementation allows users to select a different layer for batch maps directly from the print config. A few notes:

* Allowed layer options re-use the same logic from the Charts panel (per the issue description)
* In the current implementation, layer selection in the print config does *not* sync back to the primary map
* The coverage dates in the footer do not update with the layer selection (see comments)

## How to test the feature:

- [ ] Select a layer in the primary map
- [ ] Open the print dialog and enable batch maps
- [ ] Select a different layer from the dropdown and choose a date range
- [ ] Download as PDF or PNGs and observe the layer data on the exported maps

## Checklist - did you ...

Test your changes with

Screenshot 2026-03-24 at 7 42 12 PM


conor-jh:
I think the work here paves the way for an even better possible UX on #1760. Left more details here https://github.com/WFP-VAM/prism-app/pull/1773#issuecomment-4122411814

c-mac (commented):
Code looks pretty good! I did notice that "{date coverage}" is displayed in the title upon changing layers:

Image

frontend/src/components/NavBar/PrintImage/image.tsx
```diff
@@ -162,31 +161,48 @@ function DownloadImage({ open, handleClose }: DownloadImageProps) {
...
+        ? (selectableLayers.find(l => l.id === selectedLayerId) ?? null)
+        : null,
+    [selectedLayerId, selectableLayers],
+  );
+
   const shouldEnableBatchMaps =
     // selectedLayersWithDateSupport.length > 0 &&
     // selectedLayersWithDateSupport.every(layer => layer.type === 'wms');
-    false; // Temporarily disable batch maps
+    true; // Temporarily disable batch maps
```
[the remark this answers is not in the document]
conor-jh: Yeah exactly—but actually making the change allows folks to test things out on the staging deployments without having to pull the code down and make those adjustments.

conor-jh:
Thanks Amit! I believe I’ve addressed all the issues—thanks for the thorough review.

> 1. If I select a layer to print (global 3-month rainfall aggregate), select sequence of maps, change the layer (10-day NDVI), go back to the layer view (still correctly shows 3-month rainfall), select print again - the dialog still shows 10-day NDVI instead of resetting to the layer that I have selected in the main map view (3-month rainfall)

I think the recent update aligns better with expectations here… basically the primary map takes layer precedence so now if you close and re-open the modal it will default to whatever layer is selected in the primary map. One side effect of this logic is that it will default to the first available layer if no WMS layer is selected on the main map. 

I think that’s something we can/should address when we actually uncommented the batch maps enablement logic. I would expect in that case the feature just wouldn’t be available since an incompatible layer is selected. This just gets a bit more confusing IMO now that users can select layers from within the panel.

> 2. When I follow the same flow as above, the 10-day NDVI layer will show monthly and quarterly as cadence options, but should only show dekads

Should be fixed; let me know if you run into any other odd edge cases here.

> 3. When I toggle on print sequence of maps, the aspect ratio options get crunched together and on occasion a horizontal scrollbar will appear

I couldn’t exactly reproduce this but have a hunch on the root cause. If you’re still seeing it can you try to grab a screenshot/cast or provide more details on reproduction steps? 

> 4. I think we introduced a new issue here when [Remove 5-year date cap to address error in coverage window; COUNTRY=rbd #1781](https://github.com/WFP-VAM/prism-app/pull/1781) was merged earlier today. The calendar lookup in start and end dates for batch maps became sluggish

Should be resolved as well, and hopefully more performant than it was even before that merge.

wadhwamatic:
Great work on this @conor-jh - the layer selection / map preview / cadence / date selection components are working really well now.

I just noticed one small issue with layer selection. Some layers are configured in `shared/layers.json` but not specifically configured for a country (i.e. Mozambique) in `mozambique/prism.json`. Those layers shouldn't appear in the layer dropdown list for batch maps. A couple of examples of layers that shouldn't appear for Mozambique: `rainfall_agg_2month`, `population`. The available layers should be filtered based on the current country's `prism.json`.

I was also able to recreate the Aspect Ratio issue I mentioned. Here's a screenshot before selecting 'Create a sequence of maps', and one after. See that A4 Landscape gets squished.

Screenshot 2026-04-13 at 12 56 48

Screenshot 2026-04-13 at 12 56 59

wadhwamatic (commented):
Thanks for addressing those last two issues @conor-jh. This is GTG from my side

@ericboucher - can you also review please?

frontend/src/components/NavBar/PrintImage/image.tsx
```diff
@@ -175,47 +175,79 @@ function DownloadImage({ open, handleClose }: DownloadImageProps) {
 
   const [isDownloading, setIsDownloading] = useState(false);
 
+  const countryLayerIds = new Set(
+    Object.keys(configMap[safeCountry].rawLayers),
+  );
+  const selectableLayers = Object.values(LayerDefinitions).filter(
```
ericboucher, line 186: Make it clearer that only a subset of layers are available? Should we filter by "layers with dates" instead?

ericboucher (commented):
Code looks good!

One comment about available layers.

And a weird behavior currently when selecting a non date-enabled layer in the app before heading to print mode and toggling batch maps. The previously selected layers "persists" intermitently.

frontend/src/components/NavBar/PrintImage/image.tsx
ericboucher: cc @wadhwamatic

ericboucher:
Probably not introduced in this PR, but while we're at it...

Is it expected for the "quarterly" to be locked?
Screenshot 2026-04-25 at 7 03 21 PM

Can we find a better overlap styling for the closing cross? :)
Screenshot 2026-04-25 at 7 20 58 PM

conor-jh:
> Code looks good!
> 
> One comment about available layers.
> 
> And a weird behavior currently when selecting a non date-enabled layer in the app before heading to print mode and toggling batch maps. The previously selected layers "persists" intermitently.

Can you provide reproduction steps with specific layers? I bet this is due to only filtering out raster layers but not certain. For instance, I can see non-raster layers like population projection persist. Not sure what the expectation should be here…

> Is it expected for the "quarterly" to be locked?

Yes I believe so—GitHub is having some issues and I can’t get to it right now, but this logic was discussed in the cadence selector PR. 

> Can we find a better overlap styling for the closing cross? :)

Interesting—I can’t reproduce this but bumped the z-index. If you’re able to reproduce it can you let me know what your browser resolution is or any other parameters that allow for the overlap?

frontend/src/components/NavBar/PrintImage/image.tsx
conor-jh: Yeah I’ll defer to @wadhwamatic here…

conor-jh:
@wadhwamatic / @ericboucher also reverted the change to hard-code batch maps as enabled. To test you'll need to swap it back locally. Otherwise hopefully this is looking okay to merge.

frontend/src/components/NavBar/PrintImage/image.tsx
wadhwamatic: Are we asking if we should add a label to the layer selection list? If yes, we can change the label from "Layer" to "Select a layer with dates".  A tooltip could be more informative: "Only layers with date attributes are available for this feature".

wadhwamatic:
> Can you provide reproduction steps with specific layers? I bet this is due to only filtering out raster layers but not certain. For instance, I can see non-raster layers like population projection persist. Not sure what the expectation should be here…

I don't think we should allow batch maps to include a date layer AND a static layer. Odd things could happen unintentionally. But before taking action, let me check with the team to see if this is needed.

wadhwamatic:
@ericboucher - when we chatted earlier, you were asking about two date layers. I double checked and this is handled. Users will see a message in the tooltip: "Select one layer at a time to create a sequence of maps"

Screenshot 2026-04-27 at 17 23 04

wadhwamatic:
> > Can you provide reproduction steps with specific layers? I bet this is due to only filtering out raster layers but not certain. For instance, I can see non-raster layers like population projection persist. Not sure what the expectation should be here…
> 
> I don't think we should allow batch maps to include a date layer AND a static layer. Odd things could happen unintentionally. But before taking action, let me check with the team to see if this is needed.

@conor-jh - I confirmed that we don't need to show additional layers for batch maps. I see some minor changes needed for this.

If a user has a date layer and non-date layer selected, creating a sequence of maps should: 
1. Remove the non-date layer from the print preview and subsequent maps; 
2. Notify the user through a toast message that only the selected date layer will be included in the batch maps. ex: "The sequence of maps will only display the `layer-title` layer"; 
3. Display all previously selected layers (date + non-date) after exiting print preview.

There's also a small bug when going from print view, to main app view to add a 2nd date layer, and then back to print view. To recreate it:
1. country=cambodia
2. Select a valid date layer (10-day rainfall)
3. Click print, then toggle on 'create a sequence of maps' 
4. Exit print view
5. Add a second date layer (EWS-1294), click print
6. See that the toggle is correctly disabled, but the batch map UI remains in view, including the download button

Screenshot 2026-04-29 at 08 17 33

These should be the last changes! Thanks for the persistence on this one

conor-jh:
Thanks @wadhwamatic! Latest commits should resolve those issues. I wasn’t able to fully test the Cambodia scenario as I don’t think I have access to view the EWS layer, so let me know if anything does seem off and I can iterate on the fix as needed. 

Really appreciate the thorough reviews—there’s a lot of nuance here and I’m learning a lot about how all the layers work together :)

@ericboucher you may want to look at the layer filtering logic introduced in [590925f](https://github.com/WFP-VAM/prism-app/pull/1793/commits/590925f706aa7d4001c660e616d95d508adc84ed). 

Once everyone’s good I’ll push a final commit to flip the batch maps feature back to disabled and merge.

wadhwamatic:
Hey @conor-jh. Sorry this PR continues as bit longer as I found some more issues. The first two issues stem from `boundary` layers and the level to load based on the `admin_level_data` layer vs the default boundaries. We should use default boundaries when the `admin_level_data` layer is removed for a batch map. Layers with the type `admin_level_data` have an `admin_level` attribute that determines which boundary layer to use (see layers with type `boundary`)

1. If I select a non-date layer and date layer, then select batch maps, the boundaries should revert to the default. Right now, it seems to either remove boundaries altogether, or use the boundary of the now de-selected non-date layer. To see this issue in Mozambique, try Vulnerability > Poverty > Multidimensional Poverty Index, combined with any rainfall layer. See the boundaries in this screenshot are at admin 1 from the poverty layer which was removed from the map

image

2. I think this is new. Whenever I select batch maps, the boundary layer seems to lose it's order and becomes 'covered' by the date layer. You can see this in the map above and below.
3. The downloaded batch maps show the non-date layer in the legend even though it's not in the map. See the file below, and the IPC layer is in the legend, but not in the map

map_2026-01-21

4. The logic to allow batch maps to be enabled should be the same as what determines which layers to show in the layer selection dropdown in batch maps. I.e., date layers with coverage or validity configured. An example of a date layer that should **_not_** enable batch maps is Cambodia’s EWS-1294 data layer. It’s under Flood > Early warning and is accessible without credentials

conor-jh:
Couple of things in the latest push:

- I’ve uncommented the logic that determines if batch maps should be enabled, so that should help with some edge cases since the toggle itself will be suppressed in invalid scenarios. As a recap, that logic is currently:

> At least one layer with date support is selected
> Every selected layer is of type `wms` and has a coverage window or validity property

I’m thinking we *may* want to revisit that since users can now select any valid layer from within the print dialog, but ideally as a followup item…

- I’ve also revisited the way we’re rendering layers in the preview. Now we’ll strip *all* non-WMS layers, including boundary layers, from the primary map and “rebuild” the map preview with the React layer rendering engine. This seems to be the cleanest approach… but please do stress test and let me know if you find any odd behavior.

ericboucher:
@copilot instead of looking at "wms" layers, do we have a function to determine if a layer is "date-enabled" that we could extract and re-use? Investigate and open a PR if you see a way.

conor-jh:
Chatted with @gislawill a bit more about the layer selection logic. 

I think some of the edge cases we’ve been working through are impossible to achieve once the base logic for displaying vs. hiding the batch maps toggle is allowed to run. This should hopefully simplify a lot of the logic, including the need to disable the toggle and show a tooltip for multiple date layer scenarios.

For reference, that logic is:

> At least one layer with date support is selected
> Every selected layer is of type wms and has a coverage window or validity property

I think at this point I’d like to get this PR merged and if we need to revisit the display logic we do so as a separate effort.

cc: @wadhwamatic / @ericboucher — thoughts? If that makes sense to you (and testing validates the assumption above) I will go ahead and re-comment the logic out and try to merge.

wadhwamatic:
> @copilot instead of looking at "wms" layers, do we have a function to determine if a layer is "date-enabled" that we could extract and re-use? Investigate and open a PR if you see a way.

@ericboucher - @conor-jh and I had discussed using `coverage_window` and `validity` from the layer config to determine if a layer is date-enabled. This wouldn't cover the ch data layers in RBD that have `start_date` & `end_date`, but those layers are not needed for batch maps. I thought this is already implemented!

wadhwamatic:
@conor-jh - thanks for the latest updates. I took another pass at it, and the logic to show the batch map toggle works well, as does removing non-date layers (including from the legend) in the output maps.

The only issue I see still is that the admin boundary layer appears to be covered when a non-date layer has been removed. In the example below, I selected CHIRPS 10-day rainfall and IPC phase classification layers (vulnerability > food insecurity). When I enabled the batch map, the IPC layer correctly is removed in the preview and the admin boundary layer looks correct. But in the exported png, the admin boundary layer is behind the rainfall layer.

Here's the print preview in PRISM
Screenshot 2026-05-06 at 16 06 17

And the exported png
map_2026-04-21

conor-jh:
Thanks for checking it out @wadhwamatic. I noticed this too and meant to mention it. Looks like the issue exists on the master branch—this PR did not introduce it. Thinking if it’s alright with you we can open a separate issue to handle the fix?

wadhwamatic:
> Thanks for checking it out @wadhwamatic. I noticed this too and meant to mention it. Looks like the issue exists on the master branch—this PR did not introduce it. Thinking if it’s alright with you we can open a separate issue to handle the fix?

ah ok. I don't see how I can recreate that on main at the moment as the logic to remove the admin-level data layer is within this PR only. Please go ahead and create that issue when you have a chance. And agreed that we can move forward with this PR.

ericboucher (approved):

ericboucher:
@wadhwamatic I fixed the boundary order error which was due to how we did not reorg layer order in the export view.

Order should now be correct here - https://staging-prism-frontend--1793-lfzx0dll.web.app/export/?hazardLayerIds=ndvi_dekad&date=2026-04-30

ericboucher:
This PR is ready to merge imho. Two things to note :
- I am a bit unsure though how valuable it is to completely decorelate the map layers from the batch map selected layers. It is adding quite a bit of complexity to the code.

- The selection was not working as expected because of the "lazy loading" of dates and the way we filter out layers without dates. The easiet was to re-add the coverageWindow || validity condition, hopeffuly we can find something better in the future

PR #1810, opened 2026-04-08
Dashboard fix - empty json init

c-mac:
### Description

This fixes a small issue pertinent to moving dashboard configs into S3. 
Right now, if a `dashboard.json` file does not exist for a country in S3 (or locally in the public/data/<country> path) the user is shown an "invalid JSON" error toast. This is a surprisingly tricky issue to solve, and I found that simply ensuring that we have empty json files for each country is the easiest solution. As mentioned from Eric's comment on the main issue, we'll likely move away from S3 fetches for these configs at some point in the future.

## How to test the feature:

- [ ] Test the app with two different countries: one with a dashboard config, one without. The one with should work as expected, and the one without should not show an error toast.

## Checklist - did you ...

Test your changes with


ericboucher:
Is that something that could happen on build and create an empty json file if no S3 file is available?
@wadhwamatic could it make sense to have a flag or a path in prism.json to activate/deactivate dashboards? We wouldnt need this logic and empty files if we do.

wadhwamatic:
> Is that something that could happen on build and create an empty json file if no S3 file is available? @wadhwamatic could it make sense to have a flag or a path in prism.json to activate/deactivate dashboards? We wouldnt need this logic and empty files if we do.

@ericboucher - @c-mac and I talked about this briefly. It would be good to have a mechanism to activate / deactivate dashboards so that we don't the need for an empty json file. But we also don't want to have to redeploy a react build if a dashboard is created via the upcoming UI @conor-jh is working on. Do you have a suggestion on how to achieve this with the upcoming approach? 

FYI - on a call we had yesterday we agreed that I'd write up the ideal state UX for creating and publishing dashboards beyond this current scope of work. I'll share that by tomorrow.

wadhwamatic:
here's a write up of the ideal UX end state https://github.com/WFP-VAM/prism-app/issues/1815

PR #1828, opened 2026-04-16
Fix dashboard "invalid JSON" toast caused by SPA 200-HTML rewrite

ericboucher:
### Description

Alternative fix for the "invalid JSON" toast seen when a country has no `dashboard.json` (the problem #1810 addresses by committing 30 empty `[]` files).

**Root cause.** `frontend/firebase.json` uses a catch-all SPA rewrite — `{ "source": "**", "destination": "/index.html" }` — so a missing `/data/{country}/dashboard.json` is served as `index.html` with **status 200**. The existing not-found guard in `dashboardConfigQueryError.ts` only catches HTTP 404, so HTML reaches `JSON.parse`, throws, and the user sees the "Dashboard configuration is not valid JSON" notification. The same pattern would hit any S3 website endpoint or static host that rewrites on 404.

**Fix.** After the `response.ok` check in `fetchDashboardConfig.ts`, inspect `content-type`. If it's not JSON, throw the same "http 404" signal that `useDashboardConfig` already swallows silently. Real JSON corruption still surfaces (those responses advertise `application/json`).

### Trade-offs vs. committing empty per-country files

- No 30-file repo churn; no need to maintain an empty file per new country.
- Dashboard enablement stays runtime-driven, which leaves room for @conor-jh's upcoming UI-driven dashboard creation without a rebuild per enablement (see the #1810 discussion on a `prism.json` flag).
- Covers S3-URL deployments too, not just the local `public/data` path.

### How to test

- [ ] `REACT_APP_COUNTRY=cambodia yarn start` — country without a dashboard: no error toast, no Dashboard nav item.
- [ ] `REACT_APP_COUNTRY=mozambique yarn start` with a local `public/data/mozambique/dashboard.json` — dashboards load as before.
- [ ] Simulate Firebase fallback by pointing `REACT_APP_DASHBOARD_CONFIG_BUCKET_URL` at an origin that serves HTML for missing files — toast should not appear.
- [x] `yarn test fetchDashboardConfig` — new test covers the 200-plus-HTML path.

### Checklist

- [x] Add / update tests
- [x] Add / update code comments explaining the SPA-rewrite failure mode

https://claude.ai/code/session_01WMhiv3eU972baZyw2gMgCZ

### DIFF
--- a/frontend/src/dashboardConfig/fetchDashboardConfig.test.ts
+++ b/frontend/src/dashboardConfig/fetchDashboardConfig.test.ts
@@ -15,11 +15,14 @@ describe('fetchDashboardConfig', () => {
     global.fetch = originalFetch;
   });
 
+  const jsonHeaders = new Headers({ 'content-type': 'application/json' });
+
   it('returns validated dashboards on success', async () => {
     global.fetch = jest.fn().mockResolvedValue({
       ok: true,
       status: 200,
       statusText: 'OK',
+      headers: jsonHeaders,
       json: async () => validBody,
     } as unknown as Response);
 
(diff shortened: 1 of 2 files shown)


wadhwamatic:
Thanks @ericboucher. I tested the following on localhost, and worked well:
1. No dashboard configured. Toast message does not appear
2. Remote dashboard configured via env variable (REACT_APP_DASHBOARD_CONFIG_BUCKET_URL) showed as expected
3. Local dashboard added to public/data/country/dashboard.json showed correctly

@c-mac - can you have a look?

c-mac (approved):
This is great!

Issue #1830, opened 2026-04-16
[Feature Request]: PRISM user roles and permissions


wadhwamatic:
### Provide a clear and concise description of what you want to happen.

# PRISM user roles and permissions (draft for discussion)

## Summary

This issue proposes a minimal v1 role and permission model for PRISM to unblock current work on authentication, dashboard editing, and gated deployments. The goal is a model simple enough to implement now, while remaining compatible with broader WFP identity and access work as it matures.

## Proposed roles

**Admin** — full access to PRISM admin and configuration workflows. This will be limited to just Amit for now

**Editor** — can author and update dashboards, but cannot manage users or system-wide settings. There would be a handful of these users in WFP HQ, and one per country office

**Viewer** — can access gated content but cannot modify it

**Public / Unauthenticated** — access to non-gated content only.

## Permission matrix

| Capability | Admin | Editor | Viewer | Public |
| --- | --- | --- | --- | --- |
| Access backend admin | ✅ | ❌ | ❌ | ❌ |
| Create dashboard config | ✅ | ✅ | ❌ | ❌ |
| Edit dashboard config | ✅ | ✅ | ❌ | ❌ |
| Publish changes | ✅ | ✅ | ❌ | ❌ |
| View gated content | ✅ | ✅ | ✅ | ❌ |
| Manage deployment settings | ✅ | ❌ | ❌ | ❌ |
| Manage users/roles | ✅ | ❌ | ❌ | ❌ |

## Deployment-scoped permissions

Roles are scoped per country/deployment — e.g. a user who is Editor for Myanmar only, or Viewer for one restricted deployment but not others.  

## Gated content

PRISM content is non-gated by default. 

As of now, we have two types of gated content. In Myanmar, the entire application is gated. WFP SSO allows access via OAuth. This same capability should exist via CIAM. 

We also have layer-level gating in the case of Cambodia's Kobo data layers. @ericboucher - I need a reminder on how those credentials were established. Did we just copy over DMP? Ideally this can migrate to the new CIAM approach

## Open questions

- With the exception of Admin, should HQ Editors be able to create / edit country-level content?
- Do we need deployment-scoped permissions in v1, or can that come later?
- Do roles live in CIAM, or in an application-layer system PRISM controls?
- Which permissions are needed for the first dashboard editing workflow specifically?

## Next steps

- [ ] Confirm proposed roles with Cam against current implementation needs
- [ ] Resolve open questions above before or during the next dev sync
[...]

Issue #1815, opened 2026-04-10
[Feature Request]: Ideal UX end state for dashboards


wadhwamatic:
#### Ideal UX end state for dashboards
Dashboards in PRISM are curated, prebuilt views of geospatial and analytical content. The ideal end state introduces a clear separation between two user roles: editors (admin users) and viewers (regular users), and a structured lifecycle for dashboard content: create → draft / review → publish.

**Editors (Admin Users)**
Editors will have full control over the dashboard lifecycle, entirely within the PRISM UI with no file manipulation, code changes, or React build & deployment required. 

To create a dashboard, an editor initiates a "Create new dashboard" flow, selects a supported layout, and begins configuring the view. At any point, the editor can save the work as a draft. Drafts are stored server-side but remain invisible to regular users, allowing safe iteration before anything goes live. Editors can also import an existing dashboard configuration by dragging and dropping a previously exported JSON file. In this case, the UI validates the file inline and surfaces any errors clearly before loading it.

When a dashboard is ready to share, the editor can export the configuration as a JSON file (for portability or backup) and publish it through the admin console, which handles storage and access control. Publishing makes the dashboard visible to all users of the relevant deployment. Editors can update or un-publish dashboards at any time, reverting them to draft state when revisions are needed.

Throughout this workflow, the UI communicates state clearly, validation errors surface inline before saving, import failures return the UI to a stable state with a clear message, and success states confirm when a dashboard is published.

**Viewers (Regular Users)**
Viewers see only dashboards that have been explicitly published for a country deployment. There is no visibility into drafts, imports, or the authoring process. Their experience is stable and curated: they open PRISM, see the available dashboards, and explore the content.

**Why it matters**
This model removes the dependency on code changes and redeployments for routine dashboard updates. Non-developer staff can create, iterate on, and publish monitoring content independently, making the dashboard system genuinely operational rather than developer-gated.

wadhwamatic:
@c-mac @conor-jh @gislawill @ericboucher @lowriech - I wrote up the above to clarify how admins and users would experience dashboards. Please ask any questions you have here!

c-mac:
Does an editor have to specify which country the dashboard belongs to? Does the dashboard only ever belong to one country?

Your proposal mentions that viewers can only see published dashboard. I thought there might be an additional filter here, which is that certain users should only see published dashboards for certain countries, too. Is that a correct assumption?

[...]

Issue #1767, opened 2026-03-10
[Feature Request]: Admin Console (Dashboard-specific Admin)


gislawill:
### Problem

There is currently no interface for managing dashboard configurations in PRISM. Publishing or updating a dashboard requires manual file handling or code changes, which creates a bottleneck and limits who can maintain dashboard content. Administrators need a straightforward way to upload, update, and control access to dashboards without touching the underlying codebase.

### Proposed Approach

Add a Dashboard CRUD view to the Admin Console that allows authenticated administrators to manage dashboard configurations stored in the database:
- Browse dashboards with visibility, deployment association, and access settings
- Upload dashboard JSON files (exported from the PRISM UI) via drag-and-drop to create or overwrite configurations
- Update dashboard title, visibility, and deployment settings
- Dashboards in "draft" state are not visible in PRISM — only saved for future reference
- Dashboard editing UI remains in the main PRISM application; the Admin Console handles storage and access control

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

**Technical implementation details:**
- Dashboard configs stored as JSONB in Postgres (primary update mechanism is overwrite/replace)
- Admin Console view: browse by title, visibility, deployment; permission-scoped (see own vs all dashboards based on role)
- Drag-and-drop JSON upload to create or overwrite a dashboard config record
- Draft state: configs saved but not surfaced in PRISM UI until published
- Depends on Admin Console Architecture and CIAM Integration for authenticated access
- This is the server-side complement to the Dashboard Authoring & Editing UX issue (which handles the PRISM UI side)

gislawill:
I believe we can call this complete. Great work @c-mac and @conor-jh!

PR #1843, opened 2026-04-28
Dashboard config moved to admin

c-mac:
### Description

Implements #1767 and #1830 and #1815.

- Admin can now create and publish dashboards on a per-deployment basis.
- Adds `deployments` table, which mirrors the list of safe countries defined in the frontend config.
- Removes code and scripts related to dashboard configs living in S3.

## How to test the feature:

- `cd api && poetry run alembic upgrade head` (to create the new `dashboard` table)
- Navigate to `/admin` and click "Dashboards" in the left panel
- Create a new dashboard model, can use [this](https://github.com/WFP-VAM/prism-app/edit/dashboards-demo/frontend/src/config/mozambique/dashboards.json?pr=/WFP-VAM/prism-app/pull/1661) as reference
- Choose a deployment (I recommend mozambique, since that's what the example data uses) and set deployment status to "published"
- For whichever deployment you used, start the react app with the same deployment
- You should be able to view and interact with the dashboard that you just created

## Checklist - did you ...

Test your changes with

https://www.loom.com/share/56c8434aef0c4c8f983d0b1e7a1c3a72
Screenshot 2026-05-14 at 9 08 34 AM


api/alembic/versions/add_dashboard_table.py
```diff
@@ -0,0 +1,99 @@
...
+        sa.CheckConstraint(
+            "deployment IS NULL OR deployment IN "
+            "('afghanistan', 'bhutan', 'cambodia', 'cameroon', 'colombia', 'cuba', "
+            "'ecuador', 'global', 'haiti', 'indonesia', 'jordan', 'kyrgyzstan', "
+            "'malawi', 'mongolia', 'mozambique', 'myanmar', 'namibia', 'nepal', "
+            "'nigeria', 'rbd', 'sierraleone', 'somalia', 'southsudan', 'srilanka', "
+            "'sudan', 'tajikistan', 'tanzania', 'ukraine', 'zambia', 'zimbabwe')",
+            name="ck_dashboard_deployment_allowed_values",
+        ),
+    )
```
gislawill: I like validation for valid deployments but I think the pg enum validation may be a bigger maintenance pain than it's worth. With this validation, we're need a migration recreating the enum with every new deployment. At that rate, we would be better off with a deployment table with a foreign key ref. 

@c-mac, acknowledging we could use a deployment table in the AA droughts work and probably other features, how would you feel about adding that deployments table? Alternatively, I'd suggest sticking with python-based validation and just setting this as a non-nullable string.

api/prism_app/dashboard_slug.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def slugify_dashboard_name(name: str) -> str:
+    """Lowercase kebab-case slug; stable per country when paired with unique `name`."""
+    s = (name or "").strip().lower()
+    if not s:
+        return "untitled-dashboard"
+    s = re.sub(r"[^a-z0-9]+", "-", s)
+    s = s.strip("-")
+    return s or "untitled-dashboard"
```
gislawill: I think we should throw an error instead of falling back to "untitled-dashboard" based on the unique db constraint

api/alembic/versions/add_dashboard_table.py
c-mac: Good call. I like the `deployment` table idea. Makes it nice and easy for admins to edit the deployment list, too.

api/prism_app/database/dashboard_model.py
```diff
@@ -0,0 +1,132 @@
...
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.dialects.postgresql import UUID as PG_UUID
+from sqlmodel import Field, SQLModel
+
+
+class DashboardStatus(str, Enum):
+    """Publication state: drafts are not served to the PRISM UI read API."""
+
+    draft = "draft"
+    published = "published"
```
gislawill, line 20: What would you think about adding an "archived" status?

conor-jh: I don’t know if this is necessary, but do we care about differentiating between “unpublished” drafts vs. a dashboard someone is updating? 

I think the implication here is if we want to support admins making dashboard changes—I assume we may still want those to be “live” to end users but to also provide an explicit “publish changes” action to the updater.

ericboucher:
Could it make sense to rebase on the CIAM PR for readability?

gislawill:
> Could it make sense to rebase on the CIAM PR for readability?

@ericboucher, there are good reasons to rebase (migration order, readability, proper permission gating) but I don't think rebasing will help with readability. I'd like to get your review on the CIAM PR first to make sure it's in a stable place before rebasing so we don't create wasted effort.

api/prism_app/dashboard/dashboard_config_validation.py
gislawill, line 1: I think this level of validation might be overkill. Feels like it's going to be duplicating a lot of the validation on the front end, which is where that messaging should belong in my opinion. 

@conor-jh curious for your thoughts on this validation & the UX here

c-mac: Understandable, I was trying to avoid a case where an admin could put in bad data, thus causing errors when trying to load dashboard config on the frontend. 

We could drop validation entirely from both this schema file and at the database table level to fully trust the frontend for all validation, so long as frontend is robust enough for error handling.

api/README.md
```diff
@@ -207,14 +187,15 @@ To deploy the application, we need to ensure the environment variables are set i
...
-- The alerts PostgreSQL schema is current: from `api/`, run `poetry run alembic upgrade head` with `PRISM_ALERTS_DATABASE_URL` set (see **Alerts database migrations (Alembic)** above; use `alembic stamp` if the database predates Alembic and already has the tables).
+- The alerts PostgreSQL schema is current: from `api/`, run `poetry run alembic upgrade head` 
 
 To deploy, ssh into the EC2 instance:
 - Get the private key and copy it to `~/.ssh/{some name}.pem`
 - Add your IP address to the EC2 instance's whitelist in the AWS console
 - Run `ssh -i ~/.ssh/{some name}.pem ubuntu@{Public IPv4 DNS}`
 - Navigate to the api directory
 - Confirm you're on the right branch and the branch is up to date
+- Optional: if any database migrations are in the deploying branch, open a bash shell to the API container. Use the `poetry run alembic heads` and `poetry run alembic current` commands first to ensure the proper migration will be applied, and once verified, run `poetry run alembic upgrade head` to run the migration.
```
c-mac, line 198: @gislawill Here are the consolidate docs changes we chatted about in our call

api/prism_app/main.py
```diff
@@ -157,6 +158,39 @@ def healthcheck() -> str:
     return "All good!"
 
 
+@app.get(
+    "/dashboards",
+    responses={503: {"description": "Dashboard database unavailable"}},
+    summary="Published dashboard configs (deployment-scoped)",
+)
+def get_published_dashboards(
+    country: str = Query(
```
gislawill, line 169: Let's name this `deployment` not country

api/prism_app/dashboard_slug.py
gislawill, line 1: Let's create a dashboard directory in prism_app and put this in a utils file there. Still trying to avoid file bloat in the prism_app/ folder

gislawill:
Screenshot 2026-05-11 at 9 35 14 PM
Let's just have this say, "New Dashboard" (no Model)

api/prism_app/admin.py
```diff
@@ -100,6 +108,8 @@ def register_alerts_admin_views(admin: Admin) -> None:
     admin.add_view(AlertView(AlertModel))
     admin.add_view(KoboUserView(KoboUser))
     admin.add_view(AnticipatoryActionAlertsView(AnticipatoryActionAlerts))
+    admin.add_view(DeploymentView(DeploymentModel))
```
gislawill: I don't think this is a useful view. Let's cut it

gislawill:
We need the ability to drag-and-drop the json here. Readability (though nice) will be less important and json portability. I think this code editor block will add more confusion than utility. A simple drag and drop file upload would be easier for most users (we don't want users manually editing their json configs)
Screenshot 2026-05-11 at 9 38 33 PM

gislawill:
I think I may have found a bug when creating new dashboards. Could you look into this @c-mac?
https://www.loom.com/share/783b043687f244298d1c2cdf6c254944

gislawill (changes requested):
Thanks Cam! I've got a few requested changes in the proceeding comments. Let me know if you have any questions

c-mac:
@gislawill regarding the bug video, I had that issue too, but it is no longer relevant as I've replaced it with a drag and drop field that now has a clearer error message if the uploaded JSON is invalid.

api/alembic/versions/add_dashboard_table.py
```diff
@@ -0,0 +1,135 @@
...
+        "published",
+        name="dashboard_status_enum",
+        create_type=False,
+    )
+
+    op.create_table(
+        "dashboard",
+        sa.Column("id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("title", sa.String(), nullable=False),
+        sa.Column("slug", sa.String(), nullable=False),
```
conor-jh: Consider renaming to `path` to match the schema?

frontend/src/hooks/useDashboardConfig.ts
```diff
@@ -55,7 +54,7 @@ export function useDashboardConfig(): void {
           if (cancelled) {
             return;
           }
-          // No dashboard.json for this instance is expected: empty list hides the nav link.
+          // Missing config (e.g. 404) is treated as no dashboards: nav link hidden.
```
conor-jh, line 56: The logic for hiding the nav link is changing with some of the UX flows (options to create a dashboard in the UI and import via JSON live under that menu). 

We can discuss with @wadhwamatic if this should be the case, but right now those options will show up even if there are zero pre-configured dashboards available.

api/prism_app/dashboard/dashboard_config_validation.py
```diff
@@ -0,0 +1,154 @@
...
+    # Keep output shape aligned with frontend Zod optional fields:
+    # omit unset/None values instead of emitting explicit JSON nulls.
+    return config.model_dump(mode="json", exclude_none=True)
+
+
+def validate_and_dump_dashboard_config(raw: Any) -> dict[str, Any]:
+    """
+    Validate dashboard layout JSON for the ``dashboard.config`` column.
+    Only the dashboard columns are persisted in config; model fields such as
+    title, deployment, and is_editable are validated separately.
```
conor-jh: What’s the thought process behind validating `title` and other fields separately? `deployment` seems reasonable, but why remove the other fields from the raw config?

api/prism_app/dashboard/dashboard_config_validation.py
conor-jh: Hmm, I concur it’s probably good to ensure admins can’t save malformed configs via the console, so _some_ validation here is probably good. But the frontend will already need to validate configs against the schema to account for JSON imports and dashboards created via the UI. 

Honestly looking through it I probably would favor a call. Now that all three methods of adding dashboards are implemented (even as drafts) it might become pretty clear/easy on where to validate.

api/prism_app/database/dashboard_model.py
c-mac: Talked to Amit about this and we agreed to add an "archived" status.

c-mac:
@wadhwamatic 's feedback has been incorporated, and @conor-jh and I decided it is best to let frontend handle the dashboard config specific logic. Now, the backend only validates whether or not the JSON itself is valid, rather than runtime type checking (which frontend owns in `schema.ts`). Implementation details [here](https://github.com/WFP-VAM/prism-app/pull/1843/changes/6a360698a3a6fc8b71ef5fe7a23b14407084052b).

gislawill (approved):
This is looking good to me, thanks @c-mac! Glad to hear @wadhwamatic had a review and gave feedback. Now over to @ericboucher

api/alembic/versions/add_dashboard_table.py
```diff
@@ -0,0 +1,132 @@
...
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "add_dashboard_table"
+down_revision = "prism_users_permissions"
+branch_labels = None
+depends_on = None
+
+# Source of values: frontend/src/config/index.ts -> `configMap` keys.
+_DEPLOYMENT_CODES = (
```
ericboucher: How about making thisn an enum instead of a table?

And I feel like we should call this country for now rather than deployment to be consistent with the rest of the app.

See the comments here https://github.com/WFP-VAM/prism-app/pull/1861#discussion_r3230895047

api/prism_app/tests/test_published_dashboards_api.py
```diff
@@ -0,0 +1,63 @@
+"""API tests for GET /dashboards (published, deployment-scoped)."""
+
+import os
+from unittest.mock import patch
+
+# prism_app.kobo is imported by main and raises if these are unset
```
ericboucher: Might be worth loading this for all tests ?

frontend/src/config/index.ts
```diff
@@ -110,22 +109,14 @@ const {
 } = shared;
 
 // Perform deep merges between shared and country-specific configurations
-// Dashboard definitions: from S3 when REACT_APP_DASHBOARD_CONFIG_BUCKET_URL is set; otherwise
-// from public/data/{country}/dashboard.json (see useDashboardConfig).
+// Dashboard row definitions: loaded from the geospatial API
+// (GET /dashboards?deployment=<safeCountry>&status=published; see useDashboardConfig).
```
ericboucher: Can the endpoint load anything else than published ? Shouldn't it be the default for this endpoint ? :)

frontend/src/hooks/useDashboardConfig.ts
```diff
@@ -13,25 +9,28 @@ import {
...
   const enabled = isAuthenticated || !authRequired;
-  const url = dashboardConfigUrl ?? localDashboardConfigUrl;
+  const url = `${DASHBOARDS_API_URL}?${new URLSearchParams({
+    deployment: safeCountry,
+    status: 'published',
+  }).toString()}`;
 
   useEffect(() => {
-    if (!enabled || !url) {
+    if (!enabled) {
```
ericboucher, line 32: Should we just rely on having published dashboards in the future ?
cc @wadhwamatic

ericboucher (approved):

ericboucher:
Looks good to me, nice work.

Blocking changes :
Let's change voc from deployment to country to be consistent. And I don't see the value of country table for now, but maybe I'm missing something?

And a few non blocking suggestions in line

PR #1844, opened 2026-04-30
Direct from COG POC

gislawill:
### Description

This PR contains all the functionality needed to load all of our current WMS layers directly from COG. Those files are already hosted in S3. 

We add Deck.gl interwoven into MapLibre. Then we rely on the PRISM server to create pre-signed short-lived URLs that the app can use to make requests to S3 directly. We currently use our server as a proxy to hit COG layers without CORS issues. Before moving this to production, we'll want to work with Valentin to update those CORS protections, and remove the proxy. 

This allows us to bypass the WMS server, which is primarily useful during scheduled batch maps or when running large batch map jobs. It also has the less tangible benefit of simplifying our infrastructure. 

I'll note that this handles both simple COGS and 4326, as well as COGS with non-standard CRSs, such as MODIS COGS that we stitch together (I'm not sure we can quite claim to "mosaic" but it doesn't appear necessary)

Demo: https://www.loom.com/share/bcfcb244323b4b96bd678212731cedab

**Before merge:** ndvi_dekad (shared) and precip_blended_dekad (Mozambique) are switched to type: "cog" for POC testing. We'll either revert those to wms.
- [x] Reverted to wms 

**Warning:** the API server with the new endpoints isn't deployed yet

### Key files
**Backend**

- api/prism_app/presigned_cog_url.py — STAC lookup and S3 presigning for COG assets.
- api/prism_app/main.py — Adds /cog_presigned_url and /cog_proxy endpoints.
- api/prism_app/tests/test_presigned_cog_url.py — Tests for presigning, caching, and proxy safety.

**Frontend**
- frontend/src/components/MapView/Layers/COGLayer/index.tsx — Fetches presigned URLs and renders COGs on the GPU via deck.gl.
- frontend/src/components/MapView/DeckGLLayersContext.tsx — Registry for deck.gl layers on the map.
- frontend/src/components/MapView/DeckGLOverlay.tsx — Single deck.gl overlay interleaved with MapLibre.
- frontend/src/config/types.ts — CogLayerProps type for COG layer config.
- docs/cog-layers.md — How it all works and how to add a COG layer.


frontend/vite.config.ts
```diff
@@ -88,6 +88,13 @@ export default defineConfig({
   },
   build: {
     outDir: 'build',
+    target: 'es2022',
```
gislawill, line 92: Flaggin'

wadhwamatic:
This is exciting to see, thanks Will for digging into this. As I mentioned briefly last week, the bigger opportunity may be with integration of non-HDC layers. The growing list of datasets on [Source Cooperative](https://source.coop/) are really intriguing. 

Beyond direct from COG, adding deck.gl support is another one I've been wanting for a while. I'm looking forward to tinkering with this. Let me know when we have something stable and ready to try out!

frontend/vite.config.ts
gislawill: Some impacts of this change (via Claude)
> The rough browser floor this sets is around Chrome 94+, Firefox 93+, Safari 16.4+, and Edge 94+. Anything older than roughly late 2021 / early 2022 vintage won't work — the browser will hit syntax errors on features esbuild left un-transpiled.

What would the floor have been before? (also via Claude)
> Vite's default build target is 'modules', which means browsers that support native ES modules (<script type="module">, dynamic import(), and import.meta). That floor is roughly Chrome 87+, Firefox 78+, Safari 14+, and Edge 88+ — so late 2020 era browsers.
> 
> The practical difference is pretty narrow. You're moving the floor forward by maybe a year or so (late 2020 → late 2021/early 2022). The biggest real-world gap is probably Safari: the default would have supported Safari 14 (Sept 2020), and now you need Safari 16.4 (March 2023), which could matter for users on older iPhones/iPads that can't update past a certain iOS version.

gislawill:
Leaving branch open for more work, closing PR for now

gislawill:
@ericboucher could you review this architecture / code? If it looks good to you, I'll deploy the server updates to allow testing in the staging deployment.

The goal is not to actually use any of these cog layers immediately. The goal is to have the functionality and documentation so we can easily do so in the future. I'm building off of this to scope out what it would take to support layers running off [Dynamical](https://dynamical.org/catalog/).

ericboucher:
@gislawill great work, I'll review a bit more but I have a few clarifying questions :
- could our backend actually get piled up as well ?
- it seems that we are not caching anything in the frontend map-wise? Shouldn't retoggling the same layer load instantly if we're just getting the same data ?
- how much more compute-heavy does this make the frontend process? Could it impact slower computers ?

gislawill:
@ericboucher good questions, see responses below:

> * could our backend actually get piled up as well ?

It certainly could. We want to treat this proxy server is just for the proof of concept. If we're going to rely on this path in production, we should work with Valentin to update the CORS settings on the HDC buckets so it's not necessary. 

> * it seems that we are not caching anything in the frontend map-wise? Shouldn't retoggling the same layer load instantly if we're just getting the same data ?

When the layer is toggled off, we run "unregisterLayer", which clears the Deck.gl cache in JavaScript memory. I believe a similar process happens for the WMS layer with the MapLibre cache. The WMS tiles are cached by the browser so they load much faster when re-enabled. When we remove our proxy server and are hitting S3 directly, we'll have that browser layer caching as well (we could enable it on our proxy server but I don't think it's super useful IMO)

> * how much more compute-heavy does this make the frontend process? Could it impact slower computers ?

There are few layers to this performance question. 
1. We're both loading more JavaScripts and taking on a more complex map initialization process when using Deck.gl interwoven. We don't even have Deck.gl layers on offer. This is a small hit with no benefit. I'm going to do an update on this branch where I make the Deck.gl initialization optional and will try to rely on lazy loading as much as possible of optional JS modules. 
     - Done ✅
2. For the actual rendering I don't think it's more compute-heavy. A slight oversimplification but you can think about MapLibre raster render process is more CPU-heavy and Deck.gl's is more GPU-heavy. Typically this results in a net performance improvement because the CPU is already under a lot of load from other elements of the web app and the GPU typically has more capacity. Theoretically, machines with underpowered GPUs will struggle more. I haven't seen too many reports of this but I'm going to do a deeper investigation and try to pull in some performance monitoring that we've used in past projects, though for a different user base. 
     - Just added the frames per second monitor and reporting to Posthog that we use. That's the best way we track GPU performance issues.

gislawill:
@ericboucher, let me know if you have any follow-up questions. This is ready for a thorough review.

api/prism_app/presigned_cog_url.py
```diff
@@ -0,0 +1,197 @@
...
+
+import boto3
+from botocore.config import Config
+from cachetools import TTLCache, cached
+from fastapi import HTTPException
+from pystac_client import Client
+
+logger = logging.getLogger(__name__)
+
+STAC_URL = "https://api.earthobservation.vam.wfp.org/stac"
```
c-mac: Overall this file looks great, but I'd propose a small refactor:

- `stac_config.py` module to make this `STAC_` code DRY (also used in `geotiff_from_stac_api`)

api/prism_app/main.py
```diff
@@ -571,6 +577,132 @@ def post_raster_geotiff(raster_geotiff: RasterGeotiffModel):
...
 
 
+@app.get(
+    "/cog_presigned_url",
+    responses={
+        404: {"description": "Collection or asset not found in STAC catalog"},
+        500: {"description": "Internal server error"},
+    },
+)
+def get_cog_presigned_url(
```
c-mac, line 665: No need to do in this PR, but I do think `main.py` is getting quite bloated and it would be nice to slowly start refactoring API routes into their own routers (as we did for admin)

api/prism_app/main.py
```diff
@@ -571,6 +577,132 @@ def post_raster_geotiff(raster_geotiff: RasterGeotiffModel):
...
+        ),
+    ),
+):
+    """Return short-lived pre-signed S3 URLs for COG assets in a collection.
+
+    For tiled collections (e.g. MODIS sinusoidal grid) a single date may
+    return many items, one per spatial tile.  Each matching item gets its
+    own presigned URL.
+
+    The URLs are valid for 5 minutes and support HTTP byte-range requests,
```
c-mac, line 687: Is 5 minutes long enough? What if someone's session is longer than 5 minute?

frontend/src/hooks/usePerformanceMonitor.ts
```diff
@@ -0,0 +1,180 @@
...
+
+interface PerformanceMonitorOptions {
+  enabled?: boolean;
+  logToConsole?: boolean;
+  onFPSUpdate?: (fps: number, change: number | null) => void;
+  significantChangeThreshold?: number;
+  onSignificantChange?: (fps: number, change: number) => void;
+}
+
+export const usePerformanceMonitor = (
```
c-mac, line 96: Hmmmm I grepped for this and am not seeing it used anywhere, intentional?

c-mac (commented):
Looks great overall! Left some small comments

frontend/src/hooks/usePerformanceMonitor.ts
gislawill: It's used in frontend/src/components/MapView/index.tsx

api/prism_app/main.py
gislawill: I'm not sure. I wanted to keep it as short as reasonable for this proof of concept. It'll be easy to extend when we have a more clear use-case for presigned cog urls (outside of the poc, maybe to scale up batch maps eventually?)

api/prism_app/presigned_cog_url.py
gislawill: Done ✅

frontend/src/components/MapView/index.tsx
```diff
@@ -26,8 +28,15 @@ import OtherFeatures from './OtherFeatures';
...
 const MapView = memo(() => {
   const classes = useStyles();
+  const posthog = usePostHog();
   const { iso3 } = useCountryIso();
 
+  usePerformanceMonitor({
+    onSignificantChange: (fps, change) => {
+      posthog?.capture('frame_rate', { fps, change });
+    },
+  });
```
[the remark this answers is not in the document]
gislawill: Addressed in 23d2df6 by sampling (10% of session). Overhead is genuinely really light so not general concern

frontend/src/utils/server-utils.ts
```diff
@@ -776,8 +779,9 @@ export async function preloadLayerDatesForWMS(
   const WCSWMSLayers = Object.values(LayerDefinitions).filter(
     (layer): layer is WMSLayerProps =>
       layer.type === 'wms' ||
+      layer.type === 'cog' ||
       compositeLayersWithDateLayerTypeMap[layer.id] === 'wms',
-  );
+  ) as WMSLayerProps[];
```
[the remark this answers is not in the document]
gislawill: Good catch, addressed in [c96f578](https://github.com/WFP-VAM/prism-app/pull/1844/commits/c96f5788784f19a2d81493e5f9efa223f8090435)

api/prism_app/main.py
```diff
@@ -621,6 +632,132 @@ def post_raster_geotiff(raster_geotiff: RasterGeotiffModel):
...
+        )
+
+    # Forward the Range header if present (COG tile readers use byte-ranges).
+    upstream_headers = {}
+    if request and "range" in request.headers:
+        upstream_headers["Range"] = request.headers["range"]
+
+    try:
+        async with httpx.AsyncClient(follow_redirects=True, timeout=60.0) as client:
+            upstream = await client.get(url, headers=upstream_headers)
```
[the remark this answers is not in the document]
gislawill: Addressed in commit 923edc9

frontend/src/components/MapView/Layers/COGLayer/index.tsx
```diff
@@ -0,0 +1,396 @@
...
+
+    const { width, height, data } = array;
+
+    // Convert Int16 (or any integer type) to Float32 for GPU compatibility,
+    // applying the wcsConfig scale factor if present (e.g. NDVI uses 0.0001).
+    const floatData = new Float32Array(data.length);
+    const scale = config.scale;
+    for (let i = 0; i < data.length; i++) {
+      floatData[i] = data[i]! * scale;
+    }
```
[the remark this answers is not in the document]
gislawill: Addressed in commit [403a060](https://github.com/WFP-VAM/prism-app/pull/1844/commits/403a06008b804fe2885d92b493a64ce8cc48f693)

frontend/src/components/MapView/Layers/COGLayer/index.tsx
```diff
@@ -0,0 +1,396 @@
...
+    const scaledNodata = nodata !== null ? nodata * config.scale : null;
+    const pipeline: RasterModule[] = [
+      { module: CreateTexture, props: { textureName: tileData.texture } },
+      ...(scaledNodata !== null
+        ? [{ module: FilterNoDataVal, props: { value: scaledNodata } }]
+        : []),
+      {
+        module: LinearRescale,
+        props: { rescaleMin: 0, rescaleMax: config.maxValue },
+      },
```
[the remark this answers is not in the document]
gislawill: Good catch, addressed in [017a449](https://github.com/WFP-VAM/prism-app/pull/1844/commits/017a449fb3fd9eb42f196d5ff0d25abb0e7b7e55)

ericboucher (approved):

PR #1848, opened 2026-05-06
Batch maps - backend worker

ericboucher:
## Summary

Adds **async batch map export** on top of the existing **synchronous `POST /export-map`** (unchanged for single-request flows).

- **`POST /export-map/jobs`** — enqueue a job; body is **`MapExportRequestModel`**. Returns **`job_id`**, **`status`**, metadata including **`origin_url`** (derived from the first map URL). **Dedupe** by **global request fingerprint** (no per-user scope).
- **`GET /export-map/jobs/{id}`** — job status; on success, **presigned S3 `download_url`** or **`local_artifact_path`** (local/dev). **`job_id` is the capability** — routes are **not** Basic-auth gated; treat links like secrets.

**Worker** (`prism_app.worker.export_map_worker`) claims **`FOR UPDATE SKIP LOCKED`** on Postgres, runs the existing **Playwright `export_maps`** pipeline, writes **PDF/ZIP** to **S3** or **local dir**, updates **`map_export_jobs`**.

## Database

- Alembic **`map_export_jobs_001`** (after **`prism_alerts_baseline`**): table **`map_export_jobs`**, index on **`request_fingerprint`**, columns including **`origin_url`**, artifact **`s3_uri`**, errors, timestamps.

**Deploy:** run **`alembic upgrade head`** against the **same** DB as **`PRISM_ALERTS_DATABASE_URL`**.  
If a DB still references a **removed** revision (e.g. old `map_export_jobs_002_*`), fix **`alembic_version`** or reset the dev DB before upgrading.

## Config

- **`EXPORT_MAP_S3_BUCKET`** — bucket only, **`bucket/prefix`**, or **`s3://bucket/prefix`** (single env).
- **`EXPORT_MAP_LOCAL_OUTPUT_DIR`** — optional dev **`file://`** artifacts.
- If **neither** bucket nor local dir is set, worker defaults to **`DEFAULT_EXPORT_MAP_S3_BUCKET`** in **`export_s3`** (`s3://prism-wfp/batch-maps`). **`docker-compose.deploy.yml`** mirrors that default for Compose.
- **`PRISM_ALERTS_DATABASE_URL`** + AWS creds/region for S3 where used.

**Secrets / shell:** `api/set_envs.sh` loads **`EXPORT_MAP_S3_BUCKET`** from **`PRISM_EXPORT_MAP_S3_BUCKET`** in Secrets Manager (see README / comment there for **required JSON shape** so `jq` does not fail).

## Docker / ops

- **`export_map_worker`** uses **`Dockerfile.export-map-worker`** (slim base + **Playwright**; **no** GDAL/rasterio/geopandas stack). Main API still uses **`Dockerfile`** (osgeo/GDAL).
- Compose: **`docker-compose.yml`**, **`develop`**, **`deploy`**, **`docker-compose.map-export-e2e.yml`**. Local smoke: **`scripts/map_export_job_docker_e2e.sh`**.

**Worker pip pins:** when bumping shared libs in **`pyproject.toml`**, update **`requirements-export-map-worker.txt`** accordingly.

## Frontend

- **`MapExportLayout`**: stable “fully loaded” gating for **`signalExportReady`** (Playwright `/export`) to avoid empty WMS/raster frames in PDFs.

## Tests / fixtures MOZ

- Fixture base URL for job/e2e helpers: **`https://prism.moz.wfp.org`** via **`prism_app.tests.fixtures.moz_export`**.

## Reviewer notes addressed

- No extra post-**`PRISM_READY`** sleep in **`export_maps`**; readiness handled on the frontend + stable polling where needed.
- Migrations merged to a single revision; obsolete **`002`** revision removed from tree.
- Lighter worker image; optional S3 key prefix folded into **`EXPORT_MAP_S3_BUCKET`** URI forms.


api/docker-compose.yml
```diff
@@ -16,3 +16,25 @@ services:
...
+  # Map batch export worker (map_export_jobs).
+  export_map_worker:
+    build: ./
+    shm_size: "2gb"
+    volumes:
+      - "./cache:/cache:rw"
+      - ./prism_app:/prism_app:ro
+    environment:
+      - KOBO_USERNAME=${KOBO_USERNAME:-kobo_user}
+      - KOBO_PASSWORD=${KOBO_PASSWORD:-test}
```
gislawill: nit: Will this worker need kobo access? I don't believe batch maps supports kobo-gated layers

api/docker-compose.deploy.yml
```diff
@@ -61,3 +61,34 @@ services:
...
+    container_name: "export_map_worker"
+    build:
+      context: .
+      dockerfile: Dockerfile
+    shm_size: "2gb"
+    networks:
+      - traefik_default
+      - default
+    environment:
+      - BROWSER_POOL_SIZE=1
```
gislawill, line 75: No need to change now but I'm hopeful we can up this. At least 2 on the current wms and hopefully 4 on the new wms.

frontend/src/components/MapExport/MapExportLayout.tsx
gislawill, line 1: Could you summarize this change? It's not clear to me how this differs from the previous implementation

api/prism_app/export_jobs/routes.py
gislawill, line 1: What would you think about an export_job directory for these new files?

api/prism_app/utc.py
gislawill, line 1: Let's put this in api/prism_app/utils.py rather than it's own file

api/prism_app/export_maps.py
```diff
@@ -25,6 +25,10 @@
 # Timeouts
 PAGE_TIMEOUT: Final[int] = 60000
 PRISM_READY_TIMEOUT: Final[int] = 60000
+# After PRISM_READY, brief pause so WMS/raster composites can finish painting (ms).
+EXPORT_MAP_POST_READY_MS: Final[int] = int(
+    os.getenv("EXPORT_MAP_POST_READY_MS", "1000")
+)
```
gislawill: I don't believe this should be necessary if PRISM_READY is working. This has the potential to meaningfully hurt performance (10 extra seconds for a 10 map batch map process)

ericboucher: I was getting empty maps, hence the changes here and in MapLayout - which fixed it

api/scripts/e2e_export_map_pdf.py
gislawill, line 1: Could you provide more context on these scripts? Are they meant to be used in production or are they more for development? Same question for api/scripts/map_export_job_docker_e2e.sh (though I think the answer may be different)

ericboucher: Just for testing in dev mode

gislawill:
This looks like a great foundation to me. I agree that we can run with just postgres (no sqs) for the foreseeable future. That worker polling with `SKIP LOCKED` is a good set-up for us here.

Regarding our PRISM_READY logic, I think we need to keep our process tight to get the best performance. Specifically, if there's an issue with tile loading, I'd like to understand why the current approach isn't working. We should avoid introducing significant wait period to the rendering of each map if at all possible.

api/prism_app/export_maps.py
ericboucher: I'll try to test a bit better if te changes to MapLayout alone were enough.

frontend/src/components/MapExport/MapExportLayout.tsx
ericboucher: MapExportLayout had a bug: the 500ms poll called signalReady() on the first checkFullyLoaded() pass, while the idle path correctly required several stable checks. So Playwright could screenshot as soon as MapLibre briefly reported “tiles loaded” before WMS/raster finished drawing → empty page in the PDF.

api/alembic/versions/map_export_jobs_002_origin_url.py
```diff
@@ -0,0 +1,55 @@
+"""map_export_jobs: requested_by -> origin_url."""
```
ericboucher: merge the migrations since they are in the same PR.

api/prism_app/export_jobs/db.py
```diff
@@ -0,0 +1,37 @@
+"""Session factory for map_export_jobs (same DB as alerts)."""
```
ericboucher, line 1: rename DB to "prism_db" throughout ? this also makes sens for the user permission changes

api/prism_app/tests/fixtures/staging_moz_export_map_request.json
```diff
@@ -0,0 +1,10 @@
+{
```
ericboucher: use code and url constructs from dict instead of this json, closer to the test run to make it easier to adapt/evolve, inc. updating the base URL

api/prism_app/tests/test_export_maps.py
```diff
@@ -134,9 +134,7 @@ def test_validate_export_url_support():
     validate_export_url(f"http://wfp.org/?test=1&date={TEST_DATE}")
     validate_export_url(f"https://prism.wfp.org/?test=1&date={TEST_DATE}")
     validate_export_url(f"http://prism.example.wfp.org/?test=1&date={TEST_DATE}")
-    validate_export_url(
-        f"https://staging-prism-frontend--1622-38oautsx.web.app/?test=1&date={TEST_DATE}"
-    )
+    validate_export_url(f"https://prism.moz.wfp.org/?test=1&date={TEST_DATE}")
```
ericboucher: revert

api/docker-compose.deploy.yml
```diff
@@ -61,3 +61,34 @@ services:
...
+      - PRISM_ALERTS_DATABASE_URL=${PRISM_ALERTS_DATABASE_URL:?"Provide a URL for the alerts database."}
+      - KOBO_PASSWORD=${KOBO_PASSWORD:?"Add a password to access KOBO data."}
+      - KOBO_USERNAME=${KOBO_USERNAME:?"Add a user to access KOBO data."}
+      - EXPORT_MAP_S3_BUCKET=${EXPORT_MAP_S3_BUCKET:?"S3 bucket for map export PDF/ZIP artifacts."}
```
ericboucher: Add relevant EXPORT_MAP_ envs to set_env script

ericboucher: export_map_worker doesnt need all these envs to run.

frontend/src/components/MapExport/MapExportLayout.tsx
```diff
@@ -291,8 +291,13 @@ function MapExportLayout({
 
     if (shouldTrackTileLoading && map) {
       let hasSignaledReady = false;
-      let idleCheckCount = 0;
-      const MAX_IDLE_CHECKS = 3;
+      let stableLoadedTicks = 0;
+      // Idle + poll share this counter: several consecutive observations that the map is
+      // fully loaded (not a single lucky areTilesLoaded() true—avoids empty WMS/raster frames).
+      const STABLE_LOADED_TICKS = signalExportReady ? 6 : 3;
+      const EXPORT_READY_SAFETY_MS = signalExportReady ? 60_000 : 25_000;
```
ericboucher, line 309: This seems quite long and risks slowing down batch maps a lot, no?

ericboucher:
@copilot fix API flaky CI test

frontend/src/components/MapExport/MapExportLayout.tsx
gislawill: It's definitely a risk. If we do get to that and it's not ready, that means the map failed to load completely. I think the question is, do we let it be slow or do we fail earlier? 

At this point I think we let it be slow. As usage of the feature picks up, we could make it fail earlier.

frontend/src/components/MapExport/MapExportLayout.tsx
gislawill: I see. Thanks.

api/set_envs.sh
```diff
@@ -27,5 +27,8 @@ export VECTOR_STORE_ACCESS_KEY_ID=$(aws secretsmanager get-secret-value     --se
 export VECTOR_STORE_SECRET_ACCESS_KEY=$(aws secretsmanager get-secret-value     --secret-id PRISM_AWS_S3_CREDENTIALS | jq .SecretString | jq fromjson | jq -r .AWS_SECRET_ACCESS_KEY)
 export VECTOR_STORE_DEFAULT_REGION=$(aws secretsmanager get-secret-value     --secret-id PRISM_AWS_S3_CREDENTIALS | jq .SecretString | jq fromjson | jq -r .AWS_DEFAULT_REGION)
 
+# Map batch export (worker). ``EXPORT_MAP_S3_BUCKET`` in secret: bare bucket, ``s3://bucket/prefix``, etc.
+export EXPORT_MAP_S3_BUCKET="$(aws secretsmanager get-secret-value --secret-id PRISM_EXPORT_MAP_S3_BUCKET | jq -r .SecretString | jq fromjson | jq -r .PRISM_EXPORT_MAP_S3_BUCKET)"
```
gislawill: I get this error when running this script
```
jq: error (at <stdin>:1): object ({"PRISM_EXP...) only strings can be parsed
```

api/docker-compose.yml
```diff
@@ -16,3 +16,20 @@ services:
       - PRISM_ALERTS_DATABASE_URL
     ports:
       - "80:80"
+
+  # Map batch export worker (map_export_jobs).
+  export_map_worker:
+    build: ./
```
gislawill: @ericboucher did you consider using a new docker file for the worker? I don't think the image needs GDAL, rasterstats, rasterio, rioxarray, duckdb, geopandas, stac lbs, pyarrow, starlette-admin. Even just moving off of the gdal base image would make a difference in terms of image size and spin up speed

gislawill (approved):
Looks good to me. Note the error I ran into with setenv and consider the suggestion of a separate docker file for the worker. Overall this is looking good and I'm excited to try running it in AWS.

api/docker-compose.yml
ericboucher: Good call, I switched to a lighter image.

PR #1859, opened 2026-05-11
Batch map async jobs - frontend

### LINKED ISSUE #1771: [Feature Request]: Batch Map Performance, Scaling & Progress Display

Batch map generation currently runs inline on the main PRISM Python server, meaning resource-heavy browser rendering (via Playwright) competes directly with user-facing traffic. A few simultaneous batch requests can lock up the server entirely. Users also have no visibility into progress — there's no way to know how far along a long-running batch job is.

**Proposed approach:** Move batch map generation to an asynchronous queue/worker model on AWS, fully decoupled from the main server.
On request, the Python server immediately enqueues rendering jobs; Playwright-based workers pull from the queue, render maps, and upload results to S3. The frontend surfaces live progress to the user (jobs completed vs. remaining).

Technical implementation details:
- Workers are pre-warmed to eliminate cold-start delays - at least one render instance stays running at all times
- Autoscaling workers (up to a configured max) handle concurrent batch request spikes
- Job status is tracked and surfaced via API: POST to enqueue a batch, GET by job ID to poll status and retrieve output
- AWS SQS used as the job queue; completed map outputs written to S3

ericboucher:
### Description

Use async jobs for batch maps exports

Screenshot 2026-05-12 at 8 29 00 PM
Screenshot 2026-05-12 at 8 28 13 PM

[batch maps.pdf](https://github.com/user-attachments/files/27611901/batch.maps.pdf)

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with


wadhwamatic:
@ericboucher - good to see this coming along. A few inputs from my side:
1. The filename is random. Can we use the same approach as before? Ex: Mozambique_2026_04_19_to_2026_04_30.zip (or .pdf)
3. The styling of the pop up when ready doesn't match the app UI very well. Let me know if you need suggestions
4. Instead of labeling the pop-up button 'Close', let's use 'Cancel'

frontend/src/components/NavBar/PrintImage/image.tsx
gislawill, line 1: Might be time to break up this file. It's handling a lot of functionality and complexity in these nearly 1000 lines

gislawill:
One of the big benefits of this new process is the ability to track progress as maps are being worked on. I believe the API endpoint already supports that with polling. What would you think about adding a status bar to this PR @ericboucher ? Potentially multiple stacked status bars if multiple batch jobs are submitted. When completed we can show the download button alongside the completed status bar. 

cc: @wadhwamatic

wadhwamatic:
> One of the big benefits of this new process is the ability to track progress as maps are being worked on. I believe the API endpoint already supports that with polling. What would you think about adding a status bar to this PR @ericboucher ? Potentially multiple stacked status bars if multiple batch jobs are submitted. When completed we can show the download button alongside the completed status bar.

Thanks @gislawill, I didn't realize this ability is available at this stage. We talked about a statu bar, and if this is the PR to add that, let's please do that. I'd like to see status of the batch process whether in the print screen context or in the main app. Let me think about some design options. I'll follow up on this tomorrow.

frontend/src/components/NavBar/PrintImage/batchMapExport/batchExportArtifactFilename.ts
```diff
@@ -0,0 +1,111 @@
...
+  ':',
+  '"',
+  '/',
+  '\\',
+  '|',
+  '?',
+  '*',
+]);
+
+export function sanitizeFilenamePart(input: string): string {
```
ericboucher, line 16: Maybe overkill?

frontend/src/components/NavBar/PrintImage/batchMapExport/batchExportArtifactFilename.ts
```diff
@@ -0,0 +1,111 @@
...
+export function buildBatchArtifactBasenames(
+  country: string,
+  layerId: string,
+  mapTimestampsUtc: readonly number[],
+  format: MapExportJobFormat,
+): { filenameBase: string; downloadFilename: string } {
+  if (mapTimestampsUtc.length === 0) {
+    throw new Error('buildBatchArtifactBasenames: empty map timestamps');
+  }
+  let minTs = mapTimestampsUtc[0];
```
ericboucher: aren't we running this logic multiple times?

frontend/src/components/NavBar/PrintImage/batchMapExport/batchMapExportJobsContext.ts
```diff
@@ -0,0 +1,6 @@
+import { createContext } from 'react';
```
ericboucher, line 1: Does it need its own file?

frontend/src/components/NavBar/PrintImage/image.tsx
ericboucher: done :)

gislawill:
@ericboucher Are there any new env vars/credentials needed to run the downloads? I'm running into an `Unable to locate credentials` error from botocore when running locally (after all maps are rendered, during the s3_client.put_object call)?

gislawill (commented):
The code all looks good to me. Thank you for adding those translations. I'm unable to run this locally due the credentials error.

ericboucher:
> @ericboucher Are there any new env vars/credentials needed to run the downloads? I'm running into an `Unable to locate credentials` error from botocore when running locally (after all maps are rendered, during the s3_client.put_object call)? 

That's an issue in the backend yes? Maybe your aws creds don't have S3 write access?

You should be able to run with the frontend only though and use the API

frontend/src/components/NavBar/PrintImage/batchMapExport/batchExportArtifactFilename.ts
ericboucher: removed in next PR

frontend/src/components/NavBar/PrintImage/batchMapExport/BatchMapExportJobRows.tsx
```diff
@@ -0,0 +1,257 @@
...
+          <React.Fragment key={job.clientId}>
+            {idx > 0 && <Divider component="li" />}
+            <ListItem alignItems="flex-start" className={classes.listItem}>
+              <Box className={classes.jobBody}>
+                <Typography
+                  variant="subtitle2"
+                  component="div"
+                  className={classes.jobLineText}
+                >
+                  {job.layerDisplayName}
```
gislawill: Missing translate here 

```suggestion
                  t({job.layerDisplayName})
```


Image

frontend/src/config/shared/translation/portuguese.json
```diff
@@ -27,6 +27,22 @@
   "Mean": "Média",
   "Median": "Mediana",
   "Baseline Layer": "Camada de base",
+  "Batch download completed successfully.": "Transferência em lote concluída com sucesso.",
+  "Batch export failed: {{message}}": "Falha na exportação em lote: {{message}}",
+  "Batch export is ready": "Exportação em lote pronta",
+  "Batch export started. This may take several minutes; a popup will appear when the file is ready.": "Exportação em lote iniciada. Acompanhe o progresso na janela de exportação em lote; trabalhos maiores podem demorar vários minutos.",
+  "batch_export_status_and_maps": "estado: {{status}} · {{current}} de {{total}} mapas",
```
gislawill, line 34: We'll also need to translate status here

Image

frontend/src/utils/constants.ts
```diff
@@ -1,5 +1,15 @@
...
 
+const LOCAL_EXPORT_PAGE_HOSTS = new Set(['localhost', '127.0.0.1', '[::1]']);
+
+/** Origin embedded in map export job URLs. Backend cannot fetch loopback; use public PRISM host when UI runs on localhost. */
+export function getMapExportPageOrigin(pageUrl: URL): string {
+  if (LOCAL_EXPORT_PAGE_HOSTS.has(pageUrl.hostname)) {
+    return 'https://prism.moz.wfp.org';
+  }
+  return pageUrl.origin;
+}
```
gislawill, line 11: I'm not a fan of this change. Two issues for local development:
- We'll be no longer able to develop the batch maps feature end-to-end
- We won't be able to render batch maps for other deployments from our local frontends

Instead, in our export workers, we can replace localhost with host.docker.internal (which we're already doing in render_single_map
Image

I'm guessing that 3000 port was too restrictive and that's way it didn't work for you @ericboucher. Let's loosen up that port selection to allow end-to-end batch mapping again

gislawill (changes requested):
Looking pretty good. My main concern is allowing end to end local batch development. Suggestion in comment above. 

Also noting some missing translation support.

ericboucher:
@gislawill I addressed your comments, I ended up using the "runLocally" flag, we could also do !runLocally instead if you think it makes more sense.

I am not sure we should have url overwrite logic in the backend as the workers should be independent from a frontend running alongside them. But definitely feel free to implement additional changes to suit your development needs :)

gislawill:
> I addressed your comments, I ended up using the "runLocally" flag, we could also do !runLocally instead if you think it makes more sense.

Thanks @ericboucher. I just updated to !runLocally (which feels more intuitive to me). Looks good to go!

gislawill (approved):

PR #1861, opened 2026-05-12
Limit to 12 maps for batch exports

### LINKED ISSUE #1845: [Feature Request]: Enforce 12-map limit for batch map operations

Enforce a maximum of 12 maps per batch map operation to protect server performance.

- Add validation on batch map creation (UI + server) to cap at 12 maps
- Display clear error/message to user (UI) when limit is exceeded
- Applies to both one-off batch jobs, scheduled TBD

WMS server can handle ~17–30 req/s depending on config, but concurrent batch operations with too many maps risk overloading the tile server. Team agreed on 12 as the safe ceiling during 4/30 dev sync.

ericboucher:
### Description

This fixes #1845.

Screenshot 2026-05-13 at 6 24 59 PM

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with


wadhwamatic:
let's change this text and use a red font instead of orange: 

> 13 maps match your settings. Each export includes only the 12 most recent (by date) to protect map servers.

to: 
> "Your selection exceeds the export limit of 12 maps. Clicking export will generate the 12 latest maps in your selection. You can also modify the date range to export the 12 maps you need"

I don't think we need to mention the same issue in the toast message.  Suggest to delete it unless there's a compelling reason to keep it.

wadhwamatic:
I don't think this is specific to this PR but somewhere along the way in this feature branch, there's an issue where the map extent in the app view does not stay intact when switching to the map print view. 

In addition, the map extent does not remain consistent from the single map print view to the batch map print view (extent shifts when redrawing after you toggle batch maps on)

wadhwamatic:
It seems like canceling a batch map in the frontend doesn't remove it from the queue? I canceled jobs and then the subsequent jobs show status 'queued · 0 of 12 maps' for a few minutes.

frontend/src/components/NavBar/PrintImage/batchMapExport/types.ts
```diff
@@ -11,10 +11,10 @@ export type BatchMapExportEnqueuePayload = {
   viewportWidth: number;
   viewportHeight: number;
   format: MapExportJobFormat;
+  /** Country label forwarded to API for backend filename stem. */
+  country: string;
```
gislawill, line 15: Thoughts on calling this "deployment" instead of country? More accurate for deployments like rbd

gislawill: Pedantic yes but I hope it's useful to force consistency

ericboucher: I agree, however country is the name we use throughout the app. So I think we should keep country for now, and consider changing throughout the app to "region" or "deployment".

wadhwamatic: this same topic came up with @c-mac  in the context of admin console / dashboards. my suggestion was to use country instead of deployment. even though country isn't fully accurate, it's more understood among the team

gislawill: I'll happily defer 👍

gislawill (approved):
Looks solid to me and runs well. Nice work

PR #1865, opened 2026-05-15
[frontend] upgrade React to 19.2

ericboucher:
### Description

Upgrade React to 19.2. This is a BIG upgrade, which needs thorough QAing

https://github.com/WFP-VAM/prism-app/pull/1628 was trying to do too many things at once...

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with


frontend/src/components/MapView/Layers/WMSLayer/index.tsx
gislawill: Can we remove this to now?

frontend/src/shims/react-dom.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * React 19's `react-dom` ESM namespace is non-extensible in production bundles,
+ * so runtime assignment of `findDOMNode` throws. MUI v4 / react-transition-group
```
gislawill, line 3: A good reason to prioritize https://github.com/WFP-VAM/prism-app/issues/1302

gislawill:
Heads up @ericboucher, looks like the datepicker is broken on this branch. I'm seeing this in the bottom timeline, batch map export config, and analysis panel. When these elements are clicked, the date picker calendar modal does not open

frontend/src/shims/react-dom.ts
ericboucher: Definitely the goal. But I tried to do react and mui at the same time and ended up with a broken mess 😅

gislawill: Haha fair enough. Good to know it's being looked into. Let us know if we can help

frontend/src/components/MapView/Layers/WMSLayer/index.tsx
ericboucher: Should we activate you mean?

ericboucher:
@gislawill nice catch on the datepicker, should be fixed now.

frontend/src/components/MapView/Layers/WMSLayer/index.tsx
gislawill: Nope. I mean, can we remove this commented code? Given the dropped lines above (removing `_expandedBoundingBox`), does it make sense to keep this todo? Without an any context, I'd say no.

frontend/package.json
gislawill, line 69: I don't believe sentry v7 has full support for react 19 https://sentry.io/changelog/react-19-support/

I'm not sure exactly what will work and what won't but we should assume we need to update to maintain current error tracking

gislawill:
Looking good to me other than the couple of comments. Nothing blocking. Thanks for the big updates here!

frontend/package.json
ericboucher: Let's keep it like tht for now, it's enough for our usecase and we might drop sentry soon anyway

wadhwamatic:
@ericboucher - I started testing this today on the firebase auto-deployment for Mozambique. I'll test locally too. No major issues found so far. 

I did find two small issues on the firebase deployment:
1. After running an analysis, I have to click twice to clear an analysis (nothing happens on first click)
2. Running an analysis with a baseline layer that is not a boundary gives no results (screenshot below)

Screenshot 2026-06-02 at 11 33 33

wadhwamatic:
@ericboucher - I noticed a minor change. Previously, if a layer was open and we switch to the charts tab (and tables too I believe), the legend for the layer would remain on screen. Its better that we hide it as was done in this PR (I think?). Curious if this was intentional. 

Everything else looks good. I tried hard to break this with multiple scenarios and found none besides those that were already noted and addressed. Well done!

wadhwamatic (approved):
I tried and could not break this. Well done!

ericboucher:
> @ericboucher - I noticed a minor change. Previously, if a layer was open and we switch to the charts tab (and tables too I believe), the legend for the layer would remain on screen. Its better that we hide it as was done in this PR (I think?). Curious if this was intentional.

not intentional, but hey 😅

PR #1867, opened 2026-05-16
[old] Schedule public maps - for disucssion

ericboucher:
## Summary

**scheduled public map exports** (JSON config, daily-style cron, queue priority, WMS-based latest-date resolution). It is **not** framed as a fix on top of an older “scheduled batch maps” PR—that approach is replaced here.

### Why “scheduled public maps” (naming)

The old wording implied **generic batch map generation** without a clear outcome. The real intent is: **run on a schedule** to detect **new WMS data** for chosen layers and **enqueue map exports** meant for **public distribution** (operator-owned config, separate from interactive users). **“Scheduled public maps”** signals **cadence + audience** (public-facing artifacts), not a vague “batch” UI feature.

### Approach

- **Add** `api/crons/cron_scheduled_public_maps.sh` to run daily.
- **Added** `map_export_job_priority_001`: `map_export_jobs.priority` (default **200** for API/user jobs). Cron enqueues at **100**. **Claim:** `ORDER BY priority DESC, created_at ASC`.
- **Config:** `api/scheduled_public_maps/config/` — job groups with **`country`**,  **`layer_ids`**, **`export_url_template`** containing `{date}` and `{layer_id}`, optional / viewport / `format`.
- **Dates:** Latest timestep per layer from **WMS GetCapabilities** on the default WFP datacube OWS root (`https://api.earthobservation.vam.wfp.org/ows`; `/wms` appended when needed, consistent with frontend `…/ows/wms` usage). Same *idea* as `alert-worker.ts` (“layer days” → max date), implemented in Python for this job.
- **Ops:** `api/crons/cron_scheduled_public_maps.sh` — same spirit as alerting crons; runs `export_map_worker` + `python -m prism_app.scripts.scheduled_public_maps_cron`.

**Status: WIP / draft** — public bucket layout, stricter “only if new data” behavior, and production DB/Alembic steps may still change.

## Test plan

- [ ] `cd api && PYTHONPATH=. poetry run pytest prism_app/tests/test_scheduled_public_maps_cron.py prism_app/tests/test_layer_days.py prism_app/tests/test_export_job_claim.py prism_app/tests/test_export_jobs_service.py -q`
- [ ] `alembic upgrade head` (incl. `map_export_job_priority_001`) on a dev DB; plan **stamp / manual steps** for DBs stuck on removed schedule revisions.
- [ ] `python -m prism_app.scripts.scheduled_public_maps_cron --config api/scheduled_public_maps/config/scheduled_public_maps.example.json --dry-run`
- [ ] Confirm workers drain **priority 200** jobs before **100** when both exist.


c-mac:
The architecture makes sense to me, but in order to change the cadence or job configs, someone will have to make code changes to the cron or the `config` dir. This was a pain point for dashboards, which is why we moved control into the admin from S3. 

Based on our call on Thursday, it makes sense to me that having scheduled maps run daily is fine (no admin control over scheduling needed), but I'm less convinced that removing admins ability to CRUD map configs is sustainable in the long run.

wadhwamatic:
@ericboucher, I agree with @c-mac's points above. We definitely need admins to be able to CRUD batch jobs, through a specific role. I don't want to rely on code changes to add or edit batch jobs which will be unknown to me

ericboucher:
@c-mac I agree that it may not be sustainable in the long run, but my thinking was that we didnt know enough about actual needs to make an implementation. It is very likely that we will need CRUD for this, but it is unclear to me what the flow will be.

My suggestion would be to start with somwthing lowkey, and iterate as we learn more. @wadhwamatic if you have a clearer vision and feel like we need a CRUD flow right from the start, let us know what this should look like :)

PR #1856, opened 2026-05-09
Dashboards: Export to JSON

conor-jh:
### Description

This PR adds the ability for users to export draft dashboards to JSON. All exports…

* Conform to the Zod schema
* Are uniquely named (dashboard path + timestamp)
* Reflect changes made in the dashboard edit flow

## How to test the feature:

- [ ] Create a new dashboard based on the flow from #1818 
- [ ] Make some edits and select **Export JSON**
- [ ] Observe the exported file

## Checklist - did you ...

Test your changes with

- [ ] `REACT_APP_COUNTRY=rbd yarn start`
- [ ] `REACT_APP_COUNTRY=cambodia yarn start`
- [x] `REACT_APP_COUNTRY=mozambique yarn start`
- [x] Add / update necessary tests?
- [ ] Add / update outdated documentation?

## Screenshot/video of feature:
https://www.loom.com/share/ea12d015a7ce42909e27d8818c68ed2b

PR #1880, opened 2026-05-20
Fix dates issue with charts block

conor-jh:
### Description

This fixes the issue that @wadhwamatic found in #1856: https://github.com/WFP-VAM/prism-app/pull/1856#issuecomment-4426290491

The problem is that we were incorrectly initializing a start date for empty chart blocks in newly-created dashboards. But there’s already date-setting logic. Basically we doubled up on trying to set dates which was causing the end date to incorrectly get set to a year from today.

## How to test the feature:

- [ ] Run any deployment and create a new dashboard in the UI with at least one chart block
- [ ] Observe the default start and end dates

image

## Checklist - did you ...

Test your changes with

### DIFF
--- a/frontend/src/components/CreateDashboardView/utils.ts
+++ b/frontend/src/components/CreateDashboardView/utils.ts
@@ -36,7 +36,7 @@ export const defaultElementForType = (
     case DashboardElementType.CHART:
       return {
         type: DashboardElementType.CHART,
-        startDate: today(),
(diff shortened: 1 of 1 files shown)


wadhwamatic:
This works well, thanks @conor-jh

There's some future work to consider how dates should work across dashboard elements. We should consider adding an option to sync dates across each element but this needs more validation.

PR #1881, opened 2026-05-20
Load khmer and arabic fonts for export; COUNTRY=jordan

### LINKED ISSUE #1879: [Bug]: Khmer and Arabic fonts do not render correctly for batch maps
### What happened?

Maps generated through batch maps do not correctly render Khmer font. The same issue does not exist when creating a single map. The same issue exists for Arabic (Jordan)

cambodia and jordan / main branch

left side is from printing a single map, right side is from a batch map

Image

ericboucher:
### Description

This fixes #1879 by loading fonts to make sure they are available in chromium/docker.
It also :
- Adds a few translations
- Improves the loading indicator for batch maps
- Improves arabic display
- Switches to `km` shorthand instead of `kh`

## How to test the feature:

- [ ]
- [ ]

## Checklist - did you ...

Test your changes with


ericboucher:
@copilot fix API audit failing CI

.github/workflows/api.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       - name: Security audit
         run: |
           cd api
-          poetry run pip-audit --desc --format json > audit_results.json 2>/dev/null || true
+          poetry run pip-audit --desc --format json --ignore-vuln PYSEC-2025-183 > audit_results.json 2>/dev/null || true
```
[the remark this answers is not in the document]
gislawill: I've got a similar fix here https://github.com/WFP-VAM/prism-app/pull/1884

The toml seems a bit more maintainable

frontend/src/config/shared/translation/english.json
gislawill, line 1: Curious what's the benefit of maintaining english translations given we fallback to english

frontend/src/utils/exportFontFamily.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  const family = cssFontFamily(primary);
+  const sampleText = language === 'km' ? 'កខគ' : 'العربية';
+
+  try {
+    await document.fonts.load(`400 16px ${family}`, sampleText);
+    await document.fonts.load(`600 16px ${family}`, sampleText);
+  } catch {
+    // Non-fatal: PRISM_READY safety timeout still applies.
```
gislawill, line 86: Do you care about monitoring if/how often this happens? We could track in posthog/sentry

frontend/src/utils/exportFontFamily.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  const primary = getExportFontFamily(language);
+  if (primary === 'Roboto') {
+    await document.fonts.ready;
+    return;
+  }
+
+  const family = cssFontFamily(primary);
+  const sampleText = language === 'km' ? 'កខគ' : 'العربية';
```
gislawill: Am I reading it right that this loads Arabic every time we don't load Khmer? Will we still load arabic if the language is set to french?

gislawill (approved):
Some questions/suggestions but nothing blocking. Thanks for handling this! Looks like it was tougher than I expected

frontend/src/config/shared/translation/english.json
ericboucher: Good question, it's a habit we've taken so that we can consider english.json the ground truth of what needs to be translated. But we could definitely reconsider if we find it annoying

frontend/src/utils/exportFontFamily.ts
ericboucher: Hmm good quetsion, maybe add to posthog, but don't think we should be hitting that

frontend/src/utils/exportFontFamily.ts
ericboucher: Nice catch, added conditional loading

PR #1847, opened 2026-05-06
Adding posthog

gislawill:
### Description

This addresses #594 by adding Posthog tracking. I've added @wadhwamatic and @ericboucher as owners (both using wfp email addresses).

This is currently set-up on a free plan with data stored on PostHog's EU servers. We use the "cookieless" setting which means we can avoid the cookies banner.

I can also note: posthog fails silently if not configured. This means we can hand over the code to a country office and they won't experience any issues if they choose not to track with Posthog.

TODO:
- [x] Set up new env vars in deployment processes

Pricing Info:
I believe the free plan should work for the forseeable future.
Screenshot 2026-05-06 at 10 03 03 AM

Limitations associated with "cookieless" implementation:
Screenshot 2026-05-06 at 10 12 13 AM

## How to test the feature:

- [ ] Run Prism locally
- [ ] Perform various events in the app
- [ ] Check them out in the posthog ui

## Checklist - did you ...

Test your changes with

- [ ] `REACT_APP_COUNTRY=rbd yarn start`
- [ ] `REACT_APP_COUNTRY=cambodia yarn start`
- [ ] `REACT_APP_COUNTRY=mozambique yarn start`
- [ ] Add / update necessary tests?
- [ ] Add / update outdated documentation?

## Screenshot/video of feature:
Screenshot 2026-05-06 at 10 17 20 AM

PR #1894, opened 2026-05-26
Adding posthog token to CI builds

gislawill:
### Description

This continues the work from #1847 by adding the new env variables to the build process.

This PR enables posthog error/performance/usage tracking on both staging builds and master branch builds.


.github/workflows/frontend.yml
```diff
@@ -207,6 +209,8 @@ jobs:
         country: [mozambique, cambodia, rbd]
     env:
       REACT_APP_SENTRY_URL: ${{ secrets.REACT_APP_SENTRY_URL }}
+      REACT_APP_POSTHOG_TOKEN: ${{ secrets.REACT_APP_POSTHOG_TOKEN }}
+      REACT_APP_POSTHOG_HOST: ${{ secrets.REACT_APP_POSTHOG_HOST }}
```
gislawill: @ericboucher will this `master_frontend_build_and_deploy` action actually deploy to mozambique, cambodia, and rbd? I had thought not but this file leads me to think it does.

@wadhwamatic how would you feel about enabling our cookie-less, anonymous usage tracking on these deployments?

.github/workflows/frontend.yml
```diff
@@ -146,6 +146,8 @@ jobs:
     env:
       GIT_HASH: ${{ github.event.pull_request.head.sha }}
       firebase_channel: ${{ github.event.pull_request.number }}
+      REACT_APP_POSTHOG_TOKEN: ${{ secrets.REACT_APP_POSTHOG_TOKEN }}
+      REACT_APP_POSTHOG_HOST: ${{ secrets.REACT_APP_POSTHOG_HOST }}
```
ericboucher: does this need to be a secret?

.github/workflows/frontend.yml
ericboucher: It deploys to staging envs, not the country office sites like prism.moz

.github/workflows/frontend.yml
gislawill: We could hardcode. The token doesn't need to be a secret either (both are included in the client bundle). 

Do you have a preference on how we manage these non-secret environment variables in actions?

ericboucher: I would hardcode the host, in the ts code directly? since it's not likely to move regularly.
And keep the token as en env like you did.

gislawill: Updated 👍

ericboucher (approved):

ericboucher:
Added secret `REACT_APP_POSTHOG_TOKEN` to github

PR #1895, opened 2026-05-27
Scheduled batch maps

c-mac:
### Summary

Adds "scheduled map" product. Scheduled maps share similar "template" options as batch maps (configured through the print modal), but rather than create maps for past dates, creates maps for future data, filtered by cadence. This feature is enabled by checking the "create scheduled maps" checkbox within the batch map creation interface. Admin cannot create a schedule from scratch—only clone an existing schedule.

More thorough product and technical context is [here](https://github.com/WFP-VAM/prism-app/issues/1883).

Some **noteworthy** technical changes in this PR:
- A daily-run worker job that checks for `active` map export schedules and enqueues a job to process them if there is new data (WMS) and respects the user cadence preference (dekad, quarterly, monthly). Uses WFP datacube WMS GetCapabilities for date discovery.
- Exposes the first mutation endpoint API (POST) to the PRISM UI (`POST /export-map/schedules`)
- Exposes the auth URL to the frontend: the "scheduled map" feature is auth-gated. After login, user is redirected to the print modal to resume schedule creation 

Other changes:

- **DB:** add `map_export_schedules` 
- **DB:** migrate `map_export_job` to foreign key back to the schedule (if created by schedule) and add `priority` field (scheduled maps are lower priority than batch maps for the queued worker)
- **DB:** migrate adding new row in `permissions` table for managing scheduled maps (`prism.map_exports.manage`)
- **Frontend:** refactoring around shared batch/schedule export template building

**Pre-deploy checklist**

- [ ] Run migrations
- [ ] `chmod +x api/crons/cron_scheduled_public_maps.sh`
- [ ] Add crontab on the API EC2 host (see above script for the cron to add)

**Post-deploy checklist**

- [ ] Run cron shell script
- [ ] Grant `prism.map_exports.manage` to schedule operators

## Testing checklist

Note that for most of the happy path flows, you will either need an admin account or an account with the new `map_exports.manage` permission.

- [ ] Create a schedule in PRISM, should see "success" toast appear
- [ ] In admin, verify the schedule was created with all the right parameters from the "map export schedules" page. Verify in both the list view, and the details view
- [ ] In admin, edit the existing schedule
- [ ] In admin, create a new schedule by cloning from an existing schedule. When cloning, the "layer" and "cadence" fields should be unset (and have dropdowns for setting them).
- [ ] In admin, delete an existing schedule
- [ ] Use the new `make schedule-cron` to test enqueuing jobs

## Screenshots

**Schedule frontend interface**

image

**Export schedules admin view**

image

## Loom

**Frontend video**
https://www.loom.com/share/b70cde0feecd4096b875f1392d279275

**Backend video**
https://www.loom.com/share/a2ca8dde4f5a42dab926ef141231adb8


gislawill:
Thanks for this @c-mac! Considering this PR is so large (and it is quite large 😅), could you add a few things to help reviewers?
- [x] A couple screenshots of the feature (new starlette page, prism ui log-in)
- [x] A loom walk-through of the feature
- [ ] A few comments in the code noting reasons behind decisions or places needing focus from reviewers

At the moment, I find that it's a bit overwhelming to review. Thank you in advance!

.github/workflows/api.yml
```diff
@@ -111,7 +111,12 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Build ${{ matrix.image_label }} image
-        run: docker build -f api/${{ matrix.dockerfile }} -t ${{ matrix.image_tag }} api
+        run: |
+          extra_context=()
+          if [ "${{ matrix.dockerfile }}" = "Dockerfile" ]; then
+            extra_context=(--build-context "layer_config=frontend/src/config")
+          fi
+          docker build -f "api/${{ matrix.dockerfile }}" "${extra_context[@]}" -t "${{ matrix.image_tag }}" api
```
gislawill: I'm curious why this `"layer_config=frontend/src/config"` context is needed now?

api/docker-compose.develop.yml
```diff
@@ -25,17 +25,22 @@ services:
         condition: service_healthy
     volumes:
       - ./prism_app:/prism_app
+      # Live-reload layer JSON without rebuilding the image (baked copy at /frontend-config).
+      - ../frontend/src/config:/frontend-config:ro
```
gislawill: This feels like a violation of separation of concerns. Two thoughts:
- It's not clear to me why need layer JSON on the backend. I'd think the important parts would be baked into the request/stored job
- If it is really necessary, I believe that's what the shared directory is for

api/prism_app/export_schedules/routes.py
```diff
@@ -0,0 +1,199 @@
...
+    response_model=MapExportScheduleCreateResponse,
+)
+def create_map_export_schedule(
+    body: MapExportScheduleCreateRequest,
+    prism: _ScheduleCreateSession,
+    session: Session = Depends(get_export_jobs_session),
+) -> MapExportScheduleCreateResponse:
+    user, _codes = prism
+    if user.id is None:
+        raise HTTPException(status_code=500, detail="Authenticated user is missing id")
```
gislawill: I don't think 500 is the right error code (some 40x error maybe?). Is this scenario possible given it's an authenticated route?

api/alembic/versions/map_export_schedules_001.py
```diff
@@ -0,0 +1,152 @@
...
+revision = "map_export_schedules_001"
+down_revision = "map_export_job_priority_001"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "map_export_schedules",
+        sa.Column("id", sa.String(), nullable=False),
```
gislawill: Let's use the UUID type here (that's what python allows sets it to)
`sa.Column("id", postgresql.UUID(as_uuid=True), nullable=False),`

api/alembic/versions/map_export_schedules_001.py
```diff
@@ -0,0 +1,152 @@
...
+    op.create_index(
+        "ix_map_export_schedules_created_by_user_id",
+        "map_export_schedules",
+        ["created_by_user_id"],
+        unique=False,
+    )
+
+    op.add_column(
+        "map_export_jobs",
+        sa.Column("map_export_schedule_id", sa.String(), nullable=True),
```
gislawill: This can be a UUID type as well

api/prism_app/database/map_export_job_model.py
```diff
@@ -38,6 +56,22 @@ class MapExportJob(SQLModel, table=True):
         default=None,
         sa_column=Column(String, nullable=True),
     )
+    map_export_schedule_id: Optional[str] = Field(
+        default=None,
+        sa_column=Column(
+            String,
```
gislawill: Same note about uuid

.github/workflows/api.yml
ericboucher: Not a huge fan of this either, curious if still needed or if there are other options

api/alembic/versions/map_export_job_priority_001.py
```diff
@@ -0,0 +1,44 @@
+"""Add map_export_jobs.priority for queue ordering (interactive > scheduled public).
```
ericboucher: Should we merge the two migrations? I usually favor one migration file per PR, unless there is a reason not to?

api/.env.example
```diff
@@ -5,6 +5,9 @@ PRISM_OIDC_REDIRECT_URI=http://localhost/auth/callback
 # These variables are generated by the WFP CIAM team when you register an OIDC client for the API
 PRISM_OIDC_CLIENT_ID=
 PRISM_OIDC_CLIENT_SECRET=
+# Comma-separated React app origins allowed for post-login modal returns.
+# Example: http://localhost:3000,https://prism.moz.wfp.org
+PRISM_FRONTEND_REDIRECT_ORIGINS=http://localhost:3000
```
ericboucher: Hmm, we already have some sort of frontend url whitelisting, let's try to consolidate?
cc @gislawill

api/docker-compose.develop.yml
ericboucher: Can you clarify why the layers are needed?
Could the frontend expose whatever is needed here as an asset?

ericboucher: If a shared file is cumbersome, we could have an automatic copy of the file if it changes (or a check in the CI). TLDR, it;s a bit too complex and not easy to maintain to use a context extension here.

api/docker-compose.develop.yml
```diff
@@ -58,26 +63,30 @@ services:
...
     volumes:
       - ./prism_app:/prism_app
+      - ./cache:/cache:rw
     build:
       context: ./
       dockerfile: ./Dockerfile.export-map-worker
     environment:
       - PRISM_ALERTS_DATABASE_URL=
-      - POSTGRES_HOST=host.docker.internal
+      - POSTGRES_HOST=db
```
ericboucher, line 72: Why are these changes needed? If they are, might be worth a separte PR?

api/prism_app/export_jobs/priority.py
```diff
@@ -0,0 +1,5 @@
+"""Queue priority for ``map_export_jobs`` (higher = claimed first)."""
+
+# Interactive / API-originated exports should preempt scheduled public map churn.
```
ericboucher, line 3: Another option would be to have the cron job space out schedule maps, say 1 map per minute or something like that?

wdyt? depends how many maps we'll have scheduled :)

api/prism_app/export_schedules/routes.py
```diff
@@ -0,0 +1,199 @@
...
+    params = {
+        key: _query_value(value)
+        for key, value in options.query_params.model_dump(
+            mode="json",
+            by_alias=True,
+            exclude_none=True,
+        ).items()
+        if value is not None
+    }
+    params.pop("baselineLayerId", None)
```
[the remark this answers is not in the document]
ericboucher: Interesting. This highlights a risk of droping new configs in the frontend if they have not been set in the backend.

I am not convinced this level of config granularity is useful in the API.

api/prism_app/tests/test_map_export_schedules_api.py
```diff
@@ -0,0 +1,247 @@
+"""API tests for POST /export-map/schedules."""
+
+from __future__ import annotations
+
+import os
+from collections.abc import Generator
+from uuid import UUID
+
+# main imports kobo which requires these at import time
+os.environ.setdefault("KOBO_USERNAME", "pytest")
```
ericboucher: Let's do this once for all tests?

frontend/src/components/NavBar/PrintImage/batchMapExport/mapExportTemplate.ts
```diff
@@ -0,0 +1,220 @@
...
+  const parsed = new URL(exportUrl);
+  const readableSearch = Array.from(
+    parsed.searchParams.entries(),
+    ([key, value]) => `${key}=${value}`,
+  ).join('&');
+
+  return readableSearch
+    ? `${parsed.origin}${parsed.pathname}?${readableSearch}`
+    : `${parsed.origin}${parsed.pathname}`;
+}
```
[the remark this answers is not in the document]
c-mac: Removing this feature entirely

api/prism_app/export_jobs/priority.py
c-mac: Are you saying throttle the worker to only enqueue one job per minute? Or update the cron to do a scan every minute?

If the latter, I'm not sure how long each job takes. If each job is takes more than a minute, the worker could get bogged down.

api/docker-compose.develop.yml
c-mac: This is putting the export worker on the compose network so that all three of db, api, and worker are all running together

c-mac:
@ericboucher @gislawill Check out [this](https://github.com/WFP-VAM/prism-app/pull/1895/changes/136101d2f312c37cc9b2906d894c158edc9a4f6d) change specifically for a way around the docker additional context for using the frontend layers in the backend

gislawill:
> @ericboucher @gislawill Check out [this](https://github.com/WFP-VAM/prism-app/pull/1895/changes/136101d2f312c37cc9b2906d894c158edc9a4f6d) change specifically for a way around the docker additional context for using the frontend layers in the backend

@c-mac, I'm happy with this new solution. Thank you!

api/docker-compose.develop.yml
gislawill: This change seems correct to me. My understanding is before this update (aka what's on master) the db requests from our workers were leaving the docker network to the host machine and routed back into the network via the exposed port. With this update, the workers' queries run like the api queries currently do: directly to the db in docker without leaving the network. No real change to functionality, just a cleaner implementation of the workers in our dev environment

wadhwamatic:
Nice work on this @c-mac (and thanks for the help to test this locally). A few inputs from my side:
1. When editing / creating / cloning new schedules via the admin console, we should limit the dekad interval to a max value of 3. There's no logical reason to do more than every 3 dekads
2. In the prism front-end, we have logic based on the layer selected and it's coverage to determine what cadence should be available. That logic doesn't exist in the admin console. It's probably complicated to implement, but I'd like to know how much effort would be involved to see if its worthwhile
3. Can we add a 'last executed' column that would display when the last map was produced?
4. It would also be useful to link to the output directory if possible
5. Instead of displaying the admin codes in the admin areas field, we should display the name of the admin area. I believe this would require a change to the print feature and can hold until pmtiles are done
6. Let's also create a separate issue to improve the way we're getting admin areas. The current approach is likely driven by default admin layer (level 3 for Mozambique) instead of using the actual selection 
7. The filter feature in the admin console should not be case-sensitive. i.e. search for mozambique should return Mozambique
8. Not sure why this is happening, but when I run Cambodia locally, the print preview page gets stuck at 'loading preview map' which prevents me from testing. Other countries seem ok

c-mac:
@wadhwamatic Regarding point #2, I'd prefer to add it in a separate branch with some small fast-follows. It's not too much work (on the order of .5-1 day), but I don't want to make this change any bigger.

Eric has created an issue for point #6: #1905

Everything else has been addressed!

wadhwamatic:
Thanks @c-mac and @ericboucher. A few additional inputs below:

1. I see a bug where the layers available in the admin tool are not correctly filtered to the selected country. In this example for Zimbabwe, I should not see INAM layers from Mozambique.
Screenshot 2026-06-01 at 10 39 23

2. For the output directory, is a full URL possible? 
3. I don't think this has been addressed: The filter feature in the admin console should not be case-sensitive. i.e. search for mozambique should return Mozambique

ericboucher:
Super nitpick, I am not convinced by the name cron for the file cron.py as we could technically have another automation logic than crontab.
I think in the frontend we use the term “runner” which starts and the logic is within “worker”

But we can also revisit in a follow up PR and/or when we bring in the alerting scripts to python.

ericboucher (approved):
Great work, thanks!

We definitelty need to improve admin area selection 😅 
We have a similar issue for alerting. https://github.com/WFP-VAM/prism-app/issues/1839

Issue #1761, opened 2026-03-10
[Feature Request]: Feature Filtering by ISO Code (Visualization + Analysis)


gislawill:
### Problem

PRISM currently operates without any country-level scoping: the visualization layer renders all features globally, the analysis layer sends the entire dataset for aggregation, and the chart layer allows admin unit selection from any feature in the source PMTiles — regardless of which country deployment is active. This makes country-scoped deployments impossible and produces incorrect results when a deployment should be limited to a specific country.

### Proposed Approach

Apply consistent ISO code-based filtering across all three layers:
- **Visualization:** Filter map features by ISO code at render time, with correct behavior across zoom levels
- **Analysis:** Apply the same ISO-based filter to aggregation queries (currently the full parquet source is used)
- **Charts:** Restrict admin unit selection to features matching the active country's ISO code

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

**Technical implementation details:**
- Visualization layer: filter PMTiles features by ISO code at render time; validate performance and correctness across zoom levels
- Analysis layer: currently sends the entire parquet source for aggregation — apply ISO filtering before aggregation runs
- Chart layer: currently allows admin unit selection from all features in source PMTiles — restrict to active country's ISO code
- Depends on Global PMTiles Data & Metadata Foundation (consistent ISO codes must exist before filtering can be applied)

PR #1896, opened 2026-05-27
Add universal URL-driven country deployment; COUNTRY=universal

### LINKED ISSUE #1763: [Feature Request]: URL-Driven Country Context for Any-Country Deployment
### Problem

Country selection in PRISM is currently coupled to the PMTiles setup process, with no way to initialize the application for a specific country directly from the URL. This means each country deployment requires custom code wiring rather than a clean, reusable configuration pattern. Missing or invalid country codes also have no graceful fallback.

### Proposed Approach

Introduce URL-driven country context so any PRISM deployment can be initialized for a specific country simply by including an ISO code in the URL:
- Read ISO code from URL parameters at application startup
- Initialize UI state and analysis context based on that country
- Decouple country selection logic from PMTiles setup so the two can evolve independently
- Handle unsupported or missing ISO codes gracefully (fallback behavior, clear error states)

**Technical implementation details:**
- Read ISO code from URL on app initialization (e.g. ?country=MOZ)
- Use ISO code to set initial country context in Redux/app state, independent of PMTiles load sequence
- Decouple country selection logic from PMTiles setup — the two should be independently configurable
- Graceful handling: if ISO code is absent, default to existing behavior; if invalid, surface a clear error state
- Depends on Global PMTiles Data & Metadata Foundation and Feature Filtering by ISO Code

gislawill:
## Description 
Addresses #1761 and #1763. Adds a **universal deployment** (`REACT_APP_COUNTRY=universal`) where a single build serves *any* country, chosen from the URL (`/country/:iso3`) instead of being baked in at build time. Existing per-country deployments (rbd, cambodia, mozambique, …) are unchanged.

**The core idea (read this first):** in a normal deployment the active country is a build-time constant (`appConfig.countryAdmin0Id`, `appConfig.multiCountry`, a baked boundary file). In a universal deployment that context becomes *runtime* state derived from the URL. Rather than fork every component, the PR introduces a thin
"effective context" layer that returns the build-time value in classic mode and the URL-derived value in universal mode. **Most of the diff is existing components swapping a hard-coded value for an `useEffective*` / `useCountryIso` call.** Once that pattern clicks, the 58 files read quickly.

### Broader Context
This PR lays the UI foundation. Reference these other PRs for other aspects of the feature:
- The data comes from this PR: https://github.com/WFP-VAM/vt-foundry/pull/12
- Additional routing pages and functionality follows in this PR: https://github.com/WFP-VAM/prism-app/pull/1954
- Translation support upcoming in work under this issue (not ready): https://github.com/WFP-VAM/prism-app/issues/1764

### Known Issues
- Charts work on Admin level 0, somethings Admin level 1, rarely/never on Admin levels 2 and 3. That's due to a change in the admin unit IDs in the source files and the HDC stats endpoint. We're trying to work with some folks on the WFP VAM team for a fix.
- Requires backend deployment for analysis to work in staging environment
- ~Countries that wrap anti-meridian don't zoom in~

### Architecture / how to read the diff
The feature turns "which country" from a **build-time constant** into **runtime state derived from the URL**, without forking the app. Five ideas, in dependency order:

**1. A new `universal` deployment target.** A single build (`config/universal/*`) serves any country instead of one baked-in country. `metadata.json` is the source of truth for per-country facts (bounding boxes, admin-3 availability).

**2. The URL is the country selector.** `/country/:iso3` feeds a `CountryIsoProvider` that parses, normalizes, and validates the code and exposes it via `useCountryIso()`. Pure, tested helpers in `utils/universal-utils.ts` back this up.

**3. An "effective context" seam keeps the rest of the diff small.** Instead of branching everywhere, components ask helpers like `useEffectiveBoundaryLayer` / `useEffectiveCountryAdmin0Id` for the *active* value — which is the old build-time constant in classic mode and the URL-derived value in universal mode. Most of the remaining file changes are mechanical swaps onto this seam, not new logic.

**4. Boundaries become ISO3-aware and PMTiles-backed.** The boundary cache is keyed by `layer::iso3` so countries don't collide, with a retry once PMTiles tiles finish loading (fixes blank "Go To" lists and empty chart locations). Boundary loading/filtering and the map viewport now scope to the active country.

**5. The backend filters by country too.** An optional `iso3_filter` flows through the stats API and into cache keys, preventing cross-country cache/analysis bleed. It's optional everywhere, so existing single-country deployments are unaffected.

Supporting changes: a `UniversalPlaceholder` landing page, removing the `404Page` (dead code),
and CI tooling (`validate-pmtiles`) plus an idempotent `pmtiles://` protocol registration.

## How to test the feature:
- [ ] `REACT_APP_COUNTRY=universal yarn start`, open `/country/MOZ` (and another ISO3) and confirm the map fits the country, boundaries/Go-To/chart locations populate, and analysis + chart popups work for that country only.
- [ ] Switch between two `/country/:iso3` routes and confirm no cross-country bleed in boundaries, analysis results, or cached charts.

## Checklist - did you ...

Test your changes with

Screenshot 2026-05-27 at 5 10 17 PM


frontend/src/components/MapView/Layers/BoundaryLayer/index.tsx
```diff
@@ -106,13 +111,6 @@ const BoundaryLayer = memo(({ layer, before }: ComponentProps) => {
     };
   }, [selectedMap, layer.minZoom]);
 
-  useEffect(() => {
-    if (layer.format === 'pmtiles') {
-      return initPmtilesProtocol();
-    }
-    return undefined;
-  }, [layer.format]);
-
```
gislawill, line 115: Duplicated frontend/src/components/MapView/Map/index.tsx

frontend/src/components/App/index.tsx
```diff
@@ -81,6 +84,40 @@ const Wrapper = memo(() => (
...
+        // @ts-expect-error - react-router-dom v5 types incompatible with React 18
+        <Route path="/country/:iso3">
+          <CountryIsoProvider>
+            <AppShell countryPrefix="/country/:iso3" />
+          </CountryIsoProvider>
+        </Route>
+      )}
+      {isUniversal && (
+        // @ts-expect-error - react-router-dom v5 types incompatible with React 18
+        <Route>
```
c-mac: Can this be moved up under the previous `isUniversal` block under a fragment?

frontend/src/components/CountryInvalidPage/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+  Typography,
+} from '@material-ui/core';
+import { wfpLogo } from 'assets/images';
+import { useSafeTranslation } from 'i18n';
+import { colors } from 'muiTheme';
+import { Link, useParams } from 'react-router-dom';
+
+function CountryInvalidPage() {
+  const classes = useStyles();
+  const { iso3 } = useParams<{ iso3?: string }>();
```
c-mac: Does this not need the `normalizeIso` hook call that other callers used? Maybe not since it is just assumed invalid getting to this page

frontend/src/context/useCountryIso.ts
```diff
@@ -0,0 +1,18 @@
+import { useContext } from 'react';
+
+import {
+  CountryIsoContext,
+  type CountryIsoContextValue,
+} from './countryIsoContext';
+
+export function useCountryIso(): CountryIsoContextValue {
```
c-mac: Feels like these hooks could be a consolidated into one that takes a boolean arg

frontend/src/context/analysisResultStateSlice.ts
```diff
@@ -383,6 +384,15 @@ export type PolygonAnalysisDispatchParams = {
 const CACHE_TTL_MS = 24 * 60 * 60 * 1000; // 24 hours
 const MAX_CACHE_SIZE = 4;
 
+function getIso3FromUrl(): string | undefined {
```
c-mac: Let's be careful to avoid this pass-thru functions, I find they just add needless cognitive load

frontend/src/context/analysisResultStateSlice.ts
```diff
@@ -383,6 +384,15 @@ export type PolygonAnalysisDispatchParams = {
...
+  return getIso3FromPathname();
+}
+
+function getAnalysisIso3Suffix(): string {
```
c-mac, line 391: Similarly, I'd argue this could also just take a boolean on including suffix

frontend/src/config/index.ts
```diff
@@ -33,6 +33,7 @@ import sudan from './sudan';
 import tajikistan from './tajikistan';
 import tanzania from './tanzania';
 import ukraine from './ukraine';
+import universal from './universal';
```
c-mac, line 37: Worth double checking how the dashboards feature on the backend uses countries. This data might need to be updated in the backend as well

frontend/src/config/universal/layers.json
```diff
@@ -0,0 +1,101 @@
+{
+  "universal_admin3_boundaries": {
+    "type": "boundary",
+    "format": "pmtiles",
+    "zones_path": "s3://prism-vector-storage/universal_admin_boundaries/admin_level=3/admin_boundaries.parquet",
```
c-mac, line 5: Just so I'm understanding correctly: each distinct admin level has its own parquet file, but all three boundary layers use the same boundary PMTile? (note that I resisted calling it "PMTile file" at Chris' behest!)

frontend/src/utils/universal-utils.ts
```diff
@@ -0,0 +1,130 @@
...
+import { getDisplayBoundaryLayers } from 'config/utils';
+
+type CountriesKey = keyof typeof universalMetadata.countries;
+
+const UNIVERSAL_ADMIN3_LAYER_ID: LayerKey = 'universal_admin3_boundaries';
+
+const ADMIN3_ISO3_CODES = new Set(
+  universalMetadata.admin3Countries
+    .map((code: string) => code.toUpperCase())
+    .filter(code => /^[A-Z0-9]{3}$/.test(code)),
```
c-mac: This regex is duplicated in `isValidIso3Format`

frontend/src/utils/universal-utils.ts
```diff
@@ -0,0 +1,130 @@
...
+    return layers;
+  }
+  if (hasAdmin3ForCountry(iso3)) {
+    return layers;
+  }
+  return layers.filter(layer => layer.id !== UNIVERSAL_ADMIN3_LAYER_ID);
+}
+
+/** Returns the normalized ISO3 code from the URL pathname, or undefined for non-universal deployments. */
+export function getIso3FromPathname(
```
c-mac, line 74: Can we remove this function entirely and just standardize on using `useParams`?

frontend/src/utils/universal-utils.ts
```diff
@@ -0,0 +1,130 @@
+import { appConfig } from 'config';
+import { BoundaryLayerProps, LayerKey } from 'config/types';
+import universalMetadata from 'config/universal/metadata.json';
```
c-mac, line 3: I'd propose refactoring `metadata.json` into a zod schema. It gives us typed, validated data for ISO3 logic (wouldn't need to use `keyof typeof universalMetadata.countries`, for example). The schema itself drive the ISO3 parsing logic as opposed to a bunch of small helper functions.

frontend/src/components/CountryInvalidPage/index.tsx
gislawill: Yeah I think we want the rawest form of our ISO param that we have so that it's clear to the user what's not working. Normalize could theoretically hide potential issues. That feels somewhat unlikely... honestly I don't think that makes a difference either way.

frontend/src/config/universal/layers.json
gislawill: Yes but they use different layers (layer_name) in the PMTiles. Technically, you could consider the parquet file the same too - just partitioned 🖖

frontend/src/utils/universal-utils.ts
gislawill: @c-mac would the zod schema need to have every country listed out? That list will presumably change in subtle ways over time. Do we want that to source us to update our zod types? I like the idea of better type support but hesitant to type dynamic keys

frontend/src/utils/universal-utils.ts
gislawill: Yea, I get that we have a lot of utilities here - open to ideas to simplify. Unfortunately, `useParams` can't be used here since it's used in Redux thunks, not components.

frontend/src/context/analysisResultStateSlice.ts
gislawill: I'm not sure I understand. Could you clarify?

frontend/src/utils/universal-utils.ts
c-mac: Agreed that we don't want to define list of countries in multiple places. Seems could define this list as an array in constants and then import it for the schema to use. If you don't want to take that on right now that's fine too, could just make a comment.

frontend/src/context/analysisResultStateSlice.ts
c-mac: I meant rather than defining a thin util function, update `getIso3FromPathname` to accept an `includeSuffix: bool` that handles this logic. It's a nit, non-blocking for me

gislawill: I'm leaning toward keeping it separate. `getIso3FromPathname` is a generic URL parser in universal-utils, and the `_<iso3>` formatting is more analysis-cache-key concern. An `includeSuffix` boolean would change its return contract (`undefined`->`''`, `MOZ`->`_MOZ`) based on a flag, which I think is less intuitive to understand than the named `getAnalysisIso3Suffix()`.

wadhwamatic:
Great work on this Will. I still have quite a lot of testing to do but here's some initial feedback:
1. Can we add a spinner / loader while boundaries are loaded? I'd rather see that something is pending than un-styled and/or simplified boundaries
2. We should limit the admin level available in the charts module and analysis levels to admin 2 for now
3. The admin areas toggle in print maps doesn't load a list of admin areas in some countries (ex: Zambia)
4. The broken admin IDs are preventing a lot of testing. I'm pretty unclear how we have a discrepancy in some cases (ex: Mozambique admin 2). I'll spend some more time trying to revisit this tomorrow

gislawill:
Thanks for the review @wadhwamatic! I've pushed up some updates, see details below:

> 1. Can we add a spinner / loader while boundaries are loaded? I'd rather see that something is pending than un-styled and/or simplified boundaries

Working on this in the #1954 PR, will edit the comment when ready. edit: Pushed up the update!

> 2. We should limit the admin level available in the charts module and analysis levels to admin 2 for now

Updated pushed to this branch and deployed to staging preview

> 3. The admin areas toggle in print maps doesn't load a list of admin areas in some countries (ex: Zambia)

Fix pushed to this branch and deployed to staging preview

> 4. The broken admin IDs are preventing a lot of testing. I'm pretty unclear how we have a discrepancy in some cases (ex: Mozambique admin 2). I'll spend some more time trying to revisit this tomorrow

Using the spatial join with the shapefile you provided yesterday, charts for admin levels 0, 1, and 2 are now working 🎉

ericboucher (approved):

frontend/src/components/Common/ChartFormComponents/ChartAdminLevelSelector.tsx
```diff
@@ -2,8 +2,10 @@ import { makeStyles, Typography } from '@material-ui/core';
 import SimpleDropdown from 'components/Common/SimpleDropdown';
 import { AdminLevelType } from 'config/types';
 import { useSafeTranslation } from 'i18n';
-import { range } from 'lodash';
-import { getAdminLevelCount } from 'utils/admin-utils';
+import { useEffectiveAdminLevelOptions } from 'utils/universal-country-admin';
+
+// TODO: Merge AdminLevelSelector and ChartAdminLevelSelector into one shared component
```
ericboucher: Should we open an issue to track this?

frontend/src/components/MapView/Layers/BoundaryDropdown/BoundaryDropdownOptions.tsx
```diff
@@ -227,7 +237,7 @@ const BoundaryDropdownOptions = React.forwardRef(
                   }
                   const features = combinedData.features.filter(f =>
                     boundaryLayers.some(layer =>
-                      String(f.properties?.[layer.adminCode])?.startsWith(
+                      String(f.properties?.[layer.adminCode]).startsWith(
```
ericboucher: Are you sure this is safe to change?

frontend/src/components/MapView/LeftPanel/ChartsPanel/index.tsx
```diff
@@ -238,15 +248,31 @@ const ChartsPanel = memo(() => {
...
+            data,
+            String(countryAdmin0Id) as AdminCodeString,
+            0 as AdminLevelType,
+          )
+        : getProperties(data);
+      setSecondAdminProperties(properties);
+      setSecondAdminLevel(0 as AdminLevelType);
     }
-  }, [secondAdminProperties, data]);
+  }, [secondAdminProperties, countryAdmin0Id, data, isUniversal]);
```
ericboucher: dedupable? not worth spending too much time though

frontend/src/components/MapView/Layers/BoundaryDropdown/BoundaryDropdownOptions.tsx
gislawill: No, I'm not positive. Reverted

frontend/src/components/MapView/LeftPanel/ChartsPanel/index.tsx
gislawill: Done 👍

frontend/src/components/Common/ChartFormComponents/ChartAdminLevelSelector.tsx
gislawill: On further investigation, it was incredibly simple - ChartAdminLevelSelector was no longer used anywhere. I just deleted that component (dead code) and cleaned up.

PR #1901, opened 2026-05-28
Add useLatestAvailableDate for dashboard map, chart, and table blocks.

gislawill:
### Description

This update adds the ability to set our dashboard component blocks to use the latest available date rather than a pinned date. It removes the default date from the config and replaces it instead with useLatestAvailableDate (bool) and pinnedDate. 

This update is made on maps, charts, and table blocks. 

## How to test the feature:
- [ ] Create a new Dashboard with a map, a chart, and a table
- [ ] Play around with the date options and check the preview and export. 

(notice the new "Use latest available dates" options)
Screenshot 2026-05-28 at 2 14 03 PM

And here's an example of the generated config
[new-dashboard-2_1779995646210.json](https://github.com/user-attachments/files/28364745/new-dashboard-2_1779995646210.json)


frontend/src/components/CreateDashboardView/utils.ts
```diff
@@ -17,46 +12,6 @@ export interface SlotConfig {
 
 export const MAX_SIDEBAR_SLOTS = 3;
 
-const today = () => new Date().toISOString().slice(0, 10);
-
-export const defaultElementForType = (
```
gislawill, line 22: Moved to it's own file (below)

frontend/src/components/DashboardView/ChartBlock.tsx
```diff
@@ -64,16 +77,70 @@ function ChartBlock({
...
   const formState = useChartForm({
     initialChartLayerId,
     initialStartDate,
     initialEndDate,
     initialAdminLevel: initialAdminLevel as AdminLevelType | undefined,
+    useLatestAvailableDate: useLatest,
+    latestPeriod: period,
   });
 
+  const persistBlockConfig = (
```
gislawill: The chart block was missing some persist logic and was failing to update the global store when in draft mode

frontend/src/components/DashboardView/ChartBlock.tsx
```diff
@@ -351,19 +421,65 @@ function ChartBlock({
...
+                </MenuItem>
+                <MenuItem value={ChartLatestPeriod.MONTH}>
+                  {t('One month')}
+                </MenuItem>
+                <MenuItem value={ChartLatestPeriod.QUARTER}>
+                  {t('One quarter')}
+                </MenuItem>
+                <MenuItem value={ChartLatestPeriod.YEAR}>
+                  {t('One year')}
+                </MenuItem>
```
gislawill: Seemed like reasonable chart options

frontend/src/dashboardConfig/schema.ts
gislawill, line 1: This is the most important change. Everything else is down-stream

frontend/test/fixtures/dashboard-config.sample.json
```diff
@@ -5,7 +5,7 @@
     "firstColumn": [
       {
         "type": "MAP",
-        "defaultDate": "2024-12-21",
```
gislawill, line 8: defaultDate is no longer supported! I find it to be a confusing variable name.

frontend/src/components/DashboardView/MapBlock.tsx
```diff
@@ -258,15 +262,30 @@ const MapBlockContent = memo(({ exportConfig, elementId }: MapBlockProps) => {
...
+              onChange={handleTitleChange}
+              placeholder={t('Enter map title') as string}
+              variant="outlined"
+              size="small"
+              fullWidth
+              className={classes.titleInput}
+            />
+            <FormControlLabel
+              control={
+                <Checkbox
```
conor-jh: Something to consider: would it make sense to position this as part of the date selector timeline? Seems a little weird that it’s visible and toggle-able without a layer selected. 

If it’s part of (or adjacent to) the timeline, it’ll be in context and only appear once a user has selected a date-compatible layer.

frontend/src/components/DashboardView/ChartBlock.tsx
conor-jh: Are these options supposed to be rolling? When I select “One year” I get data for Jan 1 2026–today vs. the default dates of a full calendar year (May 28 2025–today)

frontend/src/components/DashboardView/TableBlock.tsx
```diff
@@ -545,11 +586,25 @@ function TableBlock({
...
-                selectedDate={formState.selectedDate}
-                onDateChange={formState.setSelectedDate}
-                availableDates={formState.availableHazardDates}
-              />
+              <Box className={classes.dateColumn}>
+                <DateSelector
+                  selectedDate={formState.selectedDate}
+                  onDateChange={formState.setSelectedDate}
+                  availableDates={formState.availableHazardDates}
+                  disabled={useLatest}
```
conor-jh: Maybe update the selected date or suppress it altogether when `useLatest` is true? Might be a little misleading if the wrong date appears, even though the selector itself becomes disabled.

frontend/src/dashboardConfig/schema.ts
conor-jh: Do we care to handle the existence of `useLatestAvailableDate` *and* date fields as an invalid config? Or okay to fall back to `useLatestAvailableDate` taking precedence in the actual UI logic?

In other words, if a user edits a config manually and has conflicting date preferences, should we consider that invalid or just let `useLatestAvailableDate` trump all other date-related config options?

conor-jh (approved):

ericboucher:
Similar UX questions as @conor-jh 
Plus, how about using date=null in the config instead of adding a boolean?

frontend/src/components/DashboardView/ChartBlock.tsx
wadhwamatic: Just noticed the same issue. Dates cutoff at the beginning of the calendar year

wadhwamatic:
Thanks for taking this on @gislawill! The UI could use a little tweaking but the concept is generally there. We should consider adding a sync dates across blocks features. This could be helpful for specific dates (like a monitoring window during the growing season), and when we're just getting the latest available data (instead of manually selecting latest for each block). 

We'd need to think about the date range of a chart, vs the specific date for maps and tables. I like the period selection for charts. Maps show the coverage period correctly. Tables would benefit from also showing coverage. This is something we'd need to extent to the analysis feature more broadly.

Side note: at some point we discussed not forcing the analysis layer to be added to the map. I'd like to make that an option when adding a table block (add table results to map).

gislawill:
Thanks @conor-jh, @ericboucher, and @wadhwamatic for the feedback here! I've made some meaningful updates to run past you all. 

1. I've moved the "use latest data" toggle to each block's top bar for consistency (which I hope helps with clarity too). Screenshot below.
Screenshot 2026-05-29 at 10 45 30 AM

2. I've updated the chart date selection to allow for a reference ("current" OR "previous") selection before the period selection. For example, this allows a user to chose between the current year to date or the last full year of data.
Screenshot 2026-05-29 at 10 48 44 AM

3. I updated the schema in zod to use `transform` so the schema requirements change when "use latest" is selected. That helps us avoid invalid states like useLatestAvailable being true while start and end dates are defined. 

Also, I opened #1903 to address this point:
> Side note: at some point we discussed not forcing the analysis layer to be added to the map. I'd like to make that an option when adding a table block (add table results to map).

frontend/src/components/DashboardView/ChartBlock.tsx
gislawill: Updated, checkout item #2 here: https://github.com/WFP-VAM/prism-app/pull/1901#issuecomment-4577182836

frontend/src/dashboardConfig/schema.ts
gislawill: I updated the schema to use zod transform to help with this. Now they'll be flagged by zod as an invalid config. I don't think we need additional error handling here - shouldn't be possible to get into that state (as far as I'm aware)

frontend/src/components/DashboardView/MapBlock.tsx
gislawill: I moved this to the same place (block's top bar) on map, charts, and table for consistency. It's a bit removed from date selection in the block itself but I think the consistency across block types is worth the trade-off. 
See item #1 here: https://github.com/WFP-VAM/prism-app/pull/1901#issuecomment-4577182836

frontend/src/components/DashboardView/TableBlock.tsx
gislawill: Agreed, I hid the date selector when useLatest is true to help with this

conor-jh:
I think this is existing, but there’s a slight visual bug when opening dropdown in the chart/table blocks on longer dashboards. The scrollbar disappears, so the footer buttons shift slightly.

wadhwamatic:
@gislawill resharing the comments I posted in Slack here:

Thanks Will. These changes are looking good. I’m not able to see chart and table data when I edit or preview an existing draft dashboard. Charts and tables work as expected for new dashboards

For current vs last complete date range, I don’t think we need current. Last complete shows rolling months, quarters, years which is what we’d want.

We also don’t need dekad as an option for a chart period since it’s just one observation

gislawill:
Thanks @wadhwamatic! Responded in slack and will keep this comment up to date

> We also don’t need dekad as an option for a chart period since it’s just one observation
- [x] Removed

> For current vs last complete date range, I don’t think we need current. Last complete shows rolling months, quarters, years which is what we’d want.
- [x] Updated to use a rolling window, no current vs last complete option

gislawill:
> I think this is existing, but there’s a slight visual bug when opening dropdown in the chart/table blocks on longer dashboards. The scrollbar disappears, so the footer buttons shift slightly.

I'm having trouble reproducing @conor-jh. Could you add a screenshot? If it's pre-existing probably better suited for a fix in a different PR

ericboucher (approved):
nitpick, the "period" term and dropdown is not super clear.
Should we explicit "Last Month" / "Last Quarter" / "Last year" or something like that?

Here is how it shows in my accounting software :

Screenshot 2026-05-31 at 9 26 59 PM

gislawill:
@ericboucher, because we're using a rolling window, terms like "Last Month" or "Last Quarter" aren't really accurate. I think the most accurate frame would be "Last 30 days", "Last 3 months", "Last 12 months". We could also remove "period" and just have "Date Range". 

@wadhwamatic does that sound right to you?

ericboucher:
Yes or something like "Rolling period"
One month / one quarter / one year

wadhwamatic:
> @ericboucher, because we're using a rolling window, terms like "Last Month" or "Last Quarter" aren't really accurate. I think the most accurate frame would be "Last 30 days", "Last 3 months", "Last 12 months". We could also remove "period" and just have "Date Range".
> 
> @wadhwamatic does that sound right to you?

yes "last 30 days", etc. describes the rolling window correctly. and yes, the word 'period' can be dropped from the UI

gislawill:
> > @ericboucher, because we're using a rolling window, terms like "Last Month" or "Last Quarter" aren't really accurate. I think the most accurate frame would be "Last 30 days", "Last 3 months", "Last 12 months". We could also remove "period" and just have "Date Range".
> > @wadhwamatic does that sound right to you?
> 
> yes "last 30 days", etc. describes the rolling window correctly. and yes, the word 'period' can be dropped from the UI

Great, just updated to the following
| When use latest is selected | When it is not |
| - | - |
| Screenshot 2026-06-01 at 12 04 49 PM | Screenshot 2026-06-01 at 12 04 53 PM |

I'll merge as soon as CI passes.

PR #1902, opened 2026-05-28
Speed up lint: Updated lint-staged targetting and add a lint cached

gislawill:
### Description

Small update: I got annoyed with how long our lint run was taking with every commit (over 30 seconds on my computer) so I made a few small updates here:
1. Updating our lint stage targeting so that we only run our lint and formatting on staged files not all files.
2. Adding a cache when we do want to run lint so we don't rerun files that haven't changed.

The result: now our linting at commits is basically instantaneous (run on a fraction of the files). And if we want to run lint the whole codebase, it's a lot faster (due to cache).


ericboucher (approved):
Nice.
Clearly not the slowest step, but does it also get faster in CI?

gislawill:
> Clearly not the slowest step, but does it also get faster in CI?

It doesn't appear to unfortunately. The --cache flag must not persist the cache between runs

PR #1903, opened 2026-05-29
Move table block "Show on map" toggle into the edit top bar.

gislawill:
### Description

@wadhwamatic raised that it was confusing how to not show table results on the map. Well the option to do so is hidden until the data loads and it's not the fastest endpoint. This moves that option to the top and has it always ready. 

## How to test the feature:

- [ ] Enter edit mode on a dashboard
- [ ] Toggle "Show on Map"
- [ ] Load data and toggle again
- [ ] Confirm it's controlling the map in non-edit mode

Screenshot 2026-05-29 at 10 31 23 AM


ericboucher (approved):

ericboucher:
@wadhwamatic what would you say is the default? It might be more intuitive to say "hide map results"?

wadhwamatic (approved):
thanks @gislawill - this is gtg

PR #1917, opened 2026-06-03
Fix audit vulnerability

c-mac:
Another CI-blocking vulnerability.

More context on why this upgrade broke the google flood tests:

The only breakage from upgrading `aiohttp` to 3.14 was in test tooling: vcrpy 8.1.x still expects aiohttp.streams.AsyncStreamReaderMixin, which was removed in 3.14. That surfaced when pytest-recording opened VCR cassettes for test_google_floods_api.py (those tests use `requests`, not `aiohttp` — VCR just loads the aiohttp stub because aiohttp is installed).


c-mac:
This got lumped into a separate PR

Issue #1923, opened 2026-06-04
[Feature Request]: Scheduled batch map follow-ups


c-mac:
### Provide a clear and concise description of what you want to happen.

- All filtering options in the scheduled map export admin page should be case insensitive. #1924 
- Add "bulk row actions" in the list view of the scheduled map export admin page for updating status and deleting schedules. Both of these actions should prompt a confirmation modal. #1940 
- Update the "output path" field in the admin view to link to the actual directory where maps are saved
- Hide "country" column in admin views when logged in as a scheduled map manager, only show them if logged in as an admin #1941 
- Admin area enhancements, linked below #1930 
- Admin log outs should redirect to admin, PRISM log outs should redirect to PRISM #1929 

I did not specify more general admin UX improvements like making list view rows clickable, as those are better left to a separate issue.

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

More context on admin areas here: #1905

wadhwamatic:
> Send an email to the creator of the schedule whenever a map is generated. Use the "PRISM Alert Triggered" email as a template.

We should add 'subscribe to updates' as an editable field in the admin console, and also include the ability to unsubscribe in the email body. Or notify them to disable updates in the admin console

gislawill:
I just updated the issue description to cross out completed items. I believe the only remaining parts are:

- Update the "output path" field in the admin view to link to the actual directory where maps are saved
- Add 'subscribe to updates' as an editable field in the admin console, and also include the ability to unsubscribe

PR #1924, opened 2026-06-04
Case insensitive filtering for schedules in admin

c-mac:
### Description

This is part of #1923.

Extends existing `starlette-admin` filter functionality to be case-insensitive for all filtering operations for scheduled maps.

## How to test the feature:

- Create or find a schedule in admin
- Use the different filter options to test case insensitive (example: "Moz" for country with "starts with" feature should show rows with "mozambique")

## Checklist - did you ...

https://www.loom.com/share/75428d73210649069778390f3d1246b7

### DIFF
--- a/api/prism_app/admin_map_export.py
+++ b/api/prism_app/admin_map_export.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Any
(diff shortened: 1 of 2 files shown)


wadhwamatic (commented):
looks good from my side, thanks Cam

Issue #1765, opened 2026-03-10
[Feature Request]: Direct render WMS Layers from HDC COGs


gislawill:
### Problem

PRISM currently relies on a WMS tile server as an intermediary for rendering raster layers and for WMS capabilities calls (metadata discovery for layer availability and date ranges). This dependency adds operational overhead, creates a potential bottleneck, and means PRISM cannot easily function without the tile service running. As HDC data becomes more widely available as cloud-hosted COGs, there is an opportunity to reduce this reliance.

### Proposed Approach

This is an alternative/exploratory workstream with two parts:

**Part 1 — Direct COG rendering:** Render HDC raster layers directly from HTTP-accessible COGs using MapLibre's COG protocol, bypassing the tile server entirely for rendering. Assumptions: COGs are already publicly accessible and have appropriate overviews for interactive visualization.

**Part 2 — Replace WMS capabilities with STAC API:** Update PRISM's metadata discovery (layer availability, date ranges) to use the STAC API instead of WMS capabilities calls, making the STAC API the source of truth for available layers and their temporal coverage.

### Is there anything else you can add about the proposal? You might want to link to related issues here, if you haven't already.

**Technical implementation details:**
- Part 1: Use MapLibre COG protocol to render directly from cloud-hosted COGs (no tile server); expected edge case: layers with admin boundary filters may need extra iteration
- Part 2: Replace WMS capabilities calls with STAC API queries for layer/date metadata; STAC browser: https://data.earthobservation.vam.wfp.org/stac — STAC API docs: https://api.earthobservation.vam.wfp.org/stac/docs
- Note on Zarr: direct-from-Zarr rendering is emerging but introduces tradeoffs (multi-resolution data duplication); not recommended unless a single resolution meets all visualization needs

PR #1932, opened 2026-06-05
PoC: FTW field pmtiles

c-mac:
### Description

This implements #1765.

Renders a PMTiles layer (right now, for just Mozambique) of all agricultural field and boundary data for 2024-2025 using the [Fields of the World PMTiles](https://source.coop/ftw/global-data/predictions/vectors/alpha/global.pmtiles).

While this is only currently enabled for Mozambique, the fields will render worldwide if you pan around the map.

## How to test the feature:

- [ ] Run PRISM with Mozambique
- [ ] Toggle the "FTW field boundaries" layer (under vegetation)
- [ ] Behold lots of beautiful fields everywhere (and a legend)!

For comparison, you can visualize the dataset on [pmtiles.io](pmtiles.io) with the same PMTiles.

## Checklist - did you ...

Test your changes with

https://www.loom.com/share/5f0625679f084a14970ff6c7e4430ed5


wadhwamatic:
Very cool! FTW is getting quite a bit of attention, and that will increase this fall. Coincidentally, the Taylor Geospatial team will visit WFP HQ in a few weeks.

gislawill:
Switching to ready for review to trigger CI staging deployment

frontend/src/utils/clipPmtilesVectorTile.ts
```diff
@@ -0,0 +1,281 @@
...
+      ? [simplified.geometry.coordinates]
+      : simplified.geometry.coordinates;
+  polys.forEach(rings => {
+    rings.forEach(ring => {
+      ring.forEach(([lon, lat]) => {
+        minLon = Math.min(minLon, lon);
+        minLat = Math.min(minLat, lat);
+        maxLon = Math.max(maxLon, lon);
+        maxLat = Math.max(maxLat, lat);
+      });
```
gislawill, line 53: Could you explain what this is doing?

gislawill:
Very cool, thanks @c-mac! It's definitely challenging to work with a ton of small vectors.

I'm seeing pretty poor performance, bordering on unusable at times. We might be able to learn from the [publisher's explorer site](https://fieldsofthe.world/ftw-inference-app/#map=8.12/-82.5647/41.3853/mode:global/threshold:70/year:2025/opacity:90/downloads:0) (raster of field density at lower zoom levels, shifts to vectors only at high zoom levels). I'm seeing field density cogs hosted here: https://source.coop/ftw/global-data/predictions/confidence/field-density. We can coordinate with this work: https://github.com/WFP-VAM/prism-app/pull/1844

There are a few errors that keep coming up in the console, any idea what's causing them?
Screenshot 2026-06-29 at 9 40 39 AM
Screenshot 2026-06-29 at 9 38 08 AM

I think we can also move the year selection into a small drop-down list in the layer selection (select either 2024 or 2025). I think it's also easier to interpret when focused on one year. It also results in halving the data that we're trying to render.

c-mac:
The PoC now transitions to a COG that shows field density, inspired by [this](https://fieldsofthe.world/ftw-inference-app/#map=7.05/25.6829/-21.5478/mode:global/threshold:70/year:2025/opacity:90/downloads:0)

Biggest product concern (in my opinion) is constricting the rendered are to the deployed country.

gislawill:
Thanks @c-mac! Looking better. A few notes: 
* I think we should be confidence - I don't think we need a scale slider necessary but let's use a sensible value like 70%. Check the difference between our implementation and the default view of their explorer
    | Confidence Threshold 70% | Ours |
    | - | - |
    |Screenshot 2026-07-31 at 2 02 13 PM | Screenshot 2026-07-31 at 2 02 04 PM |

* I also see that the opacity slider for the layer doesn't change the opacity on the map.
    Screenshot 2026-07-31 at 2 04 59 PM

* The deck.gl MaskExtension looks like an easy way to clip to the country bounds (I'm hopeful). Looks worthy of a 15 minute timeboxed attempt to me.

c-mac:
@gislawill feedback addressed!

- Both the field boundary and density layers now work with opacity
- Both are country (deployment) clipped
- Both use a fixed 70% filter value for confidence interval for their respective datasets

gislawill:
Great, thanks @c-mac. This is looking a lot better to me! Let's loop @wadhwamatic and get his thoughts as well! 

@wadhwamatic, we'll give more context when we talk later this morning.

PR #1941, opened 2026-06-09
Hide country field for non-admins

c-mac:
### Description

This implements #1923 .

- Hides the `country` field in the map export schedule view for non-admins
- Allows users with map manager permission to enter admin and only access the schedules table

## How to test the feature:

- Login as an admin user and navigate to the schedules page, you should be able to see and search on the country field
- Change your permissions to a map export manager, you should only be able to access the schedules table, and not be able to see or search on the country field

## Checklist - did you ...

Test your changes with

https://www.loom.com/share/2826f76d27f348b2975f278403c2530f

### DIFF
--- a/api/prism_app/admin_map_export.py
+++ b/api/prism_app/admin_map_export.py
@@ -3,12 +3,13 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Any, Callable, Dict, Optional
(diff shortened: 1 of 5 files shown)


ericboucher (approved):

wadhwamatic (approved):
gtg thanks

PR #1951, opened 2026-06-12
Replace export clip-path with source-level clipping.

gislawill:
### Description

This is my attempt at more performant raster clipping for our admin units in the print config + export view. 

Rather than run the clip function at each map the render event (many times a second), this runs our clip function once when a tile is "fetched" by Maplibre. Using a custom protocol handler, we handle the fetching and caching of tiles ourselves (which is the downside of this approach) but allows us to clip them on load (once) rather than on render (thousands of times if map is panned or scrolled).

This should support admin boundaries from PMTiles as well — though not yet tested.

## How to test the feature:
- [ ] Pull up a layer (any layer)
- [ ] Open Print menu
- [ ] Select an admin unit
- [ ] Pan/zoom

## Checklist - did you ...

Test your changes with

Screenshot 2026-06-12 at 4 47 40 PM


ericboucher:
Love it, super smart, thank you @gislawill

wadhwamatic:
awesome work on this @gislawill !

PR #1952, opened 2026-06-13
Fix legend scale issue in batch map exports

ericboucher:
### Batch export legend scale fix

**Problem:** Batch PNG legends looked wrong vs print preview — blurry and/or wrong scale (reported on main too).

**Root cause:** Two separate issues:
1. **Resolution** — Playwright captures at `device_scale_factor=2`; WMS `GetLegendGraphic` at default DPI looked soft when scaled.
2. **Scaling approach** — Brief detour into intrinsic sizing (width/fonts) broke aspect ratio on export (vertical squish at 50%), especially for WMS image legends.

**Fix:** Align with main preview behavior:
- Legend size via CSS `transform: scale(legendScale)` + `transformOrigin` (uniform aspect ratio)
- Server export only: request WMS legends at `dpi: 192` to match Playwright DPR 2

Preview and export now use the same scaling mechanism; export should be sharper without squish.

---

### Known remaining gap (Auto aspect ratio)

Legend sizing can still **diverge more in Auto mode** — preview renders in the print dialog at live container size; batch export renders `/export` at fixed viewport (1200px-wide, height from aspect ratio). Same `legendScale` % ≠ same visual proportion if map frame size differs.

**Worth fixing fully?** Probably **low ROI** unless users rely heavily on Auto + batch:
- Fix needs preview/export dimension parity (pass measured preview width/height into batch URLs, or render export at preview pixel size)
- Non-trivial; touches batch URL builder, viewport calc, possibly scheduled maps
- Workaround: preset aspect ratios (A4 etc.) where viewport is deterministic

Suggest we keep in mind for now; revisit if it comes up in user testing.

### DIFF
--- a/api/prism_app/export_maps.py
+++ b/api/prism_app/export_maps.py
@@ -216,7 +216,7 @@ async def render_single_map(
         else:  # PNG
             result = await page.screenshot(
                 type="png",
-                full_page=False,
+                full_page=True,
             )
 
         await page.close()
--- a/api/prism_app/export_maps.py
+++ b/api/prism_app/export_maps.py
@@ -216,7 +216,7 @@ async def render_single_map(
         else:  # PNG
             result = await page.screenshot(
                 type="png",
-                full_page=True,
+                full_page=False,
             )
 
         await page.close()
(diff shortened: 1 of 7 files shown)


wadhwamatic:
@ericboucher - looks good on the deployed version. thanks

wadhwamatic (approved):

PR #1963, opened 2026-06-17
Admin: download map from schedule

c-mac:
### Description

This finishes #1923.

Allows admin users to download the latest exported map created from a given schedule.

**Technical note:** This PR adds `rustfs` docker service for the dev environment only. `rustfs` effectively replicates S3 locally, allowing us to generate presign urls for export jobs and download them, as opposed to mounting a filesystem folder like `/cache` with the exported maps. RustSF keeps the map export pipeline aligned with prod:

enqueue → worker put → DB s3_uri → head verify → presigned download

## How to test the feature:

In one terminal, run `make api`.
In another, run `make db-seed` to generate schedules locally.
Run `make schedule-cron-dry-run`: this should show four skipped entries
Run `make schedule-cron`: this will run the export worker, which will cause two of the five new schedules to yield maps
In admin, navigate to the "map export schedules view" and verify:

- The schedules that were skipped have a greyed-out download icon
- The schedules that ran successfully have working download icons, each of which triggers a download

## Checklist - did you ...

Test your changes with

https://www.loom.com/share/20a4854c0196406ea74571154beb4eca

**NOTE:** video cut off, but all I was going to show is that the "download" action should also work from the details view.

**Download icon in row**
image

**Download action in details view**
image


frontend/src/components/MapView/LeftPanel/AnticipatoryActionPanel/AALegend/index.tsx
```diff
@@ -24,23 +24,31 @@ function AALegend({
...
+          : undefined
+      }
+    >
+      {isDroughtAA && <AADroughtLegend showDescription={showDescription} />}
+      {isStormAA && <AAStormLegend />}
+      {isFloodAA && <AAFloodLegend />}
+    </Paper>
+  );
+
+  if (forPrinting) {
```
c-mac, line 45: Flagging that this change was necessary to get exports working locally. Was timing out due to our current (old) version of MaterialUI not working with React19. Probably can remove this once the upgrade goes in. CC @ericboucher

frontend/src/components/MapView/Legends/LegendItem/index.tsx
```diff
@@ -291,92 +291,96 @@ const LegendItem = memo(
...
-          )}
-        </Paper>
+              </>
+            </Box>
+          </>
+        )}
+      </Paper>
+    );
+
+    if (forPrinting) {
```
c-mac, line 377: Same as above re: MUI and React 19

ericboucher:
@c-mac I believe minio is in archive mode? Is there another alternative we could use? I like the local S3 replication idea though.

ericboucher (approved):

PR #1986, opened 2026-06-26
Proof-of-concept: Add dynamical.org Zarr layers with viewport-tiled deck.gl rendering.

gislawill:
### Description
Proof-of-concept exploration of a Deck.gl-Zarr layer built similarly to our COG layer (both using Development Seed's deck.gl-raster library). This implement uses open data from Dynamical and pulls metadata from their STAC API. It is also driven from our layer.json configs like other layers. 

Includes pretty thorough documentation. 

## How to test the feature:

- [ ] Pull up Mozambique
- [ ] Enabled prototype layers

## Checklist - did you ...

Test your changes with

https://www.loom.com/share/720c5f3c411140b59d31e4c5532482b8


wadhwamatic:
Very exciting to see this @gislawill! Performance is indeed quite good considering it's not at all optimized. A few inputs from my side:
1. After I select a layer, if I change the date using the back / next timeline buttons, I see new dates mentioned in the legend, but the visualization is the same
2. Running an analysis is definitely going to be the first ask for how to use this. It would be great to hear what that would require
3. We're working on getting ECMWF IFS ENS into HDC and then through OWS into PRISM. It's worth syncing with @valpesendorfer on this topic to explore tradeoffs between these approaches. We should definitely test on older computers with slower connections
4.  In general, the ECMWF datasets are the ones of highest interest. I haven't heard of much adoption of AIFS yet though imagine that would come soon
5. I got a kick out of Loom's spelling of Zarr in the video title

gislawill:
Thanks for the feedback @wadhwamatic. Pushed up some updates, notes below

> 1. After I select a layer, if I change the date using the back / next timeline buttons, I see new dates mentioned in the legend, but the visualization is the same

Fixed 👍 

> 2. Running an analysis is definitely going to be the first ask for how to use this. It would be great to hear what that would require

It would require a new endpoint on the server. This would be relatively straightforward because source.coop's data are already in CNG formats like Zarr. We could use xvec (or similar) on our python server to get _decent_ performance without much effort (just a informed guess, will learn a lot from a proof of concept).

It'd be tempting to update our current zonal stats endpoint at the same time (leverage the CNG benefits of HDC's COGs to make it faster) - though I know there is a cross-org, unified version in planning

> 3. We're working on getting ECMWF IFS ENS into HDC and then through OWS into PRISM. It's worth syncing with @valpesendorfer on this topic to explore tradeoffs between these approaches. We should definitely test on older computers with slower connections
> 4. In general, the ECMWF datasets are the ones of highest interest. I haven't heard of much adoption of AIFS yet though imagine that would come soon

I just upated our config to render the ECMWF datasets instead. Check out the forecasts in the staging deployments

> 5. I got a kick out of Loom's spelling of Zarr in the video title

Lol I love the idea of someone not knowing about CNG and instead thinking I was incorporating Russian mobility into our maps.

wadhwamatic:
Thanks @gislawill for using the ECMWF layers and fixing the date issue. I have a slow connection at the moment, so it's taking a 10-15 seconds to load but does function well once it does. As a POC, this PR is good to go from my side. I'd like to spend some time trying other Zarr datasets with STAC out there but that can come in due course.

> It would require a new endpoint on the server. This would be relatively straightforward because source.coop's data are already in CNG formats like Zarr. We could use xvec (or similar) on our python server to get _decent_ performance without much effort (just a informed guess, will learn a lot from a proof of concept).
> 
> It'd be tempting to update our current zonal stats endpoint at the same time (leverage the CNG benefits of HDC's COGs to make it faster) - though I know there is a cross-org, unified version in planning

I created a new issue on this topic to keep track of it. https://github.com/WFP-VAM/prism-app/issues/1989

wadhwamatic:
@gislawill - I just took another pass at this. It's looking great. I have a couple of comments:
1. I took a look at how `subtype` is actually used, and I think the values are misnamed. It only controls cube time layout - analysis (time) vs forecast (init_time + lead_time). Could we call the values `analysis` and `forecast` instead? Naming them after the provider makes it look like we'd add a subtype per data source, and I don't think that's how this would grow. Another icechunk dataset shouldn't need a new subtype, and a non-icechunk store would need a new store opener rather than a new subtype. 

2. Printing the dynamical layer doesn't crop the raster to the admin area. I see this is pre-existing for COG too, so not a blocker here. We can address this in a follow-up issue if you prefer.

gislawill:
@wadhwamatic, good call here:

> I took a look at how subtype is actually used, and I think the values are misnamed. It only controls cube time layout - analysis (time) vs forecast (init_time + lead_time). Could we call the values analysis and forecast instead? Naming them after the provider makes it look like we'd add a subtype per data source, and I don't think that's how this would grow. Another icechunk dataset shouldn't need a new subtype, and a non-icechunk store would need a new store opener rather than a new subtype.

I've renamed time_layout with the options 'analysis' or 'forecast'. I also updated the schema and markdown docs to better note what types of Zarrs are supported and what aren't (only icechunked zarrs with stac-based metadata at the moment).

gislawill:
@wadhwamatic, I believe @c-mac added the deck.gl-based raster cropping in the FOTW branch (for COGs). As a follow-up, let's pull that functionality into this Zarr layer as well. 

> Printing the dynamical layer doesn't crop the raster to the admin area. I see this is pre-existing for COG too, so not a blocker here. We can address this in a follow-up issue if you prefer.

PR #1995, opened 2026-07-03
Country-specific permission gating

c-mac:
### Description

Implements country-scoped permissions by way of extending the `user_permissions` table. 
The global admin (Amit) is still assigned the admin permission, but is assigned the wildcard character (`*`) for country.

Before this change, a user with the "managing dashboards" permission (for example) would allow them to update dashboards for any country. Now, a user could be assigned the managing dashboard permission but only for a specific country, like Malawi. Users can be assigned multiple permissions, as I imagine there's a case where the Malawi country officer might want Malawi-specific feature access for all of scheduled maps, dashboards, and AA drought.

**Important note**: The admin permission is still meant to be country-agnostic, meaning they can perform any action for any country, as well as view any data for any country. **Avoid creating user permissions that assign a user the admin permission plus a non-wildcard country.**

The following features are now affected by country-scoping:

- Scheduled maps
- Dashboards
- AA drought uploads

Product changes:

- **Admin user permissions form**: countries now uses a dropdown instead of free text
- List/create/edit/delete/count queries filtered by country on Dashboards, Scheduled maps, and AA drought views
- **Admin AA drought**: country hidden and inferred for single-country AA managers
- **Admin AA drought**: upon successfully uploading a CSV, redirect back to list view

Technical changes:

- Updates the seed data to create test users (and permissions for) each of the following roles and countries: [dashboards, scheduled maps, aa drought], [cambodia, malawi, mozambique]
- Allows testing aforementioned users in dev by setting `PRISM_DEV_USER_ID` to one of the above user ids, while also disabling auth (`PRISM_ADMIN_AUTH_DISABLED=true`), logic in `dev_impersonation.py`
- `POST /export-map/schedules` checks country scope
- `permission_scopes` loaded into admin middleware and returned from `/whoami`
- Adds migration to extend the `user_permissions` table

## How to test the feature:

```
make db-migrate && make db-seed
PRISM_ADMIN_AUTH_DISABLED=true
PRISM_DEV_USER_ID=<id_to_test>
# Malawi dashboard manager: sees only Malawi dashboards, not Cambodia/Mozambique
# Malawi AA manager: create drought CSV → appears in list, redirects after save
# Global admin (b0000008-...): sees all countries + Users/Permissions views
```

## Checklist - did you ...

Test your changes with

User permissions view with new country column:
image


c-mac:
@wadhwamatic to give a product review when possible

wadhwamatic:
@c-mac - sorry for the late review. I tested the three functions with good and bad pathways and everything worked as expected. One minor request - can we modify the error message when trying to create a scheduled map without the right permissions? 

Screenshot 2026-07-09 at 11 56 36

 Suggestion: Schedule creation failed: {"detail" :"You do not have the right permissions for this request. Contact wfp.prism@wfp.org for more information."}

c-mac:
Done! New error message:

Screenshot 2026-07-09 at 2 36 42 PM

@ericboucher feel free to review when you get free

ericboucher (approved):

